PR #6, opened 2017-01-17
Adding minification

depinav:
I added two things to this PR. The big thing is a starting point to breaking the library out into separate modules which would make working on it and maintaining it quite a bit easier. The second piece is a minification step to the process. This reduced the file size from 99kb to 35kb which would be faster for customers to load.

The main change that gives us these new abilities are webpack for the build process and babel to compile the ES2015 import/export to ES5 code so that it can run in non supportive browsers. The other thing this would allow is the ability to improve some other parts of the code by using things like block scoping with `let` and object destructring.


StevePrice:
Looks good, thanks.

PR #82, opened 2021-01-19
Exclude query parameters from mparticle request

colinschoen:
## Summary
Per security request, https://github.com/postmates/buyer-frontend/issues/5313, we need to exclude query parameters from mparticle requests. Mparticle is working on adding a configuration option for this, but we will fork and self-host the SDK until this configuration option is available.


colinschoen:
Apologies, mean to make this PR on a forked version. Closing.

Issue #88, opened 2021-01-25
[Bug] NPM Build Script - 'ENVIRONMENT' is not recognized as an internal or external command


Ahmedie-m:
Hey guys,

There seems to be a bug when running npm build. I have confirmed that rollup is infact installed and made sure it matches the same version as the one in package.json.

Below is the error given when attempting to run: `npm run build`

```
> @mparticle/web-sdk@2.12.3 build D:\Work\mparticle-web-sdk
> ENVIRONMENT=prod BUILDALL=true rollup --config rollup.config.js

'ENVIRONMENT' is not recognized as an internal or external command,
operable program or batch file.
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! @mparticle/web-sdk@2.12.3 build: `ENVIRONMENT=prod BUILDALL=true rollup --config rollup.config.js`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the @mparticle/web-sdk@2.12.3 build script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
```

Is there a solution to this? Thanks!

rmi22186:
It appears this this may be related to the OS you are using. Are you on windows?  It looks like a package like [cross-env](https://www.npmjs.com/package/cross-env) may help with this. I can add this to our backlog, but you can also feel free to test this out locally and see if you can get a PR going 
also. We love open source contributions!

rmi22186:
PR #118  fixes this

PR #118, opened 2021-03-18
fix: update npm scripts for Windows

krsanford:
Adding https://www.npmjs.com/package/cross-env to devDependencies, and updating script invocations.

## Summary
This fix addresses open issue #88 https://github.com/mParticle/mparticle-web-sdk/issues/88 .  Currently, npm scripts fail when run on a windows environment.  Implementing cross-env as a dev dependency as suggested in the issue.

## Testing Plan
I first verified that I was able to recreate the issue as presented in the issue.  After implementing cross-env, I tested locally on a Windows 10 machine, running all updated npm scripts/paths successfully.  npm run test and npm run build complete to success as executed from a Powershell console.  This will require further testing on other Dev OS's.

## Master Issue
It looks like https://go.mparticle.com/work/ is behind an mParticle auth wall.  I'm not sure what Master issue this closes, but it relates to issue #88


alexs-mparticle:
@krsanford Thanks for adding this patch to our sdk. I'm concerned that this package, [cross-env](https://github.com/kentcdodds/cross-env/issues/257) has been archived and will not longer be actively maintained. Is there an alternative that is actively maintained, or is this package considered "the standard" for non-un*x based node development. Excuse my ignorance, but I haven't had to code on windows in over 15 years.

krsanford:
@alexs-mparticle happy to help with my Windows machine 😄 
Yeah, I was disappointed to see that cross-env has been archived.  I have a few thoughts on options:
### 1 - Use cross-env anyway
While the package won't be maintained, I think that may actually help the stability of the code, as suggested by the owner:

> However, because of the extra features which were added (features which I never needed and only sometimes understood), the codebase has become a hodge-podge of different people's attempts at making things work. 

It appears that the owner has committed to maintaining security patches if needed:
>No new features will be added (and old features will not likely be removed). I don't plan to do any more than fix security/critical bugs and keep the Node.js support up-to-date.

This package has been used very widely within Windows Node development and has many more dependent packages than the recommended alternatives.  I think the risk presented is fairly low.

### 2 - Use dotenv
If you are concerned about option 1, the alternative I'd recommend would be using something like [dotenv ](https://www.npmjs.com/package/dotenv) to manage the environment variables outside of the package.json scripts and inside .env files within the repository.  Then, the scripts can use [dotenv-cli](https://www.npmjs.com/package/dotenv-cli) to apply a specific set of environment variables as needed.  This package is commonly used for managing complex environment variable configurations in the development lifecycle.  Though, it does add more complexity by introducing a new configuration file concept and format.  I would advocate this path if there were interest in more broadly applying env files as a common practice among other code repos for mParticle.

Let me know what you think, and feel free to nuke this if you like!

alexs-mparticle:
@krsanford Thanks for handling this. Love how well thought out your reasons are and I completely agree that it _might_ be safe to use `cross-env` even though it's technically archived.

I personally prefer `dotenv` since I use that exclusively for backend JS projects, though I do agree with you that it adds extra complexity to our build process, so we'd have to weigh the options.

@rmi22186 and I need to talk this over and see which solution works for us long term, so we'll get back to you.

rmi22186:
@krsanford - We've tested on both mac and windows and tests are passing. I'll go ahead and merge this in. Thanks so much for the PR!

rmi22186 (approved):

PR #160, opened 2021-07-29
feat: Allow scope param on user alias requests

luads:
## Summary
In some scenarios, it might be necessary to define the alias scope to something other than the default "device" option. This is not allowed on the web SDK at the moment, even though it is on the Identity API.

## Testing Plan
- Unit tested
- Tested on a local application that uses the SDK, with and without the scope param

I didn't change the `convertAliasToNative` method as I'm not aware if the native clients support this param.

## Questions
- What else needs change? Docs, TypeScript types, etc.
- Can someone review the TS type definitions? https://github.com/DefinitelyTyped/DefinitelyTyped/pull/54813

### DIFF
--- a/src/identity.js
+++ b/src/identity.js
@@ -192,6 +192,7 @@ export default function Identity(mpInstance) {
                     source_mpid: aliasRequest.sourceMpid,
                     start_unixtime_ms: aliasRequest.startTime,
                     end_unixtime_ms: aliasRequest.endTime,
+                    scope: aliasRequest.scope,
                     device_application_stamp: mpInstance._Store.deviceId,
                 },
             };
--- a/test/src/tests-identity.js
+++ b/test/src/tests-identity.js
@@ -2593,6 +2593,34 @@ describe('identity', function() {
         done();
     });
 
+    it('Alias request should include scope if specified', function(done) {
+        mockServer.requests = [];
+        var aliasRequest = {
+            destinationMpid: 1,
+            sourceMpid: 2,
+            startTime: 3,
+            endTime: 4,
+            scope: 'mpid',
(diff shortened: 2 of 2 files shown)


rmi22186:
Thanks so much for the well thought out PR @luads...you even got in a TS def which is fantastic!  🎉 

I'll have to double check with our native guys to determine if there should be a change with `convertAliasToNative`, and get back to you. I'd rather include that as a new commit if necessary rather than adding it as a new PR.

luads:
No worries at all @rmi22186, this will be quite helpful for us. Happy to raise a new PR for the native bits if needed later on.

rmi22186 (changes requested):
Well done!  Props for figuring out how the mocking works by looking a other tests!  Only 1 comment, which really is on us to document better.  

Can you remove the updates to the `dist` folder from the commit?  This will automatically be generated when we release the SDK via our Github Action.  This way our PRs a bit cleaner :)

Otherwise looks great.

rmi22186:
> No worries at all @rmi22186, this will be quite helpful for us. Happy to raise a new PR for the native bits if needed later on.

It appears it's not in native.  I can add an internal ticket for this...if it's a blocker for you, you can definitely PR their too, but no pressure :)

luads:
Thanks for the review @rmi22186! That makes total sense. I had to add the changes to the dist files to test it out, but I've removed them now.

As for the native implementation, we don't need that at the moment. If we do in the future, I'll give it a go :)

rmi22186 (approved):
Perfect!  Approving and merging.  We will run a release on Wednesday with a few other items we have teed up.

PR #216, opened 2022-03-01
ci: update indentation and job name in release.yml

rmi22186:
## Summary
This release.yml file is not running due to an indentation error.  Also updating `confirm-master-branch` --> `confirm-public-repo-master-branch` as the job name was updated but the `needs` that depends on the job didn't update its name.


einsteinx2 (approved):
LGTM

PR #227, opened 2022-04-04
ci: add archive npm logs task to workflow

alexs-mparticle:
## Summary
In some cases, we may have an npm error log that gets clobbered by Github Actions stdout. Github actions writes these to its file system but they can't be picked up by the UI. Archiving these logs as an artifact will allow us to pick them up.

## Testing Plan
This was tested as during sample apps development.

## Master Issue
Closes https://go.mparticle.com/work/SQDSDKS-3626


peterjenkins (approved):
LGTM

PR #238, opened 2022-05-13
feat: add package to batch application info

alexs-mparticle:
## Summary
Add Support for Bundle ID to Web SDK so that we can track Sample App or host app usage independent of workspace.

The expectation is that a user can pass in a `packageName` similar to our Native SDK and this will attach a `package` attribute to the `application_info` which can later be filtered along with `os` to determine app usage across three core platforms (Web, iOS and Android).

## Testing Plan
Manually test that web SDK continues to work with or without this new config attribute and run automated tests.

## Master Issue
Closes https://go.mparticle.com/work/SQDSDKS-3728

### DIFF
--- a/test/src/tests-core-sdk.js
+++ b/test/src/tests-core-sdk.js
@@ -216,7 +216,7 @@ describe('core SDK', function() {
     });
 
     it('should set Package Name on Batch Payload', function (done) {
-        mParticle.config.packageName = 'my-web-package';
+        mParticle.config.package = 'my-web-package';
 
         mParticle.config.flags = {
             eventsV3: '100',
--- a/test/src/tests-core-sdk.js
+++ b/test/src/tests-core-sdk.js
@@ -215,6 +215,30 @@ describe('core SDK', function() {
         done();
     });
 
+    it('should set Package Name on Batch Payload', function (done) {
+        mParticle.config.packageName = 'my-web-package';
+
+        mParticle.config.flags = {
+            eventsV3: '100',
+            eventBatchingIntervalMillis: 0,
+        }
+
+        mParticle.init(apiKey, mParticle.config);
+
+        window.fetchMock.post(
+            'https://jssdks.mparticle.com/v3/JS/test_key/events',
+            200
+        );
+
+        window.mParticle.logEvent('Test Event');
+        
+        var batch = JSON.parse(window.fetchMock.lastOptions().body);
+        
+        batch.should.have.property('application_info');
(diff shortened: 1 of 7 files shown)


test/src/tests-core-sdk.js
```diff
@@ -215,6 +225,31 @@ describe('core SDK', function() {
...
+        mParticle.init(apiKey, mParticle.config);
+
+        window.fetchMock.post(
+            'https://jssdks.mparticle.com/v3/JS/test_key/events',
+            200
+        );
+
+        window.mParticle.logEvent('Test Event');
+
+
```
rmi22186, line 245: nit: extra whitespace

src/mp-instance.js
```diff
@@ -223,6 +231,22 @@ export default function mParticleInstance(instanceName) {
...
+     */
+    this.setPackageName = function(packageName) {
+        var queued = queueIfNotInitialized(function() {
+            self.setPackageName(packageName);
+        }, self);
+
+        if (queued) return;
+
+        self._Store.SDKConfig.packageName = packageName;
+        self._Persistence.update();
```
rmi22186, line 248: Thinking about this more, I actually don't think we need the setter or getter on the API.  The point of this is to be set on the config, right?  I could see why we'd want a `getter`, but really a packageName shouldn't be changed mid-session.  You probably followed in the `setAppName` footsteps, but I'd argue we should remove that from our next major version and only allow them to do it on the config also since appName won't change mid session.

rmi22186: More food for thought.  Does native support either a getter or setter outside of where the bundle id lives and automatically gets read and sent to our servers?  My inclination is no, but we should double check...and if Native doesn't support this (or getter) and only supports automatically reading the bundleid and passing it on, that's even more reason to not support on web.  As a plus, this would help with code bloat too, and simplify this PR!

rmi22186: That would mean not referencing `getPackageName` in `serverModel.js` and just grabbing from the `_Store` directly also.

rmi22186 (changes requested):
I left a few large comments in `mp-instance.js` that I think could simplify the code further depending on if my thoughts there are right.

src/mp-instance.js
alexs-mparticle: I think that makes sense and is more consistent with Native. I'll refactor this.

rmi22186 (approved):

rmi22186 (approved):

PR #242, opened 2022-05-16
refactor: constants.js to TypeScript

alexs-mparticle:
## Summary
Migrate Constants JS file to TypeScript

## Testing Plan
Run automated tests and test locally

## Master Issue
Closes https://go.mparticle.com/work/SQDSDKS-4044

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -1,6 +1,6 @@
 import { version } from '../package.json';
 
-var Constants = {
+const Constants = {
     sdkVersion: version,
     sdkVendor: 'mparticle',
     platform: 'web',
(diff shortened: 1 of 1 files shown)


src/constants.ts
```diff
@@ -164,6 +164,6 @@ var Constants = {
         EventBatchingIntervalMillis: 'eventBatchingIntervalMillis',
     },
     DefaultInstance: 'default_instance',
-};
+} as const;
```
rmi22186, line 167: Is this `as const` necessary?

rmi22186, line 167: Is this `as const` necessary?

alexs-mparticle: Actually yeah. This is a `Const Assertion` It helps the compiler know for certain this is a constant with a set number of options and should be read only. It's perfect for what we're trying to do here.

PR #299, opened 2022-06-28
fix: prevent errors when onCreateBatch does not return a batch

rmi22186:
## Summary
When implementing the callback `onCreateBatch` previously, there was a bug where if a user did not return a batch from the callback, the code still attempts to set `modified` on the `undefined` batch, throwing an error.   While the code still worked (it did not upload the batch), there were errors that appeared in the console.  This PR ensures the batch exists before setting `modified` on it.

In the process of this, a few eslint errors cropped up, so I resolved those as well.

## Testing Plan
Local test app

## Master Issue
Closes https://go.mparticle.com/work/SQDSDKS-4065


src/batchUploader.ts
```diff
@@ -136,7 +136,11 @@ export class BatchUploader {
 
                 if (onCreateBatchCallback) {
                     uploadBatchObject = onCreateBatchCallback(uploadBatchObject);
-                    uploadBatchObject.mutated = true;
+                    if (uploadBatchObject) {
```
alexs-mparticle, line 139: Would it make useful to check if the `uploadBatchObject` is actually mutated rather than if it only exists?

peterjenkins (approved):

PR #313, opened 2022-07-24
feat: remove non-batching event code path (v2 events endpoint)

rmi22186:
## Summary
Now that we have everyone on the batching endpoint (v3), we can remove the code paths that relate to sending single events to the server (v2).   This encompasses:
* removing the options for `config.v1SecureServiceUrl` and `config.v2SecureServiceUrl` (I wonder if I should I keep these - its technically an API that I am removing, but it doesn't do anything anymore now that everyone has been migrated)
* checks for the v3 flag.  Now that everyone has `100` as this value, we can just implement this as a feature, rather than a feature flag.  This also includes checking for `shouldEnabledBatching` since this will always be true now.

The bulk of the code changes here relate to fixing tests.  The reason for this is because the integration tests involving inspecting what event the SDK sends to the server were almost all on the v2 endpoint since it predates v3.  Because the payload for the v2 events endpoint is different from the payload for v3 events endpoint, we have to fix the tests to now match the v3 endpoint.  V2 was stubbed using mockServer and v3 using fetchMock, so I made a few core changes across lots of tests
* `getEvent` turns into `findEventFromRequest`, `findBatchFromRequest`
* `mockServer.requests` turns into `window.fetchMock._calls`
I also renamed a lot of the event names so the tests are more clear.

The V2 subdomain is still used for sending forwarder data, so that can't be deleted from our constants.  Only V1 can be deleted there.

## Testing Plan
In my tests, I found a bug related to sessions that I filed here: https://go.mparticle.com/work/SQDSDKS-4317.  We were not updating the `LastEventSent` timestamp on v3 which was messing with a session test.  That is resolved simply by calling persistence.update() after queueing the event.

## Master Issue
Closes https://go.mparticle.com/work/SQDSDKS-3720


dist/build/src/constants.d.ts
```diff
@@ -108,7 +108,6 @@ declare const Constants: {
         readonly uploadInterval: 0;
     };
     readonly DefaultUrls: {
-        readonly v1SecureServiceUrl: "jssdks.mparticle.com/v1/JS/";
         readonly v2SecureServiceUrl: "jssdks.mparticle.com/v2/JS/";
```
rmi22186, line 111: FYI - we keep v2 here because the /forwarding endpoint still uses this

src/apiClient.js
```diff
@@ -15,24 +13,8 @@ export default function APIClient(mpInstance, kitBlocker) {
             this.uploader = new BatchUploader(mpInstance, millis);
         }
         this.uploader.queueEvent(event);
-    };
-
-    this.shouldEnableBatching = function() {
```
rmi22186, line 20: Not needed anymore since the flags are all 100 in the database now.  A to do for the future would be to remove the feature flag from being sent down from the server to save on size further.

src/apiClient.js
```diff
@@ -120,48 +98,6 @@ export default function APIClient(mpInstance, kitBlocker) {
         }
     };
 
-    this.sendSingleEventToServer = function(event) {
```
rmi22186, line 123: This is the v2 code to prevent single events to the server.

src/identity.js
```diff
@@ -1056,42 +1056,36 @@ export default function Identity(mpInstance) {
                         mpInstance._Persistence.savePersistence(cookies, mpid);
                     }
 
-                    if (mpInstance._APIClient.shouldEnableBatching()) {
```
rmi22186, line 1059: Previously, UIC and UAC were only on batching.  Now that all are on batching, we don't need this check

test/src/tests-batchUploader.ts
```diff
@@ -168,8 +167,6 @@ describe('batch uploader', () => {
             pendingEvents[1].EventName.should.equal(10);
             pendingEvents[2].EventName.should.equal('Test Event');
 
-            window.fetchMock.post(urls.eventsV3, 200);
```
rmi22186, line 171: I moved all references of mocking each events v3 endpoint from each test module to the test-main.js file https://github.com/mParticle/mparticle-web-sdk/pull/313/files#diff-30fa2bc6f63bcb9fbef2dbd45dfd00db46295e495547438dfc80e67ac3f4e5afR22-R27

test/src/tests-consent.js
```diff
@@ -3,19 +3,14 @@ import sinon from 'sinon';
...
-var getEvent = Utils.getEvent,
+var findBatch = Utils.findBatch,
     mockServer;
 
 describe('Consent', function() {
     beforeEach(function() {
         mockServer = sinon.createFakeServer();
         mockServer.respondImmediately = true;
 
-        mockServer.respondWith(urls.eventsV2, [
```
rmi22186, line 14: All instances of mocking urls.eventsV2 have been removed since this endpoint is gone

test/src/tests-core-sdk.js
```diff
@@ -38,11 +34,10 @@ describe('core SDK', function() {
     it('starts new session', function(done) {
         mParticle.startNewSession();
 
-        var data = getEvent(mockServer.requests, MessageType.SessionStart);
-
-        data.should.be.ok();
+        var sessionStartEvent = findEventFromRequest(window.fetchMock._calls, 'session_start');
```
rmi22186, line 37: all instances of `mockServer.requests` should be replaced with `window.fetchMock._calls` assuming the eventsv2 endpoint is what was previously being mocked.  Identity and other calls may still be using `mockServer`.

test/src/tests-core-sdk.js
```diff
@@ -229,30 +219,11 @@ describe('core SDK', function() {
         done();
     })
 
-    it('should set client id', function(done) {
```
rmi22186, line 232: Not sure what CGID was ever for, but the v3 events endpoint does not have this so I removed this test

test/src/tests-eCommerce.js
```diff
@@ -95,36 +91,39 @@ describe('eCommerce', function() {
...
+        purchaseEvent.data.product_action.products[0].should.have.property('coupon_code', 'my-coupon-code');
+        purchaseEvent.data.product_action.products[0].should.have.property('custom_attributes');
+
+        purchaseEvent.data.product_action.products[0].custom_attributes.should.have.property('customkey', 'customvalue');
 
         done();
     });
 
     it('should not log a ecommerce event if there is a typo in the product action type', function(done) {
+        // fetchMock calls will have session start and AST events, we want to reset so that we can prove the lo product action type will not go through
```
rmi22186, line 125: @rmi22186 fix this comment

test/src/tests-event-logging.js
```diff
@@ -235,46 +244,48 @@ describe('event logging', function() {
...
     it('should log an AST with firstRun = true when first visiting a page, and firstRun = false when reloading the page', function(done) {
-        var data = getEvent(mockServer.requests, MessageType.AppStateTransition);
-        data.should.have.property('at', 1);
-        data.should.have.property('fr', true);
-        data.should.have.property('iu', false);
+        var astEvent = findEventFromRequest(window.fetchMock._calls, 'application_state_transition');
+
+        astEvent.data.should.have.property('application_transition_type', 'application_initialized');
+        astEvent.data.should.have.property('is_first_run', true);
+        astEvent.data.should.have.property('is_upgrade', false);
```
rmi22186, line 251: @rmi22186 check why there's an if clause in a test

test/src/tests-identities-attributes.js
```diff
@@ -1089,78 +1055,4 @@ describe('identities and attributes', function() {
 
         done();
     });
-    
-    it('should not send user identity change requests when not batching', function(done) {
```
rmi22186, line 1093: we are always batching now.  can remove this test

test/src/tests-identity.js
```diff
@@ -1236,19 +1223,6 @@ describe('identity', function() {
         done();
     });
 
-    it('should convert user identities object to an array if no identityRequest is passed through', function(done) {
```
rmi22186, line 1239: Sent into V3 as an object, not an array. Can remove

src/apiClient.js
```diff
@@ -2,8 +2,6 @@ import Constants from './constants';
 import Types from './types';
 import { BatchUploader } from './batchUploader';
 
-var Messages = Constants.Messages;
```
alexs-mparticle, line 5: If you're removing this instance of `Messages`, are you removing any related messages from `constants.ts`?

src/constants.ts
```diff
@@ -124,7 +124,6 @@ const Constants = {
         uploadInterval: 0, // Maximum milliseconds in between batch uploads, below 500 will mean immediate upload
     },
     DefaultUrls: {
-        v1SecureServiceUrl: 'jssdks.mparticle.com/v1/JS/',
         v2SecureServiceUrl: 'jssdks.mparticle.com/v2/JS/',
```
alexs-mparticle, line 127: Shouldn't we also remove v2?

test/src/tests-consent.js
```diff
@@ -3,19 +3,14 @@ import sinon from 'sinon';
 import { urls } from './config';
 import { apiKey, MPConfig, testMPID } from './config';
 
-var getEvent = Utils.getEvent,
+var findBatch = Utils.findBatch,
```
alexs-mparticle, line 6: Nitpick: Can we rename `test/src/utils.js` to `test/src/test-utils.js`? I even think the object should be renamed to `TestUtils` for clarity, but that's more of a personal preference.

src/apiClient.js
```diff
@@ -15,24 +13,8 @@ export default function APIClient(mpInstance, kitBlocker) {
             this.uploader = new BatchUploader(mpInstance, millis);
         }
         this.uploader.queueEvent(event);
-    };
-
-    this.shouldEnableBatching = function() {
-        // Returns a string of a number that must be parsed
-        // Invalid strings will be parsed to NaN which is falsey
-        var eventsV3Percentage = parseInt(
-            mpInstance._Helpers.getFeatureFlag(Constants.FeatureFlags.EventsV3),
```
alexs-mparticle, line 24: Are we going to remove `EventsV3` from `constants.ts` and other locations as well, since there's no need to choose v3 and v2?

alexs-mparticle (changes requested):

src/constants.ts
rmi22186: I initially _did_ v2 also, but v2 is actually used when creating the /forwarder url when we report forwarding stats, so we need to keep this.

src/apiClient.js
rmi22186: it's a good call out.  I think I'll remove the block about 8/12/2022 as well since it references it

test/src/tests-consent.js
alexs-mparticle: My main reason for this is so that we don't confuse `Utils` for `TestUtils`.

src/apiClient.js
rmi22186: yes. I removed the eventsv3 flag there

alexs-mparticle (approved):

rmi22186:
I will merge this o 8/12/2022, which is 1 month after we moved everyone to v3, and [per the TODO we had about removing the logging warning](https://github.com/mParticle/mparticle-web-sdk/pull/313/files#diff-b7bffff8ba6302afccbb9379c8890f043015e79c992b2f7ca1cbb4845018f5fcL1221-L1227).

rmi22186:
This has to be rebased.  It' a very old PR.  I am closing.

PR #314, opened 2022-07-25
ci: Add NPM Scripts to compress and log bundle size

alexs-mparticle:
## Summary
Add NPM Scripts that will allow us to compress the js bundle and report file size to be used in a later github action

## Testing Plan
N/A

## Master Issue
Closes https://go.mparticle.com/work/SQDSDKS-4316

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,12 @@
         "lint": "eslint src/ test/src/",
         "gts:check": "gts check",
         "gts:fix": "gts fix",
-        "prettier": "node_modules/.bin/prettier --check \"**/*.js\""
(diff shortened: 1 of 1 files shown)


rmi22186 (commented):
I'm still not convinced that we need to report the unbundled size. Maybe there's a use case and there is value to it, but customers won't ever need to deal with our unbundled script in a production setting.  It would simplify the code here and your other related PRs 50% if we remove it.  

Otherwise this looks great.

alexs-mparticle:
@rmi22186 Works either way for me, but I'll remove it since we don't need it. We can easily add it back if we find a need for it.

rmi22186 (approved):

PR #318, opened 2022-07-26
ci: Create action to calculate percent difference

alexs-mparticle:
## Summary
Create Reusable Github Action to calculate difference as a percent between two numbers

## Master Issue
Closes https://go.mparticle.com/work/SQDSDKS-4352

### DIFF
--- a/.github/workflows/calculate-difference.yml
+++ b/.github/workflows/calculate-difference.yml
@@ -24,8 +24,8 @@ jobs:
         name: Calculate Bundle Size Difference
         runs-on: ubuntu-latest
         outputs:
-            diff: ${{ steps.get-diff.outputs.diff }}
-            diff_percent: ${{ steps.get-diff.outputs.diff_percent }}
+            diff: ${{ steps.report-diff.outputs.diff }}
+            diff_percent: ${{ steps.report-diff.outputs.diff_percent }}
         steps:
             - name: NPM install
               uses: actions/setup-node@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/calculate-difference.yml
```diff
@@ -0,0 +1,51 @@
...
+              with:
+                  node-version: 16.x
+
+            - name: Output Sizes
+              run: |
+                  echo Initial Size: ${{ inputs.initial_bundle_size }}
+                  echo Updated Size: ${{ inputs.updated_bundle_size }}
+
+            - name: Report Diff
+              id: get-diff
```
rmi22186, line 41: Should this be `report-diff` instead?

.github/workflows/calculate-difference.yml
```diff
@@ -0,0 +1,51 @@
...
+              run: |
+                  div=$(npx mathjs `echo ${{ inputs.updated_bundle_size }}` / `echo ${{ inputs.initial_bundle_size }}` )
+                  diff=$(npx mathjs `echo $div` -1)
+                  result=$(npx mathjs `echo $diff` / .01)
+                  result_percent=$(npx mathjs "floor($result, 2)")
+                  echo "::set-output name=diff::$result"
```
rmi22186, line 47: I don't see `diff` being used anywhere below here, so I don't think re-setting it is necessary.
```suggestion
```

rmi22186 (changes requested):

.github/workflows/calculate-difference.yml
alexs-mparticle: Both `diff` and `diff_percent` are registered as outputs on lines 14 through 20. I'm exposing both in case we ever need either one.

rmi22186: This is less of an issue than I had with unbundled and bundled, so that's fine.

rmi22186 (approved):

PR #334, opened 2022-08-16
refactor: Migrate Store to TypeScript

alexs-mparticle:
## Summary
- Migrate Store Module to TypeScript

## Testing Plan
- Run automated tests
- Manually test in a sample app to verify store builds correctly

## Master Issue
Closes https://go.mparticle.com/work/SQDSDKS-4328

### DIFF
--- a/src/sdkRuntimeModels.ts
+++ b/src/sdkRuntimeModels.ts
@@ -1,6 +1,6 @@
 import { DataPlanVersion } from '@mparticle/data-planning-models';
 import { Batch } from '@mparticle/event-models';
-import { MPConfiguration } from '@mparticle/web-sdk';
+import { MPConfiguration, SDKEventCustomFlags } from '@mparticle/web-sdk';
 import { SDKConfig } from './store';
 import { Dictionary } from './utils';
 import Validators from './validators';
@@ -135,7 +135,8 @@ export interface MParticleWebSDK {
     logEvent(
         eventName: string,
         eventType?: number,
-        attrs?: { [key: string]: string }
+        attrs?: { [key: string]: string },
+        customFlags?: SDKEventCustomFlags
     ): void;
     logBaseEvent(event: any): void;
     eCommerce: any;
--- a/src/sdkRuntimeModels.ts
+++ b/src/sdkRuntimeModels.ts
@@ -1,11 +1,7 @@
-import * as EventsApi from '@mparticle/event-models';
 import { DataPlanVersion } from '@mparticle/data-planning-models';
-import {
-    IdentifyRequest,
-    IdentityCallback,
-    MPConfiguration,
-    SDKEventCustomFlags,
-} from '@mparticle/web-sdk';
+import { Batch } from '@mparticle/event-models';
+import { MPConfiguration } from '@mparticle/web-sdk';
+import { SDKConfig } from './store';
 import { Dictionary } from './utils';
 import Validators from './validators';
 
@@ -124,7 +120,7 @@ export interface SDKProduct {
 }
 
 export interface MParticleWebSDK {
-    addForwarder(mockForwarder: any); // TODO: Should be type MPForwarder
+    addForwarder(mockForwarder: MPForwarder): void;
     Identity: SDKIdentityApi;
     Logger: SDKLoggerApi;
     _Store: SDKStoreApi;
@@ -149,66 +145,26 @@ export interface MParticleWebSDK {
     isIOS?: boolean;
 }
 
-// TODO: Move into Store.ts
-export interface SDKConfig {
-    isDevelopmentMode?: boolean;
-    logger: {
-        error?(msg);
-        warning?(msg);
-        verbose?(msg);
-    };
-    onCreateBatch(batch: EventsApi.Batch): EventsApi.Batch;
-    customFlags?: SDKEventCustomFlags;
-    dataPlan: SDKDataPlan;
-    dataPlanOptions: KitBlockerOptions; // when the user provides their own data plan
-    dataPlanResult?: DataPlanResult; // when the data plan comes from the server via /config
-
-    appName?: string;
-    appVersion?: string;
-    package?: string;
-    flags?: { [key: string]: string | number | boolean };
-    kitConfigs: KitConfigs[];
-    kits: Dictionary<Kit>;
-    logLevel?: LogLevelType;
-    cookieDomain?: string;
-    maxCookieSize?: number | undefined;
-    minWebviewBridgeVersion: 1 | 2 | undefined;
-    identifyRequest: IdentifyRequest;
-    identityCallback: IdentityCallback;
-    integrationDelayTimeout: number;
-
-    aliasMaxWindow: number;
-    deviceId?: string;
-    forceHttps?: boolean;
-    aliasUrl?: string;
-    configUrl?: string;
-    identityUrl?: string;
-    isIOS?: boolean;
-    maxProducts: number;
-    requestConfig?: boolean;
-    sessionTimeout?: number;
-    useNativeSdk?: boolean;
-    useCookieStorage?: boolean;
-    v1SecureServiceUrl?: string;
-    v2SecureServiceUrl?: string;
-    v3SecureServiceUrl?: string;
-}
-
 export type LogLevelType = 'none' | 'verbose' | 'warning' | 'error';
 
 // TODO: Create true types for Kits and Kit Configs
 export type KitConfigs = Dictionary;
 export type Kit = Dictionary;
+export type MPForwarder = Dictionary;
 
 // TODO: This should eventually be moved into wherever init logic lives
 // TODO: Replace/Merge this with MPConfiguration in @types/mparticle__web-sdk
+// SDK Init Config represents the config that is passed into mParticle.init when
+// the sdk is initialized.
+// Currently, this extends MPConfiguration in @types/mparticle__web-sdk
+// and the two will be merged in once the Store module is refactored
 export interface SDKInitConfig
     extends Omit<MPConfiguration, 'dataPlan' | 'logLevel'> {
     dataPlan?: DataPlanConfig | KitBlockerDataPlan; // TODO: These should be eventually split into two different attributes
     logLevel?: LogLevelType;
 
     kitConfigs?: KitConfigs[];
-    kits?: Record<string, Kit>;
+    kits?: Dictionary<Kit>;
     dataPlanOptions?: KitBlockerOptions;
     flags?: Dictionary;
 
@@ -272,7 +228,7 @@ export interface SDKConfigApi {
     v3SecureServiceUrl?: string;
     isDevelopmentMode: boolean;
     appVersion?: string;
-    onCreateBatch(batch: EventsApi.Batch): EventsApi.Batch;
+    onCreateBatch(batch: Batch): Batch;
 }
 export interface MParticleUser {
     getMPID(): string;
--- a/src/sdkRuntimeModels.ts
+++ b/src/sdkRuntimeModels.ts
@@ -1,5 +1,13 @@
 import * as EventsApi from '@mparticle/event-models';
-import { DataPlanVersion, DataPlan } from '@mparticle/data-planning-models';
+import { DataPlanVersion } from '@mparticle/data-planning-models';
+import {
+    IdentifyRequest,
+    IdentityCallback,
+    MPConfiguration,
+    SDKEventCustomFlags,
+} from '@mparticle/web-sdk';
+import { Dictionary } from './utils';
+import Validators from './validators';
 
 export interface SDKEvent {
     DeviceId: string;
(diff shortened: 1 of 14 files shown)


src/store.ts
```diff
@@ -200,10 +263,13 @@ export default function Store(config, mpInstance) {
...
             };
-            if (config.dataPlan.hasOwnProperty('planId')) {
-                if (typeof config.dataPlan.planId === 'string') {
-                    if (mpInstance._Helpers.isSlug(config.dataPlan.planId)) {
-                        this.SDKConfig.dataPlan.PlanId = config.dataPlan.planId;
+
+            // TODO: Can we refactor to avoid unnecessary type checking
+            const dataPlan = config.dataPlan as DataPlanConfig;
+            if (dataPlan.hasOwnProperty('planId')) {
+                if (typeof dataPlan.planId === 'string') {
```
alexs-mparticle, line 270: Note to self: Can we use a helper for this?

src/utils.ts
```diff
@@ -70,6 +75,8 @@ const converted = (s: string): string => {
     return s;
 };
 
+const isSlug = (str: string) => str === slugify(str);
```
alexs-mparticle, line 78: Note: Add return type of `boolean` and maybe add `isString` functionality?

src/sdkRuntimeModels.ts
```diff
@@ -116,59 +124,116 @@ export interface SDKProduct {
...
+
+export type LogLevelType = 'none' | 'verbose' | 'warning' | 'error';
+
+// TODO: Create true types for Kits and Kit Configs
+export type KitConfigs = Dictionary;
+export type Kit = Dictionary;
+
+// TODO: This should eventually be moved into wherever init logic lives
+// TODO: Replace/Merge this with MPConfiguration in @types/mparticle__web-sdk
+export interface SDKInitConfig
```
alexs-mparticle, line 205: Note to self: Explain why this is different than SDKConfig and MPConfiguration

src/sdkRuntimeModels.ts
```diff
@@ -116,59 +124,116 @@ export interface SDKProduct {
...
+    extends Omit<MPConfiguration, 'dataPlan' | 'logLevel'> {
+    dataPlan?: DataPlanConfig | KitBlockerDataPlan; // TODO: These should be eventually split into two different attributes
+    logLevel?: LogLevelType;
+
+    kitConfigs?: KitConfigs[];
+    kits?: Record<string, Kit>;
```
alexs-mparticle, line 211:
```suggestion
    kits?: Dictionary<Kit>;
```

src/utils.ts
alexs-mparticle: @rmi22186 I'm torn on this. Do we explicitly need to check if something is both a slug and a string, or can we just simplify things and force slug and string checks in one place?

src/store.ts
```diff
@@ -200,10 +322,13 @@ export default function Store(config, mpInstance) {
                 PlanVersion: null,
                 PlanId: null,
             };
-            if (config.dataPlan.hasOwnProperty('planId')) {
-                if (typeof config.dataPlan.planId === 'string') {
-                    if (mpInstance._Helpers.isSlug(config.dataPlan.planId)) {
-                        this.SDKConfig.dataPlan.PlanId = config.dataPlan.planId;
+
+            // TODO: Can we refactor to avoid unnecessary type checking
```
rmi22186, line 326: I'd say yes, now that these are typed

src/utils.ts
rmi22186: Are these comments outdated? you have `isString` on line 78 and a separate `isSlug` method with no `isString` check inside of it, so I'm a bit confused.

rmi22186: if it matters, when you call `slugify(8)`, it throws an error as opposed to logging an error

test/src/tests-batchUploader.ts
```diff
@@ -249,7 +248,7 @@ describe('batch uploader', () => {
             window.mParticle.config.identityCallback = function(result) {
                 let currentUser = result.getUser()
                 if (currentUser) {
-                    currentUser.setUserAttribute("number", Math.floor((Math.random() * 1000) + 1))
+                    currentUser.setUserAttribute("number", `${Math.floor((Math.random() * 1000) + 1)}`)
```
rmi22186, line 251: not sure why this was changed or if some test was failing, but I don't think this is necessary. I think the idea was we are setting a user attribute to an actual number, not a string of a number

test/src/tests-core-sdk.js
```diff
@@ -924,6 +923,7 @@ describe('core SDK', function() {
         'minWebviewBridgeVersion',
         'aliasMaxWindow'
     ];
+    // TODO: Refactor this test to respect type safety or remove as this duplicates tests in store.ts
```
rmi22186, line 926: I'm fine with that.  I do remember writing a test and I suppose I put it in here instead of a `test-store.js` because I figured it was core to the SDK, and it was a single test.  since this is more an integration test I think we should move it over and label it as such.

test/src/tests-store.ts
```diff
@@ -0,0 +1,139 @@
...
+        window.mParticle.init(apiKey, window.mParticle.config);
+    });
+
+    afterEach(function() {
+        window.mParticle._resetForTests(MPConfig);
+        sandbox.restore();
+        clock.restore();
+    });
+
+	it('should initialize with defaults', () => {
```
rmi22186, line 31: indentation is wonky here.  remove 1 tab.  hard to tell if my suggestion is doing that properly, so I'll leave a comment instead of a suggestion

test/src/tests-store.ts
```diff
@@ -0,0 +1,139 @@
...
+            expect(store.SDKConfig.minWebviewBridgeVersion, 'minWebviewBridgeVersion').to.eq(1);
+            expect(store.SDKConfig.isIOS, 'isIOS').to.eq(false);// ?: boolean;
+            expect(store.SDKConfig.integrationDelayTimeout, 'integrationDelayTimeout').to.eq(5000);
+            expect(store.SDKConfig.aliasUrl, 'aliasUrl').to.eq('jssdks.mparticle.com/v1/identity/');
+            expect(store.SDKConfig.configUrl, 'configUrl').to.eq('jssdkcdns.mparticle.com/JS/v2/');
+            expect(store.SDKConfig.identityUrl, 'identityUrl').to.eq('identity.mparticle.com/v1/');
+
+            expect(store.SDKConfig.v1SecureServiceUrl, 'v1SecureServiceUrl').to.eq('jssdks.mparticle.com/v1/JS/');// ?: string;
+            expect(store.SDKConfig.v2SecureServiceUrl, 'v2SecureServiceUrl').to.eq('jssdks.mparticle.com/v2/JS/');// ?: string;
+            expect(store.SDKConfig.v3SecureServiceUrl, 'v3SecureServiceUrl').to.eq('jssdks.mparticle.com/v3/JS/');// ?: string;
```
rmi22186, line 109: can you clarify the comments here?

rmi22186 (changes requested):

test/src/tests-batchUploader.ts
rmi22186: Ah, very interesting. Seems like from the [docs](https://docs.mparticle.com/developers/sdk/web/users/#set-user-attributes) for `setUserAttribute` we allow `string` and `null`, and in the case of `setUserAttributeList` we allow `array`, but the `setUserAttribute` code shows we run this [validation code](https://github.com/mParticle/mparticle-web-sdk/blob/master/src/validators.ts#L15-L17), which _does allow_ `numbers` and `booleans in addition to `strings`.  I think for now this is fine, and I'm going to make a ticket to sort out how this works on the other platforms.

src/utils.ts
```diff
@@ -70,6 +75,17 @@ const converted = (s: string): string => {
     return s;
 };
 
+const isString = (value: any): boolean => typeof value === 'string';
+const isNumber = (value: any): boolean => typeof value === 'number';
+const isFunction = (fn: any): boolean => typeof fn === 'function';
+
+// TODO: Refactor this to a regex
+const isDataPlanSlug = (str: string): boolean =>
+    isString(str) && str === slugify(str, { replacement: '_', lower: true });
```
rmi22186, line 84: Why are we adding `replacement` here?  If they pass a string to us with spaces, we should be denying it, so I'm unclear how your tests with the spaces are passing. I must be misled?

rmi22186 (changes requested):

test/src/tests-batchUploader.ts
jared-dickman: i dont have much context here but setting a user attribute to a `number` seems like a valid case. if it is, then the type is wrong. if its not, then the test is _wrong_

rmi22186: Yep it absolutely is from a user perspective. I'll add some more context based on my research yesterday.  I added a sustained engineering issue yesterday at go.mparticle.om/SQDSDKS-4576 to track this and summarize the above. I should have referenced it in my comment.

When the client sends a number to the server, the server casts it to a `string` for some reason, which is probably why I _think_ iOS/Android only allows `string`/`null` to not mislead the customer into thinking it's saved as a `number`.  Our SDK does allow numbers (but our types in definitely typed _don't_), so that's causing the issue here. 

We should _probably_ match iOS/Android, and decide what to do about definitely typed. Since DT is potentially wrong, we should match it, but perhaps add a comment to the go link above.
```suggestion
                    // TODO: Investigate if we should update definitely typed typings for setUserAttribute which only allows strings right now - more context at go.mparticle.om/SQDSDKS-4576
                    currentUser.setUserAttribute("number", `${Math.floor((Math.random() * 1000) + 1)}`)
```

src/sdkRuntimeModels.ts
```diff
@@ -116,59 +120,76 @@ export interface SDKProduct {
...
     getInstance();
     ServerModel();
     upload();
     setPosition(lat: number | string, lng: number | string): void;
-    logEvent(eventName: string, eventType?: number, attrs?: { [key: string]: string }): void;
+    logEvent(
+        eventName: string,
+        eventType?: number,
+        attrs?: { [key: string]: string }
+    ): void;
```
rmi22186, line 139: this should have customFlags as a 4th argument

```suggestion
        attrs?: { [key: string]: string },
        customFlags?: { [key: string]: string }
    ): void;
```

src/store.ts
```diff
@@ -113,14 +240,9 @@ export default function Store(config, mpInstance) {
...
-            this.SDKConfig.useNativeSdk = config.useNativeSdk;
-        } else {
-            this.SDKConfig.useNativeSdk = false;
-        }
-        if (config.hasOwnProperty('kits')) {
-            this.SDKConfig.kits = config.kits;
-        }
+        this.SDKConfig.useNativeSdk = !!config.useNativeSdk;
+
+        this.SDKConfig.kits = config.kits || {};
```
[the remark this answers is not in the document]
rmi22186: ignore this.  I conflated `kits` and `kitConfig`.  However, let me keep doing some digging to convince myself I am 100% on these types.

rmi22186 (changes requested):

src/store.ts
```diff
@@ -1,30 +1,157 @@
...
+    MPForwarder,
+    SDKConsentState,
+    SDKDataPlan,
+    SDKEvent,
+    SDKGeoLocation,
+    SDKInitConfig,
+    SDKProduct,
+} from './sdkRuntimeModels';
+import { isNumber, isDataPlanSlug } from './utils';
+import { Dictionary } from './utils';
```
rmi22186, line 27:
```suggestion
import { isNumber, isDataPlanSlug, Dictionary } from './utils';
```

src/store.ts
```diff
@@ -1,30 +1,157 @@
...
+// This represents the runtime configuration of the SDK AFTER
+// initialization has been complete and all settings and
+// configurations have been stitched together.
+export interface SDKConfig {
+    isDevelopmentMode?: boolean;
+    logger: {
+        error?(msg);
+        warning?(msg);
+        verbose?(msg);
+    };
```
rmi22186, line 38:
```suggestion
        error?:(msg: string) => void;
        warning?:(msg: string)=> void;
        verbose?:(msg: string)=> void;
    };
```

src/store.ts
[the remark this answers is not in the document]
alexs-mparticle: Ignoring this for now

src/utils.ts
alexs-mparticle: I think I was doing that to better support testing for kabob case, but it doesn't make a difference. I can remove.

test/src/tests-utils.ts
```diff
@@ -99,4 +108,35 @@ describe('Utils', () => {
...
+        it('handles numbers', function () {
+            expect(isDataPlanSlug(35 as unknown as string)).to.equal(false);
+        });
+
+        it('handles spaces', function () {
+            expect(isDataPlanSlug('Slug with spaces in')).to.equal(false);
+        });
+
+        it('handles PascalCase', function () {
+            expect(isDataPlanSlug('PascalSlug')).to.equal(false);
```
rmi22186, line 122: add a similar TODO: from line 126 for pascal case

test/src/tests-store.ts
```diff
@@ -0,0 +1,190 @@
...
+        window.mParticle.init(apiKey, window.mParticle.config);
+    });
+
+    afterEach(function() {
+        window.mParticle._resetForTests(MPConfig);
+        sandbox.restore();
+        clock.restore();
+    });
+
+    it('should initialize with defaults', () => {
```
rmi22186, line 31:
```suggestion
    it('should initialize store with defaults', () => {
```

test/src/tests-store.ts
```diff
@@ -0,0 +1,190 @@
...
+            0
+        );
+        expect(
+            store.integrationDelayTimeoutStart,
+            'integrationDelayTimeoutStart'
+        ).to.eq(clock.now);
+    });
+
+    describe('#createStore', () => {
+        it('should set valid defaults', () => {
```
rmi22186, line 87:
```suggestion
        it('should initialize store.SDKConfig with valid defaults', () => {
```

test/src/tests-store.ts
```diff
@@ -0,0 +1,190 @@
...
+        expect(store.pixelConfigurations.length, 'pixelConfigurations').to.eq(
+            0
+        );
+        expect(
+            store.integrationDelayTimeoutStart,
+            'integrationDelayTimeoutStart'
+        ).to.eq(clock.now);
+    });
+
+    describe('#createStore', () => {
```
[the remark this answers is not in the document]
alexs-mparticle: I think that's a mistake on my part. Probably should be `#createSDKConfig`

rmi22186 (approved):

PR #342, opened 2022-08-23
refactor: Migrate Consent to TypeScript

alexs-mparticle:
## Instructions

## Summary
- Refactors Consent Module to use TypeScript Interfaces

 ## Testing Plan

## Master Issue
Closes https://go.mparticle.com/work/SQDSDKS-4465


src/consent.ts
```diff
@@ -258,176 +264,233 @@ export default function Consent(this: IConsent, mpInstance: MParticleWebSDK) {
...
-                return null;
-            }
-            var trimmedPurpose = purpose.trim();
-            if (!trimmedPurpose.length) {
-                return null;
-            }
-            return trimmedPurpose.toLowerCase();
-        }
+// TODO: Fix collisison between this and ConsentState
+class ConsentStateOBJ {
```
jared-dickman, line 291: is this like a `ConsentStateModel` and the other is more of a `ConsentStateFactory`??

src/consent.ts
```diff
@@ -0,0 +1,496 @@
...
+export interface IConsentSerialization {
+    toMinifiedJsonObject: (state: ConsentState) => ConsentJSONObject;
+    fromMinifiedJsonObject: (json: ConsentJSONObject) => ConsentState;
+}
+
+export interface IConsent {
+    mPinstance;
+    isEnabledForUserConsent: (consentRules: any, user: any) => boolean;
+    createPrivacyConsent: (
+        consented: boolean,
```
jared-dickman, line 41: we should extract these 5 properties into an interface, its reused in several other places, ~lines 160 and 360 in this file [consent.ts] for example

src/consent.ts
```diff
@@ -1,9 +1,131 @@
...
+
+export interface IConsentSerialization {
+    toMinifiedJsonObject: (state: ConsentState) => IMinifiedConsentJSONObject;
+    fromMinifiedJsonObject: (json: IMinifiedConsentJSONObject) => ConsentState;
+}
+
+// TODO: Resolve discrepency between ConsentState and SDKConsentState
+//       https://go.mparticle.com/work/SQDSDKS-5009
+export interface SDKConsentState
+    extends Omit<ConsentState, 'getGDPRConsentState' | 'getCCPAConsentState'> {
```
jared-dickman, line 48: nice use of omit!

src/consent.ts
```diff
@@ -1,9 +1,131 @@
...
+    getGDPRConsentState(): SDKGDPRConsentState;
+    getCCPAConsentState(): SDKCCPAConsentState;
+}
+
+export interface SDKGDPRConsentState {
+    [key: string]: SDKConsentStateData;
```
jared-dickman, line 54: [key: string] can probably be replaced by a dictionary

src/consent.ts
```diff
@@ -1,9 +1,131 @@
...
+    ccpa?: ICCPAConsentStateDTO;
+}
+
+export interface IConsentRuleValues {
+    consentPurpose: string;
+    hasConsented: boolean;
+}
+export interface IConsentRules {
+    includeOnMatch: boolean;
+    values: IConsentRuleValues[];
```
jared-dickman, line 100: should this be `IConsentRulesValues` with an extra s, to match `IConsentRules`

src/consent.ts
```diff
@@ -1,9 +1,131 @@
...
+// TODO: Remove this if we can safely deprecate `removeCCPAState`
+export interface IConsentState extends ConsentState {
+    removeCCPAState: () => ConsentState;
+}
+
+// Represents Actual Interface for Consent Module
+// TODO: Should eventually consolidate with SDKConsentStateApi
+export interface IConsent {
+    mPinstance;
+    // TODO: Define these ANYs
```
jared-dickman, line 112: is this something that can be done now?

src/consent.ts
```diff
@@ -15,20 +137,21 @@ export default function Consent(mpInstance) {
...
+
         if (consentState) {
             // the server hashes consent purposes in the following way:
             // GDPR - '1' + purpose name
             // CCPA - '2data_sale_opt_out' (there is only 1 purpose of data_sale_opt_out for CCPA)
-            var GDPRConsentHashPrefix = '1';
-            var CCPAPurpose = 'data_sale_opt_out';
-            var CCPAHashString = '2' + CCPAPurpose;
+            const GDPRConsentHashPrefix = '1';
+            const CCPAPurpose = 'data_sale_opt_out'; // FIXME: is this redundant?
```
jared-dickman, line 149: yes it is

src/consent.ts
```diff
@@ -303,8 +457,11 @@ export default function Consent(mpInstance) {
          * @return {Object} A GDPR Consent State
          */
 
-        function getGDPRConsentState() {
-            return mpInstance._Helpers.extend({}, gdpr);
+        function getGDPRConsentState(): GDPRConsentState {
+            // TODO: Can we replace the usage of extend and either use
```
jared-dickman, line 461: yes i think so

jared-dickman: assuming there are no subtle gotchas with `mpInstance._Helpers.extend`

src/sdkRuntimeModels.ts
```diff
@@ -6,7 +6,7 @@ import Validators from './validators';
 import { Dictionary } from './utils';
 import { IServerModel } from './serverModel';
 import { IKitConfigs } from './configAPIClient';
-import { SDKConsentApi, SDKConsentState } from './consent.interfaces';
+import { IConsent, SDKConsentApi, SDKConsentState } from './consent';
```
rmi22186, line 9: This can be removed - it is not being used in the file

test/src/tests-consent.ts
```diff
@@ -710,6 +710,7 @@ describe('Consent', function() {
         done();
     });
 
+    // TODO: Can we remove this now?
```
rmi22186, line 713: Technically it's still in the API and we never gave notice.  Let's flag as a TODO: Remove in next major version.  As part of modernization.

src/consent.ts
```diff
@@ -1,9 +1,131 @@
...
+    ConsentState,
+    GDPRConsentState,
+    PrivacyConsentState,
+} from '@mparticle/web-sdk';
+import { MParticleUser, MParticleWebSDK } from './sdkRuntimeModels';
+import { Dictionary, isObject } from './utils';
 
-    // this function is called when consent is required to determine if a cookie sync should happen, or a forwarder should be initialized
-    this.isEnabledForUserConsent = function(consentRules, user) {
+export const CCPAPurpose = 'data_sale_opt_out' as const;
```
rmi22186, line 10: Is the as const necessary?

src/consent.ts
```diff
@@ -1,9 +1,131 @@
...
+}
+
+export interface ICCPAConsentStateDTO {
+    data_sale_opt_out: IPrivacyV2DTO;
+}
+
+export interface IConsentStateDTO {
+    gdpr?: IGDPRConsentStateDTO;
+    ccpa?: ICCPAConsentStateDTO;
+}
```
rmi22186, line 92: Nit - I would label anything with `DTO` in it as `V2DTO` instead like you have in `IPrivacyV2DTO`

It's more clear this way, and also when the V2 DTO is removed, it'll be easier to identify which interfaces can also be removed.

src/consent.ts
```diff
@@ -1,9 +1,131 @@
...
+
+// TODO: Remove this if we can safely deprecate `removeCCPAState`
+export interface IConsentState extends ConsentState {
+    removeCCPAState: () => ConsentState;
+}
+
+// Represents Actual Interface for Consent Module
+// TODO: Should eventually consolidate with SDKConsentStateApi
+export interface IConsent {
+    mPinstance;
```
rmi22186, line 111: Can this be removed? I don't believe Consent should have `mpInstance` as a member.  I caught this because `mPinstance` here has different casing than normal (lowercase vs capital i)

src/consent.ts
```diff
@@ -170,12 +295,12 @@ export default function Consent(mpInstance) {
             return jsonObject;
         },
 
-        fromMinifiedJsonObject: function(json) {
-            var state = self.createConsentState();
+        fromMinifiedJsonObject(json: IMinifiedConsentJSONObject): ConsentState {
+            const state = self.createConsentState() as ConsentState;
```
rmi22186, line 299:
```suggestion
            const state: ConsentState = self.createConsentState();
```

rmi22186: nit, but i think it's better practice to put the type beforehand unless it is required at the end like in line 240.

src/consent.ts
rmi22186: yep

src/consent.ts
```diff
@@ -203,30 +328,45 @@ export default function Consent(mpInstance) {
...
 
-    this.createConsentState = function(consentState) {
-        var gdpr = {};
-        var ccpa = {};
+    // TODO: Refactor this method into a constructor
+    this.createConsentState = function (
+        this: ConsentState,
+        consentState?: ConsentState
+    ): IConsentState {
+        let gdpr = {};
```
rmi22186, line 336: Perhaps we should have a note of what the difference between `IConsentState` and `ConsentState` to help out future Alex and Rob.  I see that one has the deprecated method but can't quite figure out when you decide to use `ConsentState` vs `IConsentState`

src/consent.ts
rmi22186: yes we can.  Helpers.extend should just have been a polyfill, so now that this is being migrated to TS, Object.assign works

rmi22186 (changes requested):

src/consent.ts
alexs-mparticle: I kind of is. the `as const` tells the compiler that this value is `readonly` and lets us use it in line 15 without issues.

rmi22186: maybe i should have been more clear - is saying **const** CCPAPurpose = .... **as const** redundant?

src/consent.ts
alexs-mparticle: Good catch

src/consent.ts
alexs-mparticle: It looks like that, but `const` at the beginning of the statement versus at the end signify two different things.

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions

src/consent.ts
alexs-mparticle: The difference is that `ConsentState` is defined in DefTyped and does not have `removeCCPAState`, but `IConsentState` is a local override to keep `removeCCPAState` since we have not properly deprecated the method.

alexs-mparticle: See line 99 of src/consent.ts in this PR.

https://github.com/mParticle/mparticle-web-sdk/pull/342/files#diff-2fe6c689e234643987ca8ef0b2c744e50bc2ab8ab16fd454b31a155fc204abaeR99

rmi22186: makes sense

rmi22186 (approved):
Approving, but I'd add a TODO to figure out why `as const` is actually needed.

rmi22186:
@alexs-mparticle This is good to merge now that we didn't do a release this morning for Braze.

PR #345, opened 2022-08-24
refactor: Migrate ServerModel to TypeScript

alexs-mparticle:
## Summary
- Migrate Server Models Module to TypeScript
- Define Interfaces for Server DTOs

## Testing Plan
- Automated Tests should continue to pass

## Master Issue
Closes https://go.mparticle.com/work/SQDSDKS-4476


src/serverModel.ts
```diff
@@ -1,47 +1,219 @@
...
+}
+
+export interface UploadObject extends SDKEvent {
+    ClientGeneratedId: string;
+    Store: IStore;
+    ExpandedEventCount: number;
+    ProfileMessageType: number;
+}
+
+export interface IServerModel {
```
jared-dickman, line 143: do we have rules for using `I` as a prefix for interfaces vs not??

I think there can be readability benefits from being consistent

src/store.ts
```diff
@@ -115,8 +115,8 @@ export interface IStore {
     devToken: string | null;
     migrationData: MigrationData;
     serverSettings: ServerSettings;
-    dateLastEventSent: number | null;
-    sessionStartDate: number | null;
+    dateLastEventSent: Date | null; // TODO: Number or Date?
+    sessionStartDate: Date | null; // TODO: Number or Date?
```
rmi22186, line 119: You're right, this is `Date`

src/serverModel.ts
alexs-mparticle: I'm fine with using `I` as a prefix

src/serverModel.ts
```diff
@@ -1,47 +1,222 @@
...
+    var valueArray: string[] = [];
+    dto.flags = {};
+
+    for (var prop in event.CustomFlags) {
+        valueArray = [];
+
+        if (event.CustomFlags.hasOwnProperty(prop)) {
+            if (Array.isArray(event.CustomFlags[prop])) {
+                event.CustomFlags[prop].forEach(function(customFlagProperty) {
+                    // TODO: Can we use our utility functions here?
```
rmi22186, line 161: yes

src/serverModel.ts
```diff
@@ -1,47 +1,222 @@
...
-                    typeof event.CustomFlags[prop] === 'string' ||
-                    typeof event.CustomFlags[prop] === 'boolean'
-                ) {
-                    valueArray.push(event.CustomFlags[prop].toString());
-                }
+                        valueArray.push(customFlagProperty.toString());
+                    }
+                });
+            } else if (
+                // TODO: Can we use our utility functions here?
```
rmi22186, line 171: yes

src/serverModel.ts
```diff
@@ -1,47 +1,222 @@
...
+}
 
-    this.appendUserInfo = function(user, event) {
+export default function ServerModel(
+    this: IServerModel,
+    mpInstance: MParticleWebSDK
+) {
+    var self = this;
+
+    // TODO: Can we refactor this to not mutate the event?
```
rmi22186, line 218: yes

src/serverModel.ts
```diff
@@ -153,28 +304,37 @@ export default function ServerModel(mpInstance) {
...
-        var eventObject = {};
+    this.createEventObject = function(
+        event: BaseEvent,
+        user?: MParticleUser
+    ): SDKEvent | IUploadObject {
+        var uploadObject: Partial<IUploadObject> = {};
+        var eventObject: Partial<SDKEvent> = {};
 
+        // TODO: What is the purpose of setting OptOut to the inverse of the Store's
+        //       isEnabled property and not just `false`?
```
rmi22186, line 318: There is a method, [mParticle.setOptOut(boolean)](https://github.com/mParticle/mparticle-web-sdk/blob/master/src/mp-instance.js#L1087-L1116), which accepts a boolean and that sets the `isEnabled` flag to be used elsewhere in the SDK for if it should log or not (line 321 below).  It doesn't set opt out to true all the time, which isn't clear from the code here.

src/serverModel.ts
```diff
@@ -153,28 +304,37 @@ export default function ServerModel(mpInstance) {
...
         var optOut =
             event.messageType === Types.MessageType.OptOut
                 ? !mpInstance._Store.isEnabled
                 : null;
 
+        // TODO: Why is Webview Bridge Enabled or Opt Out necessary here?
```
rmi22186, line 324: I believe if web view bridge is enabled, there may not be a session id because the event is associated with the native sdk's  session.  I would have to dig in more to remember why optout is included here.

src/serverModel.ts
```diff
@@ -153,28 +304,37 @@ export default function ServerModel(mpInstance) {
...
-            event.messageType == Types.MessageType.OptOut ||
+            event.messageType === Types.MessageType.OptOut ||
             mpInstance._Store.webviewBridgeEnabled
         ) {
             if (event.hasOwnProperty('toEventAPIObject')) {
                 eventObject = event.toEventAPIObject();
             } else {
                 eventObject = {
-                    EventName: event.name || event.messageType,
+                    // TODO: Do we need the number from the enum or the string it evaluates to?
```
rmi22186, line 334: This is a quirk of the backend I believe for V2 - it should be the number from the enum.  For v2 of the events endpoint, the name for Session Start/End, AST's are the numbers (1/2 and 10).  For V3 this resolves back to `session_start/end`, and `application_state_transition`.

src/serverModel.ts
```diff
@@ -190,12 +350,14 @@ export default function ServerModel(mpInstance) {
...
 
+            // TODO: Should we move this side effect outside of this method?
             if (event.messageType !== Types.MessageType.SessionEnd) {
                 mpInstance._Store.dateLastEventSent = new Date();
             }
 
             uploadObject = {
-                Store: mpInstance._Store.serverSettings,
+                // TODO: Why are we passing Server Settings as Store?
+                Store: (mpInstance._Store.serverSettings as unknown) as IStore,
```
rmi22186, line 360: This is definitely misleading.  The  `Store`  on the V2 SDK is something the server was expecting, and it is NOT the `IStore` which contains all the configuration and other SDK attributes.

rmi22186: This is removed in V3.  I never knew what the server did with this.

src/serverModel.ts
```diff
@@ -222,19 +384,23 @@ export default function ServerModel(mpInstance) {
...
+            self.appendUserInfo(currentUser, eventObject as SDKEvent);
 
             if (event.messageType === Types.MessageType.SessionEnd) {
                 eventObject.SessionLength =
                     mpInstance._Store.dateLastEventSent.getTime() -
                     mpInstance._Store.sessionStartDate.getTime();
                 eventObject.currentSessionMPIDs =
                     mpInstance._Store.currentSessionMPIDs;
+
+                // TODO: Why are we using session attributes instead of the event's Event Attributes?
```
rmi22186, line 399: Will have to experiment.  This might be a bug

src/serverModel.ts
```diff
@@ -222,19 +384,23 @@ export default function ServerModel(mpInstance) {
...
                 eventObject.EventAttributes =
                     mpInstance._Store.sessionAttributes;
 
+                // TODO: We should move this out of here to avoid side effects
```
rmi22186, line 403: 👍

src/serverModel.ts
rmi22186: Ah. this is because it is inside of a session end event.  For a session end event, you cannot set an event attribute, it is simply:

mParticle.endSession().  

So we take the session attributes and turn them into event attributes for this.  Based on my quick search of the code, when you set a session attribute, it isn't actually ever sent with anything, and only with the session end event.

src/sdkRuntimeModels.ts
```diff
@@ -123,12 +129,18 @@ export interface MParticleWebSDK {
...
-    _resetForTests(MPConfig: SDKConfig): void;
+    _ServerModel: IServerModel;
+    _SessionManager: any; // TODO: Set up Session Manager
+    _Consent: any; // TODO: Set up Consent SDK
+    _resetForTests(MPConfig?: SDKConfig): void;
     init(apiKey: string, config: SDKInitConfig, instanceName?: string): void;
-    getInstance();
+    getAppName(): string;
+    getAppVersion(): string;
+    getInstance(): MParticleWebSDK;
```
alexs-mparticle, line 143:
```suggestion
    getInstance(): MParticleWebSDK; // TODO: Create a new type for MParticleWebSDKInstance
```

src/serverModel.ts
alexs-mparticle: Set up for future work here: https://go.mparticle.com/work/SQDSDKS-4813

src/serverModel.ts
```diff
@@ -153,28 +305,37 @@ export default function ServerModel(mpInstance) {
...
-        var eventObject = {};
+    this.createEventObject = function(
+        event: BaseEvent,
+        user?: MParticleUser
+    ): SDKEvent | IUploadObject {
+        var uploadObject: Partial<IUploadObject> = {};
+        var eventObject: Partial<SDKEvent> = {};
 
+        // TODO: What is the purpose of setting OptOut to the inverse of the Store's
+        //       isEnabled property and not just `false`?
```
alexs-mparticle, line 319:
```suggestion
        //  The `optOut` variable is passed later in this method to the `uploadObject`
        //  so that it can be used to denote whether or not a user has "opted out" of being
        //  tracked. If this is an `optOut` Event, we set `optOut` to the inverse of the SDK's 
        // `isEnabled` boolean which is controlled via `MPInstanceManager.setOptOut`.
```

src/serverModel.ts
```diff
@@ -153,28 +305,39 @@ export default function ServerModel(mpInstance) {
...
-            event.messageType == Types.MessageType.OptOut ||
+            event.messageType === Types.MessageType.OptOut ||
             mpInstance._Store.webviewBridgeEnabled
         ) {
             if (event.hasOwnProperty('toEventAPIObject')) {
                 eventObject = event.toEventAPIObject();
             } else {
                 eventObject = {
-                    EventName: event.name || event.messageType,
+                    // TODO: Do we need the number from the enum or the string it evaluates to?
```
alexs-mparticle, line 337:
```suggestion
                    // This is an artifact from v2 events where SessionStart/End and AST event
                    //  names are numbers (1, 2, or 10), but going forward with v3, these lifecycle 
                    //  events do not have names, but are denoted by their `event_type`
```

src/serverModel.ts
```diff
@@ -190,12 +355,14 @@ export default function ServerModel(mpInstance) {
...
             }
 
+            // TODO: Should we move this side effect outside of this method?
             if (event.messageType !== Types.MessageType.SessionEnd) {
                 mpInstance._Store.dateLastEventSent = new Date();
             }
 
             uploadObject = {
-                Store: mpInstance._Store.serverSettings,
+                // TODO: Why are we passing Server Settings as Store?
```
rmi22186, line 364:
```suggestion
                // FIXME: Deprecate when we get rid of V2
```

src/serverModel.ts
alexs-mparticle: Note to self: Make this an any or empty object for now

src/serverModel.ts
```diff
@@ -222,19 +384,23 @@ export default function ServerModel(mpInstance) {
                 eventObject.LaunchReferral = window.location.href || null;
             }
 
+            // TODO: why is this duplicated with `uploadObject`?
```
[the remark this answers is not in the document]
alexs-mparticle: OK, will remove the lower one

src/serverModel.ts
```diff
@@ -222,19 +389,23 @@ export default function ServerModel(mpInstance) {
...
+            self.appendUserInfo(currentUser, eventObject as SDKEvent);
 
             if (event.messageType === Types.MessageType.SessionEnd) {
                 eventObject.SessionLength =
                     mpInstance._Store.dateLastEventSent.getTime() -
                     mpInstance._Store.sessionStartDate.getTime();
                 eventObject.currentSessionMPIDs =
                     mpInstance._Store.currentSessionMPIDs;
+
+                // TODO: Why are we using session attributes instead of the event's Event Attributes?
```
alexs-mparticle, line 404:
```suggestion
                // Session attributes are assigned on a session level, but only uploaded
                // when a session ends. As there is no way to attach event attributes to
                // a `SessionEnd` event, we are uploading the session level attributes
                // as event level attributes in a `SessionEnd` event.
```

src/serverModel.ts
```diff
@@ -247,8 +421,9 @@ export default function ServerModel(mpInstance) {
         return null;
     };
 
-    this.convertEventToDTO = function(event) {
-        var dto = {
+    // TODO: Should this take an event or upload object?
```
alexs-mparticle, line 424:
```suggestion

```

src/serverModel.ts
alexs-mparticle: Tracking here: https://go.mparticle.com/work/SQDSDKS-4815

src/serverModel.ts
alexs-mparticle: Tracking here: https://go.mparticle.com/work/SQDSDKS-4815

src/serverModel.ts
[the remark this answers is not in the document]
alexs-mparticle: Tracking as part of: https://go.mparticle.com/work/SQDSDKS-4816

test/src/tests-serverModel.ts
```diff
@@ -0,0 +1,1675 @@
...
+                },
+            };
+
+            const event = {
+                MPID: '123456',
+            };
+
+            ServerModel.appendUserInfo(user, event as SDKEvent);
+
+            expect(event.MPID).to.equal('123456');
```
rmi22186, line 170: Can we add an expect for event to only have the key of `MPID` and nothing else?  Something like
```suggestion
            expect(event.MPID).to.equal('123456');
            expect(Object.keys(event).length.to.equal(1)
```

test/src/tests-serverModel.ts
```diff
@@ -0,0 +1,1675 @@
...
+        });
+
+        it('returns early if event and user are not valid', () => {
+            expect(ServerModel.appendUserInfo(null, null)).to.be.undefined;
+        });
+    });
+
+    describe('#convertToConsentStateDTO', () => {
+        it('should convert Consente State with GDPR to a DTO', () => {
```
rmi22186, line 204:
```suggestion
        it('should convert Consent State with GDPR to a DTO', () => {
```

test/src/tests-serverModel.ts
```diff
@@ -0,0 +1,1675 @@
...
+                    },
+                },
+            };
+
+            expect(ServerModel.convertToConsentStateDTO(consentState)).to.eql(
+                expectedDTO
+            );
+        });
+
+        it('should convert Consente State with CCPA to a DTO', () => {
```
rmi22186, line 236:
```suggestion
        it('should convert Consent State with CCPA to a DTO', () => {
```

test/src/tests-serverModel.ts
```diff
@@ -0,0 +1,1675 @@
...
+                    },
+                },
+            };
+
+            expect(ServerModel.convertToConsentStateDTO(consentState)).to.eql(
+                expectedDTO
+            );
+        });
+
+        it('should convert Consente State with both GDPR and CCPA to a DTO', () => {
```
rmi22186, line 266: global find and replace for `Consente` --> `Consent`

test/src/tests-serverModel.ts
```diff
@@ -0,0 +1,1675 @@
...
+                null
+            );
+            expect(actualEventObject.DataPlan, 'DataPlan').to.eql({});
+            expect(actualEventObject.Timestamp, 'Timestamp').to.equal(
+                mPStore.dateLastEventSent.getTime()
+            );
+        });
+
+        it('should create an event object with a user', () => {
+            const event: BaseEvent = {
```
rmi22186, line 477: I think we should have a TODO to pull these sample BaseEvents and sample Users into it's own file so it can be imported and more DRY.

test/src/tests-serverModel.ts
```diff
@@ -0,0 +1,1675 @@
...
+                Types.EventType.Media
+            );
+            expect(actualEventObject.EventDataType, 'EventDataType').to.equal(
+                Types.MessageType.Media
+            );
+        });
+    });
+
+    describe('#convertEventToDTO', () => {
+        it('should return defaults', () => {
```
rmi22186, line 748: I don't think this test name accurately reflects what is being tested. In the test, it is not returning default values, but instead the values you are giving it.

test/src/tests-serverModel.ts
```diff
@@ -0,0 +1,1675 @@
...
+                'test-device'
+            );
+            expect(actualDTO.mpid, 'event.MPID (mpid)').to.equal('test-mpid');
+            expect(
+                actualDTO.smpids,
+                'event.currentSessionMPIDs (smpids)'
+            ).to.eql(['test-mpids']);
+        });
+
+        it('should add data plan id to DTO', () => {
```
rmi22186, line 823:
```suggestion
        it('should add data plan id and version to DTO', () => {
```

test/src/tests-serverModel.ts
```diff
@@ -0,0 +1,1675 @@
...
+                    foo: 'bar', // TODO: test will nullify these
+                },
+            } as unknown) as IUploadObject;
+
+            const actualDTO = ServerModel.convertEventToDTO(uploadObject);
+
+            expect(actualDTO.fr, 'event.IsFirstRun (fr)').to.equal(true);
+
+            // TODO: What is IU?
+            expect(actualDTO.iu, '(iu)').to.equal(false);
```
rmi22186, line 907: V2 events endpoint abbreviation for `isUpgrade`

test/src/tests-serverModel.ts
```diff
@@ -0,0 +1,1675 @@
...
+            expect(
+                actualDTO.at,
+                'event.ApplicationTransitionType.AppInit (at)'
+            ).to.equal(Types.ApplicationTransitionType.AppInit);
+            expect(actualDTO.lr, 'event.LaunchReferral (lr)').to.equal(
+                'https://mparticle.com/test-referral'
+            );
+
+            // TODO: Why do we nullify this?
```
rmi22186, line 916: I think because there can't be any event attributes on an AST (not sure why, perhaps a server required thing).  It's unlikely that someone would ever log an AST on their own intentionally, but it's possible, via `logBaseEvent` where you can set the `MessageType` to be an AST

test/src/tests-serverModel.ts
```diff
@@ -0,0 +1,1676 @@
...
+                Types.EventType.Media
+            );
+            expect(actualEventObject.EventDataType, 'EventDataType').to.equal(
+                Types.MessageType.Media
+            );
+        });
+    });
+
+    describe('#convertEventToDTO', () => {
+        it('should return defaults', () => {
```
alexs-mparticle, line 749: Agreed. Does this make more sense: 

```suggestion
        it('should convert an event to a DTO', () => {
```

test/src/tests-serverModel.ts
alexs-mparticle: Good catch

test/src/tests-serverModel.ts
alexs-mparticle: Cool. I'll add a comment for posterity.

test/src/tests-serverModel.ts
```diff
@@ -0,0 +1,1675 @@
...
+            expect(actualDTO.pi[0].pil, 'ProductImpressionList').to.equal(
+                'test-product-impression'
+            );
+            expect(actualDTO.pi[0].pl, 'ProductList').to.eql(expectedProducts);
+        });
+
+        it('should add profile to DTO', () => {
+            const uploadObject = ({
+                EventDataType: Types.MessageType.Profile,
+                ProfileMessageType: 'foo-message-type',
```
rmi22186, line 1285: I didn't recall what `ProfileMessageType` was, but after searching, it [looks like it should be a number](https://github.com/mParticle/mparticle-web-sdk/blob/b3ac3630e998e8e192c3d6db5fe47c01de909fa3/src/serverModel.ts#L148), but by setting this `as unknown` below it is not typing it properly.  By assigning it this way, 
```const uploadObject: IUploadObject = ({ ```
you'd have to ensure all the properties existed on this object, so I'm not sure what the best approach is.

File: test/src/tests-serverModel.ts [hunk 30]
```diff
@@ -0,0 +1,1677 @@
...
+            expect(sdkEvent).to.be.ok;
+            expect(sdkEvent.UserIdentities).to.eql(expectedUserIdentities);
+            expect(sdkEvent.MPID).to.equal('98765');
+            expect(sdkEvent.UserAttributes).to.eql(expectedUserAttributes);
+            expect(sdkEvent.ConsentState).to.be.ok;
+
+            done();
+        });
+
+        it('Should append identities when user present', function(done) {
```
rmi22186, line 1517:
```suggestion
        it('Should append identities when user is present', function(done) {
```

test/src/tests-serverModel.ts
```diff
@@ -0,0 +1,1677 @@
...
+        it('Should append user attributes when user present', function(done) {
+            let sdkEvent = mParticle
+                .getInstance()
+                ._ServerModel.createEventObject(event);
+
+            sdkEvent.should.be.ok;
+            expect(sdkEvent.UserAttributes).to.eql({});
+            var attributes = { foo: 'bar', 'foo-arr': ['bar1', 'bar2'] };
+            var user: MParticleUser = {
+                // TODO: Compiler is not happy with this
```
rmi22186, line 1580: My compiler seemed okay, is it because this TODO was not removed and you added the `as unknown` below to rectify this?

test/src/tests-serverModel.ts
```diff
@@ -0,0 +1,1677 @@
...
+                    return null;
+                },
+            };
+            mParticle.getInstance()._ServerModel.appendUserInfo(user, sdkEvent);
+            expect(sdkEvent.MPID).to.equal('98765');
+            done();
+        });
+
+        it('convertEventToDTO should contain launch referral', function(done) {
+            // TODO: Verify if this should be an SDKEvent or UploadObject
```
rmi22186, line 1637: Should be an upload object.

test/src/tests-serverModel.ts
```diff
@@ -0,0 +1,1675 @@
...
+            ).to.equal(mPStore.clientId);
+            expect(actualEventObject.DeviceId, 'DeviceId').to.equal(
+                mPStore.deviceId
+            );
+            expect(
+                actualEventObject.IntegrationAttributes,
+                'IntegrationAttributes'
+            ).to.eql({});
+
+            // TODO: Should this default to USD?
```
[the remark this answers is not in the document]
alexs-mparticle: Looks like Native does defaults to USD. Creating a follow up task: https://go.mparticle.com/work/SQDSDKS-4817

test/src/tests-serverModel.ts
alexs-mparticle: The reason for setting this to `unknown` and then to `IUploadObject` is so that our tests can have smaller, more basic sample events without getting bogged down with unnecessary requirements. The actual assignment of `pet` inside of `convertEventToDTO` is actually testing the validation.

test/src/tests-serverModel.ts
alexs-mparticle: Turns out it was a typo on line 1583

rmi22186 (approved):
🚢

PR #354, opened 2022-08-25
ci: Update Dependabot Automerge

alexs-mparticle:
## Summary
- Automate Dependabot Automerge

## Master Issue
Closes https://go.mparticle.com/work/SQDSDKS-4487

### DIFF
--- a/.github/workflows/dependabot-automerge.yml
+++ b/.github/workflows/dependabot-automerge.yml
@@ -1,0 +1,13 @@
+name: "Dependabot Automerge"
+
+on:
+  workflow_run:
+    workflows:
+      - "Web SDK Build & Test"
+    types:
+      - completed
+
+jobs:
+  automerge-dependabot:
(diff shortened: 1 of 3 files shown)


.github/workflows/dependabot-automerge.yml
```diff
@@ -0,0 +1,13 @@
+name: "Dependabot Automerge"
+
+on:
+  workflow_run:
+    workflows:
+      - "Web SDK Build & Test"
```
rmi22186, line 6: So this means that whenever we put a PR in, this will run, right?

alexs-mparticle: Yeah, but technically it's when our Build and Test workflow "completes". So it'll only run if a PR finishes running tests. If we want, we can eventually tie this into other tasks, but I'm trying to be consistent with other projects that run this job.

rmi22186 (approved):

PR #361, opened 2022-09-16
fix: Sending multiple UAC events for same value UA

mmustafa-tse:
## Summary
A client reported that UAC events are forwarded on initialization everytime for some attributes even though the values are not changed. Per investigating it seemed that values that are set to false are only affected by this issue, furthermore I was able to also confirm that values such as 0 and an empty string '' would have the same issue since Javascript recognize them as null values. The code change here is to only change previousUserAttributeValue when it comes in as undefined since this is the value when it is not seen before.

## Testing Plan
Tested locally with overrides in sample app and also added unit testing case

### DIFF
--- a/src/identity.js
+++ b/src/identity.js
@@ -1819,7 +1819,7 @@ export default function Identity(mpInstance) {
         deleted,
         user
     ) {
-        if (!previousUserAttributeValue) {
+        if (previousUserAttributeValue === undefined) {
             previousUserAttributeValue = null;
         }
         var userAttributeChangeEvent;
--- a/src/identity.js
+++ b/src/identity.js
@@ -1830,7 +1830,7 @@ export default function Identity(mpInstance) {
                     userAttributeChanges: {
                         UserAttributeName: key,
                         New: newValue,
-                        Old: previousUserAttributeValue || null,
+                        Old: previousUserAttributeValue,
                         Deleted: deleted,
                         IsNewAttribute: isNewAttribute,
                     },
--- a/src/identity.js
+++ b/src/identity.js
@@ -1819,7 +1819,7 @@ export default function Identity(mpInstance) {
         deleted,
         user
     ) {
-        if (previousUserAttributeValue === undefined) {
+        if (typeof previousUserAttributeValue === 'undefined') {
             previousUserAttributeValue = null;
         }
         var userAttributeChangeEvent;
--- a/test/src/tests-identities-attributes.js
+++ b/test/src/tests-identities-attributes.js
@@ -1219,7 +1219,7 @@ describe('identities and attributes', function() {
         event3.event_type.should.equal('user_attribute_change');
         event3.data.new.should.equal('');
         (event3.data.old === null).should.equal(true);
-        event3.data.user_attribute_name.should.equal('testFalse');
+        event3.data.user_attribute_name.should.equal('testEmptyString');
         event3.data.deleted.should.equal(false);
         event3.data.is_new_attribute.should.equal(true);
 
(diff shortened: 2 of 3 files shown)


mmustafa-tse:
@rmi22186

rmi22186:
Thanks @mmustafa-tse!  You should actually make the change inside of the `src/identity.js` folder.  Please undo the change to dist.  Once a PR is merged, our release process automatically builds the src files into the dist folder.

mmustafa-tse:
@rmi22186 - changes requested done, thanks for the clarification!

src/identity.js
```diff
@@ -1819,7 +1819,7 @@ export default function Identity(mpInstance) {
         deleted,
         user
     ) {
-        if (!previousUserAttributeValue) {
+        if (previousUserAttributeValue === undefined) {
```
alexs-mparticle, line 1822: I think this is a better way to check to make sure that this variable is actually defined.

```suggestion
        if (typeof previousUserAttributeValue === 'undefined') {
```

mmustafa-tse:
@alexs-mparticle Agree, changes committed. Thanks!

test/src/tests-identities-attributes.js
```diff
@@ -1163,4 +1163,46 @@ describe('identities and attributes', function() {
 
         done();
     });
+
+    it('should not send user attribute change requests when user attribute already set with same value', function(done) {
```
rmi22186, line 1167: This test would pass even with the original code for `if (!previousUserAttributeValue)`.  Please add tests for where the attribute is `0`, `false`, and `undefined`.

rmi22186 (changes requested):

test/src/tests-identities-attributes.js
mmustafa-tse: @rmi22186 should I also remove the test case for attribute age: 25?

rmi22186: i think you can add a few user attributes, so it's fine to keep them and just add the others.

mmustafa-tse:
@rmi22186 testing cases for 0, '' and  false added.

rmi22186:
This looks good, but I would add one more test.  In the case where you are going from one false-y value to another, let's confirm that a UAC is sent since going from `0` to `''` to `false` are actually different values.  Let's make this a new test also, and just do something like:

`should send user attributes when setting different falsey values`:
1. set a user attribute to 0
2. set the same  user attribute to `''`
3. check that the event exists and is a UAC
4. set the same user attribute to `false`
5. check that the event exists and is a UAC
6. set the same  user attribute back to `0`
7. 5. check that the event exists and is a UAC

test/src/tests-identities-attributes.js
```diff
@@ -1163,4 +1163,171 @@ describe('identities and attributes', function() {
...
+        mParticle.Identity.getCurrentUser().setUserAttribute('testFalse', false);
+        mParticle.Identity.getCurrentUser().setUserAttribute('testEmptyString', '');
+        mParticle.Identity.getCurrentUser().setUserAttribute('testZero', 0);
+        (window.fetchMock.lastOptions() === undefined).should.equal(true);
+        (window.fetchMock._calls.length === 0).should.equal(true);
+
+        done()
+    });
+
+    it('should send user attributes when setting different falsey values', function(done) {
```
rmi22186, line 1259:
```suggestion
    it('should send user attribute change event when setting different falsey values', function(done) {
```

src/identity.js
```diff
@@ -1830,7 +1830,7 @@ export default function Identity(mpInstance) {
                     userAttributeChanges: {
                         UserAttributeName: key,
                         New: newValue,
-                        Old: previousUserAttributeValue || null,
+                        Old: previousUserAttributeValue,
```
rmi22186, line 1833: Any reason for removing the null here?

mmustafa-tse: Another little bug noticed when testing, this is causing the 'old' values in the event batches to be null instead showing their value for the case of falsey attributes. Let me know if this needs to be changed back but isn't the only time we want old to be null when it's never seen before? Hence the condition on line 1822

test/src/tests-identities-attributes.js
mmustafa-tse: @rmi22186 done

src/identity.js
rmi22186: great catch

rmi22186 (approved):

PR #365, opened 2022-11-07
fix: Process queued events after there are no more integration delays

rmi22186:
## Summary
There are cases (GA4 and Adobe) when a customer checks `forward web requests server side`, but the server needs additional information from the client to forward information properly. In the case of GA4, they need a `client_id`, and you can only get this client side from the GA4 SDKs.  We classify `client_id` as an `integration attribute` or `ia`.

([GA4 server side web kit](https://github.com/mparticle-integrations/mparticle-javascript-integration-google-analytics-4/blob/master/packages/GA4Server/src/GoogleAnalytics4EventForwarderServerSide.js) for example) is a simplified kit whose sole purpose is to grab the `client_id` and set it as an `ia` on a payload so that the server has all the required information before forwarding events.

The steps are:
1. [kit tells core SDK to delay sending events](https://github.com/mparticle-integrations/mparticle-javascript-integration-google-analytics-4/blob/master/packages/GA4Server/src/GoogleAnalytics4EventForwarderServerSide.js#L30).  Events must be delayed because `client_id` doesn't exist yet, so if events are sent without it, the server doesn't know what to do.
2. [kit grabs the integration attribute and sets it](https://github.com/mparticle-integrations/mparticle-javascript-integration-google-analytics-4/blob/master/packages/GA4Server/src/GoogleAnalytics4EventForwarderServerSide.js#L30-L71)
3. [kit tells core SDK to not delay sending events anymore for this specific kit](https://github.com/mparticle-integrations/mparticle-javascript-integration-google-analytics-4/blob/master/packages/GA4Server/src/GoogleAnalytics4EventForwarderServerSide.js#L72)

The bug is discovered in the following use case:
* user navigates to a website that has GA4 server side
* user does nothing else

In this scenario, no events are sent.  Events are queued between when the kit sets the delay to `true` and `false`.  There is no logic that says once all integration delays are `false`, go and process the queued events.  Only if the user takes a further action will it kick start the process to queue events.

This PR adds code that checks to see if we should process the queued events after an integration delay is set to `false`.

This bug is a long tail edge case, as it is likely that steps 1, 2, and 3 happen above pretty quickly and the user logs other events.

## Testing Plan
Integration test and manually tested

## Master Issue
Closes https://go.mparticle.com/work/SQDSDKS-4686

### DIFF
--- a/src/mp-instance.js
+++ b/src/mp-instance.js
@@ -1233,7 +1233,7 @@ export default function mParticleInstance(instanceName) {
         // If the integration delay is set to false, check to see if there are any
         // other integration delays set to true.  It not, process the queued events/.
 
-        var integrationDelaysKeys = Object.keys(
+        const integrationDelaysKeys = Object.keys(
             self._preInit.integrationDelays
         );
 
@@ -1241,7 +1241,7 @@ export default function mParticleInstance(instanceName) {
             return;
         }
 
-        var hasIntegrationDelays = integrationDelaysKeys.some(function(
+        const hasIntegrationDelays = integrationDelaysKeys.some(function(
             integration
         ) {
             return self._preInit.integrationDelays[integration] === true;
--- a/src/mp-instance.js
+++ b/src/mp-instance.js
@@ -1219,7 +1219,6 @@ export default function mParticleInstance(instanceName) {
     this._setIntegrationDelay = function(module, boolean) {
         self._preInit.integrationDelays[module] = boolean;
 
-
         // If the integration delay is set to true, no further action needed
         if (boolean === true) {
             return;
--- a/src/mp-instance.js
+++ b/src/mp-instance.js
@@ -1219,12 +1219,12 @@ export default function mParticleInstance(instanceName) {
     this._setIntegrationDelay = function(module, boolean) {
         self._preInit.integrationDelays[module] = boolean;
 
-        
+
         // If the integration delay is set to true, no further action needed
         if (boolean === true) {
             return;
         }
-        // If the integration delay is set to false, check to see if there are any 
+        // If the integration delay is set to false, check to see if there are any
         // other integration delays set to true.  It not, process the queued events/.
         if (Object.keys(self._preInit.integrationDelays).length) {
             if (
--- a/src/mp-instance.js
+++ b/src/mp-instance.js
@@ -1207,6 +1207,21 @@ export default function mParticleInstance(instanceName) {
     };
     this._setIntegrationDelay = function(module, boolean) {
         self._preInit.integrationDelays[module] = boolean;
+
+        if (boolean === true) {
+            return;
+        }
+        if (Object.keys(self._preInit.integrationDelays).length) {
+            if (
+                Object.keys(self._preInit.integrationDelays).filter(function(
+                    singleInt
+                ) {
+                    return self._preInit.integrationDelays[singleInt] === true;
+                }).length === 0
+            ) {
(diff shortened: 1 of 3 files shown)


test/src/tests-forwarders.js
```diff
@@ -2250,6 +2250,28 @@ describe('forwarders', function() {
         done();
     });
 
+    it('integration test - after an integration delay is set to false, should fire an event after the event timeout', function(done) {
+        var clock = sinon.useFakeTimers();
```
alexs-mparticle, line 2254: I think we can safely use `let` and `const` in our tests.

```suggestion
        const clock = sinon.useFakeTimers();
```

src/mp-instance.js
```diff
@@ -1205,8 +1205,44 @@ export default function mParticleInstance(instanceName) {
     this._getIntegrationDelays = function() {
         return self._preInit.integrationDelays;
     };
+    /*
+        An integration delay is a workaround that prevents events from being sent when it is necessary to do so.
+        Some server side integrations require a client side value to be included in the payload set to successfully
```
rmi22186, line 1210:
```suggestion
        Some server side integrations require a client side value to be included in the payload to successfully 
```

src/mp-instance.js
```diff
@@ -1205,8 +1205,44 @@ export default function mParticleInstance(instanceName) {
     this._getIntegrationDelays = function() {
         return self._preInit.integrationDelays;
     };
+    /*
+        An integration delay is a workaround that prevents events from being sent when it is necessary to do so.
+        Some server side integrations require a client side value to be included in the payload to successfully 
+        forward.  This value can only be pulled from the client side partner SDK.
+        
+        During the kit initialization the kit:
```
rmi22186, line 1213:
```suggestion
        During the kit initialization, the kit:
```

src/mp-instance.js
```diff
@@ -1205,8 +1205,37 @@ export default function mParticleInstance(instanceName) {
...
+        attribute is now on the batch payload for the server side to read.
+
+        If there is not delay, then the events sent before an integration attribute is included would not
+        be forwarded successfully server side.
+    */
     this._setIntegrationDelay = function(module, boolean) {
         self._preInit.integrationDelays[module] = boolean;
+
+        // If the integration delay is set to true, no further action needed
+        if (boolean === true) {
```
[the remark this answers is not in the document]
rmi22186: Good catch. I'll update this.

src/mp-instance.js
```diff
@@ -1205,8 +1205,37 @@ export default function mParticleInstance(instanceName) {
...
+            return;
+        }
+        // If the integration delay is set to false, check to see if there are any
+        // other integration delays set to true.  It not, process the queued events/.
```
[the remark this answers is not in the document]
rmi22186: Yea I originally tried to do it on one function and was hoping the comment helped with readability, but breaking it down this way is a good suggestion.  I'm going to change the Array.some suggestion a tiny bit.  Will follow up shortly.

src/mp-instance.js
```diff
@@ -1205,8 +1205,50 @@ export default function mParticleInstance(instanceName) {
...
+
+        var integrationDelaysKeys = Object.keys(
+            self._preInit.integrationDelays
+        );
+
+        if (integrationDelaysKeys.length === 0) {
+            return;
+        }
```
alexs-mparticle, line 1242:
```suggestion
        }

```

alexs-mparticle (approved):

PR #366, opened 2022-11-07
feat: Expose getEnvironment() for mParticle Instance

alexs-mparticle:
## Summary
- expose `getEnvironment()` function to mParticle Instance which will return the current environment setting in `Store`

## Testing Plan
- Run automated tests

## Master Issue
Closes https://go.mparticle.com/work/SQDSDKS-4725

### DIFF
--- a/test/src/tests-types.js
+++ b/test/src/tests-types.js
@@ -8,7 +8,7 @@ var EventType = Types.EventType,
 
 describe('Types', function() {
     describe('Environment', function() {
-        it('should return Production', function() {
+        it('should return `production`', function() {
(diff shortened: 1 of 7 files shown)


test/src/tests-types.js
```diff
@@ -7,6 +7,16 @@ var EventType = Types.EventType,
     IdentityType = Types.IdentityType;
 
 describe('Types', function() {
+    describe('Environment', function() {
+        it('should return Production', function() {
```
rmi22186, line 11:
```suggestion
        it('should return `production`', function() {
```

rmi22186 (approved):

PR #367, opened 2022-11-22
fix: Prevent Batches from queuing empty events

alexs-mparticle:
## Summary
- Within `BatchUploader.queueEvent`, we are only checking if the event is `null`, but not if it is actually empty.
- We don't have reliable test data directly related to this bug report, and is unlikely that an event can get to this point in the code while being empty, but we should be more defensive, just in case.

## Testing Plan
- Using the browser debugger, manually send events into the batch uploader and verify that batches without events are not created.

## Master Issue
Closes https://go.mparticle.com/work/SQDSDKS-4737

### DIFF
--- a/src/batchUploader.ts
+++ b/src/batchUploader.ts
@@ -203,17 +203,18 @@ export class BatchUploader {
 
     private async upload(
         logger: SDKLoggerApi,
-        uploads: Batch[],
+        _uploads: Batch[],
         useBeacon: boolean
     ): Promise<Batch[]> {
         let uploader;
 
-        if (
-            isEmpty(uploads) ||
-            uploads.some(upload => isEmpty(upload.events))
-        ) {
+        // Filter out any batches that don't have events
+        const uploads = _uploads.filter(upload => !isEmpty(upload.events));
+
+        if (isEmpty(uploads)) {
             return null;
         }
+
         logger.verbose(`Uploading batches: ${JSON.stringify(uploads)}`);
         logger.verbose(`Batch count: ${uploads.length}`);
 
--- a/src/batchUploader.ts
+++ b/src/batchUploader.ts
@@ -66,7 +66,7 @@ export class BatchUploader {
     }
 
     queueEvent(event: SDKEvent): void {
-        if (!isEmptyObject(event)) {
+        if (!isEmpty(event)) {
             this.pendingEvents.push(event);
             this.mpInstance.Logger.verbose(
                 `Queuing event: ${JSON.stringify(event)}`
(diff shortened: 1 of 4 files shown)


test/src/tests-batchUploader.ts
```diff
@@ -27,11 +33,66 @@ describe('batch uploader', () => {
...
+
+            it('should reject batches without events', () => {
+                window.mParticle._resetForTests(MPConfig);
+                window.mParticle.init(apiKey, window.mParticle.config);
+
+                const mpInstance = window.mParticle.getInstance();
+
+                const uploader = new BatchUploader(mpInstance, 1000);
+
+                // debugger;
```
rmi22186, line 85:
```suggestion
```

rmi22186 (changes requested):
A few nits.

For one final manual test, can you launch an app locally and ensure that loading the bundle based off these code changes that events are still sent?

src/utils.ts
```diff
@@ -96,6 +96,9 @@ const isDataPlanSlug = (str: string): boolean =>
 const isStringOrNumber = (value: any): boolean =>
     isString(value) || isNumber(value);
 
+const isEmptyObject = (val: Dictionary<any> | null | undefined): boolean =>
```
rmi22186, line 99: I originally said to call this `isEmptyObject` because I didn't see the signature when you were sharing your screen, but it seems like you're checking for arrays here as well (and in your tests too).  If that's the case, maybe reverting to `isEmpty` is more accurate.  Was this function signature pulled from somewhere (and if so, a comment with the original source would be useful like we do throughout [polyfill.js](https://github.com/mParticle/mparticle-web-sdk/blob/master/src/polyfill.js))

rmi22186:
lastly, just to include another hopefully fail safe - can we do something here in the upload function?

https://github.com/mParticle/mparticle-web-sdk/blob/master/src/batchUploader.ts#L199-L214
```
if (uploads[i].events.length<0) {
   continue;
}
```

src/batchUploader.ts
```diff
@@ -64,8 +65,8 @@ export class BatchUploader {
         return false;
     }
 
-    queueEvent(event: SDKEvent) {
-        if (event) {
+    queueEvent(event: SDKEvent): void {
+        if (!isEmptyObject(event)) {
```
rmi22186, line 69:
```suggestion
        if (!isEmpty(event)) {
```

src/utils.ts
alexs-mparticle: Nope. I'm completely to blame for this.

src/utils.ts
```diff
@@ -96,6 +96,9 @@ const isDataPlanSlug = (str: string): boolean =>
 const isStringOrNumber = (value: any): boolean =>
     isString(value) || isNumber(value);
 
+const isEmpty = (val: Dictionary<any> | null | undefined): boolean =>
+    val == null || !(Object.keys(val) || val).length;
+
```
alexs-mparticle, line 101:
```suggestion
const isEmpty = (value: Dictionary<any> | null | undefined): boolean =>
    value == null || !(Object.keys(value) || value).length;

```

test/src/tests-batchUploader.ts
```diff
@@ -686,4 +747,68 @@ describe('batch uploader', () => {
...
+            await (<any>uploader).upload(newLogger, testBatches, false);
+
+            window.fetchMock._calls.forEach(call =>
+                console.log('body events', JSON.parse(call[1].body).events)
+            );
+
+            console.log(
+                'fetch mock calls length',
+                window.fetchMock.calls().length
+            );
```
rmi22186, line 802:
```suggestion
```

test/src/tests-utils.ts
```diff
@@ -165,4 +166,30 @@ describe('Utils', () => {
...
+            expect(isEmpty({'foo': 'bar'})).to.equal(false);
+        });
+
+        it('returns true if object is null', () => {
+            expect(isEmpty(null)).to.equal(true);
+        });
+
+        it('returns true if object is null', () => {
+            expect(isEmpty(null)).to.equal(true);
+        });
```
rmi22186, line 193: looks like the test is a repeat. i assume you want undefined

```suggestion
        it('returns true if object is undefined', () => {
            expect(isEmpty(undefined)).to.equal(true);
        });
```

rmi22186 (changes requested):

rmi22186 (approved):

PR #368, opened 2022-11-23
feat: Add Persistence Wrapper

alexs-mparticle:
## Summary
- Create a Wrapper Class around Local Storage
- Wrapper should allow batches to be saved and retrieved from local storage
- Wrapper should allow batches to be purged without affecting other local storage items
- This wrapper will be part of a larger initiative to refactor persistence

## Testing Plan
- N/A

## Master Issue
Closes https://go.mparticle.com/work/SQDSDKS-4714

### DIFF
--- a/src/vault.ts
+++ b/src/vault.ts
@@ -1,0 +1,83 @@
+import { Batch } from '@mparticle/event-models';
+import { Dictionary, isEmpty } from './utils';
+
+export default class Vault {
+    // TODO: Make this generic
+    public contents: Dictionary<Batch>;
+
+    constructor(public key: string) {
+        this.key = key;
+        this.contents = this.getItems();
+    }
+
+    /**
+     * Stores Batches using `source_request_id` as an index
+     * @method storeBatches
+     * @param {Batches[]} batches an Array of Batches
+     */
+    storeBatches(batches: Batch[]): void {
+        this.contents = this.getItems() || {};
+
+        batches.forEach(batch => {
+            if (batch.source_request_id) {
+                this.contents[batch.source_request_id] = batch;
+            }
+        });
+
+        this.saveItems(this.contents);
+    }
+
+    /**
+     * Removes a single batch based on `source_request_id`
+     * @method removeBatch
+     * @param {String} source_request_id
+     */
+    removeBatch(source_request_id: string): void {
+        this.contents = this.getItems() || {};
+
+        delete this.contents[source_request_id];
+
+        this.saveItems<Batch>(this.contents);
+    }
+
+    /**
+     * Retrieves all batches from local storage as an array
+     * @method retrieveBatches
+     * @returns {Batch[]} a list of Batches
+     */
+    retrieveBatches(): Batch[] {
+        this.contents = this.getItems() || {};
+
+        return Object.keys(this.contents).map(item => this.contents[item]);
+    }
+
+    /**
+     * Removes all persisted data from local storage based on this vault's `key`
+     * @method purge
+     */
+    purge(): void {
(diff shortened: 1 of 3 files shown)


src/vault.ts
```diff
@@ -0,0 +1,83 @@
...
+
+        this.saveItems<Batch>(this.contents);
+    }
+
+    /**
+     * Retrieves all batches from local storage as an array
+     * @method retrieveBatches
+     * @returns {Batch[]} a list of Batches
+     */
+    retrieveBatches(): Batch[] {
```
jared-dickman, line 48: *suggestion* can add modifiers here `public/private` to be explicit about the expected usages

test/src/tests-vault.ts
```diff
@@ -0,0 +1,271 @@
...
+                },
+                'item-456': {
+                    foo: 'bar',
+                    source_request_id: 'item-456',
+                    events: [
+                        { eventName: 'test-event-3' },
+                        { eventName: 'test-event-5' },
+                    ],
+                },
+            };
```
rmi22186, line 70: Similar to above, for readability, this block of code should go right before the expect statement on line 79.

test/src/tests-vault.ts
```diff
@@ -0,0 +1,271 @@
...
+            const batch2 = {
+                foo: 'bar',
+                source_request_id: 'item-456',
+            };
+
+            const batches = ([batch1, batch2] as unknown) as Batch[];
+
+            const vault = new Vault('test-key-store-batches');
+
+            const expectedContents = { 'item-123': batch1, 'item-456': batch2 };
```
rmi22186, line 25: Minor, but I suggest moving `expectedContents` to immediately before the expect assertion on line 29 to not interrupt the actions the tests are taking.

rmi22186 (changes requested):
Mostly nits.  Looks great though.

src/vault.ts
```diff
@@ -0,0 +1,83 @@
+import { Batch } from '@mparticle/event-models';
+import { Dictionary, isEmptyObject } from './utils';
+
+export default class Vault {
+    // TODO: Make this generic
+    public contents: Dictionary<Batch>;
+
+    constructor(public key: string) {
+        this.key = key;
+        this.contents = this.getItems();
```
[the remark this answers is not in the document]
alexs-mparticle: Not sure if this is necessary as `getItems` is already returning an empty object if nothing exists. I can still add it here if you think it makese sense from a readability standpoint.

src/vault.ts
alexs-mparticle: Good call out

src/vault.ts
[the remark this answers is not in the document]
rmi22186: True, not necessary then.

src/vault.ts
```diff
@@ -0,0 +1,83 @@
+import { Batch } from '@mparticle/event-models';
+import { Dictionary, isEmptyObject } from './utils';
+
+export default class Vault {
+    // TODO: Make this generic
+    public contents: Dictionary<Batch>;
+
+    constructor(public key: string) {
+        this.key = key;
```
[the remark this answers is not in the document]
rmi22186: Yea I did see that.  I was thinking about it being overwritten internally. Will TS be smart enough to know that it is a string if it is being set immediately by the constructor?  If so then it'd be overkill to add it.  Otherwise, we should make the key private so that it can only ever be set by the constructor.

rmi22186 (approved):
lgtm besides the one comment.

rmi22186 (approved):

PR #373, opened 2022-12-06
feat: Add Support for Batch Persistence when Offline

alexs-mparticle:
## Instructions

 ## Summary
 - As events are queued, they are placed into local storage for later processing. Once the `prepareAndUpload` method is triggered, events are removed from local storage and converted to `Batches`, then placed into the batch vault.
 - This pull request also refactors parts of the BatchUploader to better handle asynchronous uploads. Currently, a single asynchronous request is made that iterates through a list of batches, yet each batch is individually uploaded to the server. Loops inside of promises are not recommended in JavaScript as they could lead a longer turnaround time to process the response or other unexpected results. By breaking up the single upload promise into individual promises, each upload is a single process that returns quicker requires less processing.
 - Also in this PR are updates to our event conversion logic so that we can make sure that consent state is not lost when deserializing events and batches when storing in local storage.
 - Lastly, Beacon Uploader tests are moved out of the Batch Uploader test suite as there was an issue with race condition in the test runner.

 ## Testing Plan
 - Use a sample app to manually test the event upload flow
 - Update your config to make `eventBatchingIntervalMillis` longer than `5000` milliseconds
 - Verify that events are stored in local storage and then removed and converted into batches
 - Verify that events upload as expected
 - Set browser to offline and verify that batches and events continue to get stored in local storage
 - Verify that batches are uploaded once connectivity is restored

 - Closes https://go.mparticle.com/work/SQDSDKS-4715


src/batchUploader.ts
```diff
@@ -172,29 +200,52 @@ export class BatchUploader {
...
+        // this.pendingEvents = [];
+
+        // TODO: Retrieve and Purge events from Event Vault
+        const currentEvents = this.eventVault.retrieveItems();
+        // Should this really purge, or just empty local storage?
+        this.eventVault.purge();
+        // TODO: Deprecate pending Events
         this.pendingEvents = [];
+
         const newUploads = BatchUploader.createNewUploads(
```
rmi22186, line 213: let's rename newUploads to be more specific, and add the type to it also

src/batchUploader.ts
```diff
@@ -208,6 +259,9 @@ export class BatchUploader {
     ): Promise<Batch[]> {
         let uploader;
 
+        // TODO: Looks like tests are not running this uploader, and not clearing out
+        //       the batches in the vault?
+
```
rmi22186, line 264: resolved?

src/batchUploader.ts
```diff
@@ -172,29 +200,52 @@ export class BatchUploader {
     private async prepareAndUpload(triggerFuture: boolean, useBeacon: boolean) {
         const currentUser = this.mpInstance.Identity.getCurrentUser();
 
-        const currentEvents = this.pendingEvents;
+        // const currentEvents = this.pendingEvents;
+        // this.pendingEvents = [];
+
+        // TODO: Retrieve and Purge events from Event Vault
+        const currentEvents = this.eventVault.retrieveItems();
+        // Should this really purge, or just empty local storage?
```
rmi22186, line 208: Is the difference between purging and emptying local storage that it will be an array or empty? if so, i think it should just be purged into an empty array.

src/batchUploader.ts
```diff
@@ -172,29 +200,52 @@ export class BatchUploader {
     private async prepareAndUpload(triggerFuture: boolean, useBeacon: boolean) {
         const currentUser = this.mpInstance.Identity.getCurrentUser();
 
-        const currentEvents = this.pendingEvents;
+        // const currentEvents = this.pendingEvents;
```
rmi22186, line 203: This section would benefit most from a visual flow of how an event gets uploaded, and if it is in memory, local storage, when it's removed, etc.  some of the TODOs and in memory are difficult to grok without that flow chart

src/batchUploader.ts
```diff
@@ -233,6 +287,8 @@ export class BatchUploader {
                 let blob = new Blob([fetchPayload.body], {
                     type: 'text/plain;charset=UTF-8',
                 });
+                // TODO: Can we flag anytihng in the vault to say it's
```
rmi22186, line 290: i think this would be helpful

test/src/tests-batchUploader.ts
```diff
@@ -88,13 +204,15 @@ describe('batch uploader', () => {
                 uploader.queueEvent(({} as unknown) as SDKEvent);
 
                 expect(uploader.pendingEvents).to.eql([]);
-                expect(uploader.pendingUploads).to.eql([]);
+                // TODO: Should we remove because of Vault?
+                // expect(uploader.pendingUploads).to.eql([]);
```
rmi22186, line 208: i think the answer to this depends on the flow chart you make. we need to be really clear on when things remain in memory vs when they are cleared.  it's a layer of complexity that needs to be diagrammed

test/src/tests-batchUploader.ts
```diff
@@ -88,13 +204,15 @@ describe('batch uploader', () => {
...
             });
         });
     });
 
     describe('batching via window.fetch', () => {
         beforeEach(function() {
+            window.fetchMock.restore();
```
rmi22186, line 215: Nit - this is probably fine as is, but technically if should just be in `afterEach` and maybe once in a `before` block.

test/src/tests-batchUploader.ts
```diff
@@ -305,7 +424,11 @@ describe('batch uploader', () => {
             // start.
 
             window.mParticle._resetForTests(MPConfig);
-            
+
+            // TODO: For some reason, with offline batching enabled,
```
rmi22186, line 428: is this outdated now? i believe only the batchuploader for dom and window events are failing now right?

rmi22186 (changes requested):

src/batchUploader.ts
```diff
@@ -59,27 +79,55 @@ export class BatchUploader {
...
         return false;
     }
 
+    // Triggers a setTimeout for prepareAndUpload
+    private triggerUploadInterval(
+        triggerFuture: boolean = false,
+        useBeacon: boolean = false
+    ): void {
+        // Clear out existing interval before creating a new one
+        clearTimeout(this.uploadIntervalTimeout);
```
rmi22186, line 96: I see what you are trying to do here, but `clearTimeout` accepts a timeout id, and not a number of seconds. You can have multiple timeouts, and each one returns a timeout id.  The individual timeouts would need to be set to different variables which would hold the timeout id and then you can pass the variable to `clearTimeout`.   [For reference](https://developer.mozilla.org/en-US/docs/Web/API/clearTimeout).

If I am reading this right, it should be:
```
clearTimeout(this.uploadIntervalTimeout)
```

src/batchUploader.ts
```diff
@@ -59,27 +79,55 @@ export class BatchUploader {
...
+            this.prepareAndUpload(triggerFuture, useBeacon);
+        }, this.uploadIntervalMillis);
+    }
+
     queueEvent(event: SDKEvent): void {
         if (!isEmpty(event)) {
+            // TODO: This is where we should store events in Vault
             this.pendingEvents.push(event);
+            // TODO: Maybe we need a storeItem function?
+            // this.eventVault.storeItems([event]);
```
rmi22186, line 107: uncomment?

src/batchUploader.ts
```diff
@@ -31,22 +35,38 @@ export class BatchUploader {
         this.pendingEvents = [];
         this.pendingUploads = [];
 
+        // this.eventVault = new Vault<SDKEvent>(
+        //     `${mpInstance._Store.storageName}-events`,
+        //     'SourceMessageId',
+        //     {
+        //         logger: mpInstance.Logger,
+        //     }
+        // );
```
rmi22186, line 44: uncomment?

src/batchUploader.ts
```diff
@@ -59,27 +79,55 @@ export class BatchUploader {
...
                 `Queuing event: ${JSON.stringify(event)}`
             );
             this.mpInstance.Logger.verbose(
                 `Queued event count: ${this.pendingEvents.length}`
             );
 
+            // TODO: Do we still needto check if batching is enabled?
+            //       I thought we are moving towards having this on for everyone?
+            // TODO: Need a workaround for the set timeout being triggered
+            // TODO: Refactor this logic so that it is clear that priority events are procesed
```
rmi22186, line 119: Could this just be a comment to clarify?

src/batchUploader.ts
```diff
@@ -174,31 +222,59 @@ export class BatchUploader {
...
+        // that is already in transit
         if (remainingUploads && remainingUploads.length) {
             this.pendingUploads.unshift(...remainingUploads);
+        } else if (!isEmpty(this.batchVault.contents)) {
+            this.pendingUploads.push(...this.batchVault.retrieveItems());
         }
 
         if (triggerFuture) {
-            setTimeout(() => {
+            // TODO: Nothing catches this. We should store the timer in the instance
```
rmi22186, line 265: When you say catches, do you mean try/catch for an error?

src/batchUploader.ts
```diff
@@ -174,31 +222,59 @@ export class BatchUploader {
...
+            //       and clear the timeout when this is done
+            this.uploadIntervalTimeout = setTimeout(() => {
                 this.prepareAndUpload(true, false);
             }, this.uploadIntervalMillis);
         }
+        if (
+            isEmpty(this.pendingUploads) &&
+            // isEmpty(this.eventVault.contents) &&
```
rmi22186, line 273: uncomment?

rmi22186 (changes requested):

src/batchUploader.ts
```diff
@@ -20,8 +20,9 @@ export class BatchUploader {
     mpInstance: MParticleWebSDK;
     uploadUrl: string;
     batchingEnabled: boolean;
-    private eventVault: Vault<SDKEvent>;
+    // private eventVault: Vault<SDKEvent>;
```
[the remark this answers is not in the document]
alexs-mparticle: I think I commented it out to test something. You can ignore.

src/batchUploader.ts
alexs-mparticle: yeah, ignore

src/batchUploader.ts
alexs-mparticle: We should do this as part of a future revision to prevent scope creep in this PR.

src/batchUploader.ts
```diff
@@ -169,156 +217,163 @@ export class BatchUploader {
...
     }
 }
 
 class FetchUploader extends AsyncUploader {
-    private async upload(
-        fetchPayload: fetchPayload,
-        uploads: Batch[],
-        i: number
-    ) {
+    public async upload(fetchPayload: fetchPayload) {
```
jared-dickman, line 360:
```suggestion
    public async upload(fetchPayload: fetchPayload): XHRResponse {
```

src/batchUploader.ts
```diff
@@ -169,156 +217,163 @@ export class BatchUploader {
...
     }
 }
 
 class XHRUploader extends AsyncUploader {
-    private async upload(
-        fetchPayload: fetchPayload,
-        uploads: Batch[],
-        i: number
-    ) {
+    public async upload(fetchPayload: fetchPayload) {
```
jared-dickman, line 367:
```suggestion
    public async upload(fetchPayload: fetchPayload): XHRResponse {
```

test/src/tests-batchUploader.ts
```diff
@@ -305,157 +331,223 @@ describe('batch uploader', () => {
...
+            expect(batch3.events.length, 'Batch 3: Session End').to.equal(1);
+
+            // session start, AST
+            expect(batch4.events.length, 'Batch 4: UAC event').to.equal(1);
+
+            // UAC event
+            expect(
+                batch5.events.length,
+                'Batch 5: Session Start, AST'
+            ).to.equal(2);
```
rmi22186, line 396:
```suggestion
            // UAC event
            expect(batch4.events.length, 'Batch 4: UAC event').to.equal(1);

            // session start, AST
            expect(
                batch5.events.length,
                'Batch 5: Session Start, AST'
            ).to.equal(2);
```

test/src/tests-batchUploader.ts
```diff
@@ -305,157 +331,223 @@ describe('batch uploader', () => {
...
+            (typeof batch2.source_request_id).should.equal('string');
+            (typeof batch3.source_request_id).should.equal('string');
+            (typeof batch4.source_request_id).should.equal('string');
+            (typeof batch5.source_request_id).should.equal('string');
+
+            batch1.source_request_id.should.not.equal(batch2.source_request_id);
+            batch1.source_request_id.should.not.equal(batch3.source_request_id);
+            batch1.source_request_id.should.not.equal(batch4.source_request_id);
+            batch1.source_request_id.should.not.equal(batch5.source_request_id);
+            batch2.source_request_id.should.not.equal(batch3.source_request_id);
```
rmi22186, line 450: Adding a line to distinguish between batch groups better
```suggestion
            batch1.source_request_id.should.not.equal(batch5.source_request_id);

            batch2.source_request_id.should.not.equal(batch3.source_request_id);
```

rmi22186 (changes requested):
mostly nits.  looking really great!

src/batchUploader.ts
```diff
@@ -67,7 +91,11 @@ export class BatchUploader {
 
     queueEvent(event: SDKEvent): void {
         if (!isEmpty(event)) {
+            // TODO: This is where we should store events in Vault
             this.pendingEvents.push(event);
+            // TODO: Maybe we need a storeItem function?
+            this.eventVault.storeItems([event]);
```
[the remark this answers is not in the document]
rmi22186: If not now, then add as a sustained engineering task?

src/batchUploader.ts
```diff
@@ -169,156 +219,165 @@ export class BatchUploader {
...
         }
 
-        if (triggerFuture) {
-            setTimeout(() => {
-                this.prepareAndUpload(true, false);
-            }, this.uploadIntervalMillis);
+        this.batchVault.storeItems([...newBatches]);
+
+        const currentUploads = this.batchVault.retrieveItems();
+        const remainingUploads: Batch[] = [];
```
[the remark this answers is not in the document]
rmi22186: or another name

src/batchUploader.ts
```diff
@@ -169,156 +219,165 @@ export class BatchUploader {
...
+        batch: Batch,
         useBeacon: boolean
-    ): Promise<Batch[]> {
-        let uploader;
-
-        // Filter out any batches that don't have events
-        const uploads = _uploads.filter(upload => !isEmpty(upload.events));
-
-        if (isEmpty(uploads)) {
+    ): Promise<Batch> {
```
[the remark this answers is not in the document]
rmi22186: i think it's the 1st one.

src/vault.ts
```diff
@@ -61,24 +83,25 @@ export default class Vault {
...
             );
         } catch (err) {
             console.error('Cannot Save Items to Local Storage', err);
         }
     }
 
-    private getItems<T>(): Dictionary<T> {
-        const itemString = window.localStorage.getItem(this._key);
+    private getItems(): Dictionary<StorableItem> {
+        // TODO: return contents if local storage is unavailable?
```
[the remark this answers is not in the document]
rmi22186: Sus eng task?  P4

src/batchUploader.ts
```diff
@@ -169,156 +219,165 @@ export class BatchUploader {
...
-        const remainingUploads: Batch[] = await this.upload(
-            this.mpInstance.Logger,
-            currentUploads,
-            useBeacon
-        );
-        if (remainingUploads && remainingUploads.length) {
-            this.pendingUploads.unshift(...remainingUploads);
+        if (isEmpty(newBatches)) {
+            // TODO: We should refactor queueEvents to fire the interval if new events
+            //       are ready for processing rather than do it here
```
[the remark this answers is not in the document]
alexs-mparticle: This indentation is intentional

test/src/tests-main.js
```diff
@@ -29,6 +28,9 @@ afterEach(function() {
     window.fetchMock.restore();
 });
 
+// NOTE: Beacon upload must come before core sdk tests beause of race condition
+import './tests-beaconUpload';
+
```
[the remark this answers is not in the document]
alexs-mparticle: This space is intentional for readability so we can show that the first test is a special case.

src/vault.ts
```diff
@@ -36,12 +59,7 @@ export default class Vault<StorableItem extends Dictionary> {
     public storeItems(items: StorableItem[]): void {
         this.contents = this.getItems();
```
rmi22186, line 60: I think line 60 and 64 can be removed since this now invokes `storeItem` which `savesItem` in that function

rmi22186 (changes requested):
One last comment

rmi22186 (approved):
Approved!

PR #375, opened 2022-12-21
feat: Add wrapper SDK name and version properties

einsteinx2:
## Summary
Add wrapper SDK name and version properties (not used yet)

## Testing Plan
Confirmed approach with Rob, will test thoroughly once we get to the next steps of actually using these values

## Master Issue
Closes https://go.mparticle.com/work/SQDSDKS-4758

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -70,6 +70,9 @@ export interface SDKConfig {
     v1SecureServiceUrl?: string;
     v2SecureServiceUrl?: string;
     v3SecureServiceUrl?: string;
+
+    wrapperSDKName: string;
+    wrapperSDKVersion?: string;
 }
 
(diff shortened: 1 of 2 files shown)


src/store.ts
```diff
@@ -70,6 +70,9 @@ export interface SDKConfig {
     v1SecureServiceUrl?: string;
     v2SecureServiceUrl?: string;
     v3SecureServiceUrl?: string;
+
+    wrapperSDKName: string;
```
alexs-mparticle, line 74: Looks like there's a build test that is failing because this field is missing in the interface for SDKInitConfig in `sdkRuntimeModels.ts`

einsteinx2:
Closing this to change approach to match new iOS solution

PR #376, opened 2022-12-27
fix: Add wrapper SDK name/version setter for diagnostics

einsteinx2:
## Summary
Add ability for wrapper SDKs to set their name and version on the core SDK.

## Testing Plan
Added unit tests and confirmed locally

## Master Issue
Closes https://go.mparticle.com/work/SQDSDKS-4758

[SQDSDKS-4758]: https://mparticle-eng.atlassian.net/browse/SQDSDKS-4758?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -102,6 +102,14 @@ export type PixelConfiguration = Dictionary;
 export type MigrationData = Dictionary;
 export type ServerSettings = Dictionary;
 
+type WrapperSDKTypes = 'flutter' | 'none';
+interface WrapperSDKInfo {
+    name: WrapperSDKTypes;
+    version: string | null;
+    isInfoSet: boolean;
+}
+
+
 // Temporary Interface until Store can be refactored as a class
 export interface IStore {
     isEnabled: boolean;
@@ -144,6 +152,7 @@ export interface IStore {
     pixelConfigurations: PixelConfiguration[];
     integrationDelayTimeoutStart: number; // UNIX Timestamp
     webviewBridgeEnabled?: boolean;
+    wrapperSDKInfo: WrapperSDKInfo;
 }
 
 // TODO: Merge this with SDKStoreApi in sdkRuntimeModels
@@ -191,6 +200,11 @@ export default function Store(
         kits: {},
         configuredForwarders: [],
         pixelConfigurations: [],
+        wrapperSDKInfo: {
+            name: 'none',
+            version: null,
+            isInfoSet: false,
+        },
     };
 
     for (var key in defaultStore) {
--- a/test/src/tests-core-sdk.js
+++ b/test/src/tests-core-sdk.js
@@ -1143,7 +1143,7 @@ describe('core SDK', function() {
         done();
     });
 
-    it('should set the wrapper sdk info in store when mParticle._setWrapperSDKInfo() method is called and init is called', function(done) {
+    it('should set the wrapper sdk info in Store when mParticle._setWrapperSDKInfo() method is called after init is called', function(done) {
         mParticle._resetForTests(MPConfig);
 
         mParticle.getInstance()._setWrapperSDKInfo('flutter', '1.0.3');
--- a/test/src/tests-core-sdk.js
+++ b/test/src/tests-core-sdk.js
@@ -1119,7 +1119,7 @@ describe('core SDK', function() {
         done();
     });
 
-    it('should not set the wrapper sdk info in store when mParticle._setWrapperSDKInfo() method is called if init not called', function(done) {
+    it('should not set the wrapper sdk info in Store when mParticle._setWrapperSDKInfo() method is called if init not called', function(done) {
         mParticle._resetForTests(MPConfig);
 
         mParticle.getInstance()._setWrapperSDKInfo('flutter', '1.0.3');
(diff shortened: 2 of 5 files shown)


alexs-mparticle (changes requested):
LGTM but I see that something is wrong with the checks. It looks like the PR Build and Test actions are not being executed. I think there might be an issue in a previous commit in the `development` branch that is not triggering these actions.

rmi22186 (changes requested):
Looks good so far.  I believe you mentioned on iOS that there was some logic that stated if previously set, then do not set again.  That should be here as well.

einsteinx2:
> Looks good so far. I believe you mentioned on iOS that there was some logic that stated if previously set, then do not set again. That should be here as well.

Good catch, I rewrote it with that logic, also I organized all of the wrapper sdk info values in the store inside an object to pollute the top level less. Let me know if the boolean is an ok approach. I used the boolean so that the logic wouldn't have to check for specific default values which may change in the future, but I'm open to any other suggestions. Or if there's a way to mark code to only run once like on iOS.

einsteinx2:
> LGTM but I see that something is wrong with the checks. It looks like the PR Build and Test actions are not being executed. I think there might be an issue in a previous commit in the `development` branch that is not triggering these actions.

Rebased and force pushed. Looks like it's running the tests now.

test/src/tests-core-sdk.js
```diff
@@ -1118,4 +1118,58 @@ describe('core SDK', function() {
 
         done();
     });
+
+    it('should not set the wrapper sdk info in store when mParticle._setWrapperSDKInfo() method is called if init not called', function(done) {
```
rmi22186, line 1122:
```suggestion
    it('should not set the wrapper sdk info in Store when mParticle._setWrapperSDKInfo() method is called if init not called', function(done) {
```

test/src/tests-core-sdk.js
```diff
@@ -1118,4 +1118,58 @@ describe('core SDK', function() {
...
+        mParticle.init(apiKey, window.mParticle.config);
+
+        mParticle.getInstance()._Store.wrapperSDKInfo.name.should.equal('none');
+        (mParticle.getInstance()._Store.wrapperSDKInfo.version === null).should.equal(true);
+        mParticle.getInstance()._Store.wrapperSDKInfo.isInfoSet.should.equal(false);
+
+        done();
+    });
+
+    it('should set the wrapper sdk info in store when mParticle._setWrapperSDKInfo() method is called and init is called', function(done) {
```
rmi22186, line 1146:
```suggestion
    it('should set the wrapper sdk info in Store when mParticle._setWrapperSDKInfo() method is called after init is called', function(done) {
```

test/src/tests-core-sdk.js
```diff
@@ -1118,4 +1118,58 @@ describe('core SDK', function() {
...
+        mParticle.init(apiKey, window.mParticle.config);
+
+        mParticle.getInstance()._Store.wrapperSDKInfo.name.should.equal('flutter');
+        mParticle.getInstance()._Store.wrapperSDKInfo.version.should.equal('1.0.3');
+        mParticle.getInstance()._Store.wrapperSDKInfo.isInfoSet.should.equal(true);
+        
+        done();
+    });
+
+    it('should not set the wrapper sdk info in store after it has been set', function(done) {
```
rmi22186, line 1160:
```suggestion
    it('should not set the wrapper sdk info in Store after it has previously been set', function(done) {
```

rmi22186 (changes requested):
Minor nits added here for the tests for clarity.

The only thing that is missing is you should follow the format in `mparticle-instance-manager` to allow both setting it via instance and globally.  I misunderstood what you meant when we chatted about this on Slack when I said it should be accessible by the instance.  But it should be accessible both from `mp-instance.js` and `mparticle-intsance-manager.js`.

src/store.ts
```diff
@@ -102,6 +102,14 @@ export type PixelConfiguration = Dictionary;
 export type MigrationData = Dictionary;
 export type ServerSettings = Dictionary;
 
+type WrapperSDKTypes = 'flutter' | 'none';
+interface WrapperSDKInfo {
```
alexs-mparticle, line 106: I think `WrapperSDKOptions` makes more sense

einsteinx2: Sure, I'll make the change

alexs-mparticle: @einsteinx2 I meant the interface, not `name`

rmi22186 (changes requested):
This is still missing the public api in the instance manager.  You can follow the format of any of the setters [like this one](https://github.com/mParticle/mparticle-web-sdk/blob/master/src/mparticle-instance-manager.js#L118-L120)

rmi22186:
> This is still missing the public api in the instance manager. You can follow the format of any of the setters [like this one](https://github.com/mParticle/mparticle-web-sdk/blob/master/src/mparticle-instance-manager.js#L118-L120)

nevermind, looks like i was looking at an older PR.

rmi22186 (approved):

alexs-mparticle (approved):

PR #377, opened 2022-12-27
ci: Fix security lint checks

rmi22186:
## Summary
Ever since the security check was added to the Github action, it prevented the build and test from running.  This should fix that.  Example failure here - https://github.com/mParticle/mparticle-web-sdk/actions/runs/3789930377

## Testing Plan
NA

### DIFF
--- a/.github/workflows/reusable-workflows.yml
+++ b/.github/workflows/reusable-workflows.yml
@@ -15,6 +15,6 @@ jobs:
     uses: mParticle/mparticle-workflows/.github/workflows/pr-branch-target-gitflow.yml@stable
   security-lint-checks:
     name: Security Lint Checks
-    uses: mparticle/mparticle-workflows/.github/workflows/security-checks.yml@main
+    uses: mparticle/mparticle-workflows/.github/workflows/security-checks.yml@stable
     with:
       base_branch: "development"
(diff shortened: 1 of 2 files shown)


.github/workflows/reusable-workflows.yml
```diff
@@ -13,3 +13,8 @@ jobs:
   pr-branch-target-gitflow:
     name: Check PR for semantic target branch
     uses: mParticle/mparticle-workflows/.github/workflows/pr-branch-target-gitflow.yml@stable
+  security-lint-checks:
+    name: Security Lint Checks
+    uses: mparticle/mparticle-workflows/.github/workflows/security-checks.yml@main
+    with:
+      base_branch: "development"
```
alexs-mparticle, line 20: Shouldn't this be running in the current PR branch, rather than development?

.github/workflows/reusable-workflows.yml
```diff
@@ -13,3 +13,8 @@ jobs:
   pr-branch-target-gitflow:
     name: Check PR for semantic target branch
     uses: mParticle/mparticle-workflows/.github/workflows/pr-branch-target-gitflow.yml@stable
+  security-lint-checks:
+    name: Security Lint Checks
+    uses: mparticle/mparticle-workflows/.github/workflows/security-checks.yml@main
```
alexs-mparticle, line 18: I forget the difference, but shouldn't we use `stable` instead of `main`?

```suggestion
    uses: mparticle/mparticle-workflows/.github/workflows/security-checks.yml@stable
```

alexs-mparticle, line 18: I forget the difference, but shouldn't we use `stable` instead of `main`?

```suggestion
    uses: mparticle/mparticle-workflows/.github/workflows/security-checks.yml@stable
```

alexs-mparticle (approved):

PR #382, opened 2023-01-12
refactor: Migrate Config API Client to TypeScript

alexs-mparticle:
## Instructions

 ## Summary
 - Migrate ConfigAPIClient to TypeScript
 - Add Unit Tests

 ## Testing Plan
 - Run automated tests
 - Test manually via Sample App

 - Closes https://go.mparticle.com/work/SQDSDKS-5010

### DIFF
--- a/src/configAPIClient.ts
+++ b/src/configAPIClient.ts
@@ -1,5 +1,11 @@
 import { DataPlanConfig } from '@mparticle/web-sdk';
-import { KitConfigs, MParticleWebSDK, SDKInitConfig } from './sdkRuntimeModels';
+import {
+    BooleanStringLowerCase,
+    MParticleWebSDK,
+    SDKEventCustomFlags,
+    SDKInitConfig,
+} from './sdkRuntimeModels';
+import { Dictionary } from './utils';
 
 export type SDKInitFunction = (
     apiKey: string,
@@ -7,9 +13,52 @@ export type SDKInitFunction = (
     mpInstance: MParticleWebSDK
 ) => void;
 
+export interface IKitConfigs {
+    name: string;
+    moduleId: number;
+    isDebug: boolean;
+    isVisible: boolean;
+    isDebugString: BooleanStringLowerCase;
+    hasDebugString: BooleanStringLowerCase;
+    settings: Dictionary;
+    screenNameFilters: number[];
+    screenAttributeFilters: number[];
+    userIdentityFilters: number[];
+    userAttributeFilters: number[];
+    eventNameFilters: number[];
+    eventTypeFilters: number[];
+    attributeFilters: number[];
+    filteringEventAttributeValue: Dictionary;
+    filteringConsentRuleValues: Dictionary;
+    consentRegulationFilters: number[];
+    consentRegulationPurposeFilters: number[];
+    messageTypeFilters: number[];
+    messageTypeStateFilters: number[];
+    eventSubscriptionId: number;
+    excludeAnonymousUser: boolean;
+}
+
+export interface IPixelConfig {
+    name: string;
+    moduleId: number;
+    esId: number;
+    isDebug: boolean;
+    isProduction: boolean;
+    settings: Dictionary;
+    frequencyCap: number;
+    pixelUrl: string;
+    redirectUrl: string;
+}
+
 export interface IConfigResponse {
     appName: string;
-    kitConfigs: KitConfigs[];
+    kitConfigs: IKitConfigs[];
+    serviceUrl: string;
+    secureServiceUrl: string;
+    minWebviewBridgeVersion: number;
+    workspaceToken: string;
+    pixelConfigs: IPixelConfig[];
+    flags: SDKEventCustomFlags;
 }
 
 export interface IConfigAPIClient {
--- a/src/configAPIClient.ts
+++ b/src/configAPIClient.ts
@@ -28,7 +28,6 @@ export default function ConfigAPIClient(this: IConfigAPIClient) {
         completeSDKInitialization,
         mpInstance
     ): void => {
-        // debugger;
         let url: string;
         try {
             const xhrCallback = function() {
--- a/src/configAPIClient.ts
+++ b/src/configAPIClient.ts
@@ -1,13 +1,14 @@
 import { DataPlanConfig } from '@mparticle/web-sdk';
 import {
     BooleanStringLowerCase,
+    DataPlanResult,
     MParticleWebSDK,
     SDKEventCustomFlags,
     SDKInitConfig,
 } from './sdkRuntimeModels';
 import { Dictionary } from './utils';
 
-export type SDKInitFunction = (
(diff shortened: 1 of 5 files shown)


src/configAPIClient.ts
```diff
@@ -1,13 +1,37 @@
...
+export default function ConfigAPIClient(this: IConfigAPIClient) {
+    this.getSDKConfiguration = (
         apiKey,
         config,
         completeSDKInitialization,
         mpInstance
-    ) {
-        var url;
+    ): void => {
+        // debugger;
```
alexs-mparticle, line 31:
```suggestion
```

src/configAPIClient.ts
```diff
@@ -1,13 +1,37 @@
-export default function ConfigAPIClient() {
-    this.getSDKConfiguration = function(
+import { DataPlanConfig } from '@mparticle/web-sdk';
+import { KitConfigs, MParticleWebSDK, SDKInitConfig } from './sdkRuntimeModels';
+
+export type SDKInitFunction = (
```
rmi22186, line 4: The name of this type is misleading.  It's not really the `SDKInitFunction`, since that does't pass the mpInstance.  Kind of wordy, but I'd rename this to `CompleteSDKInitFunction`

src/configAPIClient.ts
```diff
@@ -1,13 +1,37 @@
...
+    apiKey: string,
+    config: SDKInitConfig,
+    mpInstance: MParticleWebSDK
+) => void;
+
+export interface IConfigResponse {
+    appName: string;
```
rmi22186, line 11: While this is sufficient to get your tests running, I'd make this config response more complete.  Here is an example.
https://jssdkcdns.mparticle.com/js/v2/beab3f4d34281d45bfcdbbd7eb21c083/config
(if you have a JSON reading plugin this makes it easier to read)
image

So I'd suggest making this the full interface for the Config response to include the following as well:
* pixelConfigs: PixelConfig (see below comment on defining this)
* serviceUrl: string
* secureServiceUrl: string
* minWebviewBridgeVersion: number
* workspaceToken: string
* flags:
   * eventsV3: string
   * eventBatchingIntervalMillis: string
 * dataPlanResult: DataPlanResult (this is not part of the above screenshot because i don't have a dataplan, but I just slacked you [a link](https://mparticle.slack.com/archives/DHWS0FJTA/p1674060642699989) to see)

Also, add:
`export type PixelConfig = Dictionary` to `sdkrunTimemodels.ts` (right after `export type KitConfigs = Dictionary` on line 164) and import that here as the type

src/configAPIClient.ts
```diff
@@ -1,13 +1,37 @@
...
+export interface IConfigAPIClient {
+    getSDKConfiguration: (
+        apiKey: string,
+        config: SDKInitConfig,
+        completeSDKInitialization: SDKInitFunction,
+        mpInstance: MParticleWebSDK
+    ) => void;
+}
+
+export default function ConfigAPIClient(this: IConfigAPIClient) {
```
rmi22186, line 24: As part of this migration this should be converted into a TS class instead of remaining an ES5 class-like function:

```
export class ConfigAPIClient {
   constructor(blah blah)
}
```

test/src/tests-config-api-client.ts
```diff
@@ -0,0 +1,182 @@
...
+    });
+
+    afterEach(() => {
+        mockServer.reset();
+        sinon.restore();
+    });
+
+    describe('#getSDKConfiguration', () => {
+        describe('with defaults', () => {
+            beforeEach(() => {});
```
rmi22186, line 36: If nothing is in this `beforeEach` block, let's take it out

```suggestion
```

test/src/tests-config-api-client.ts
```diff
@@ -0,0 +1,182 @@
...
+                    planVersion: 42,
+                };
+
+                mockServer.respondWith(configUrl, [
+                    200,
+                    {},
+                    JSON.stringify({
+                        appName: 'Test App',
+                        kitConfigs: [],
+                        dataPlan,
```
rmi22186, line 121: This is actually returned as a `dataPlanResult`, which I linked you [in slack to](https://mparticle.slack.com/archives/DHWS0FJTA/p1674060642699989)

test/src/tests-config-api-client.ts
```diff
@@ -0,0 +1,182 @@
...
+                expect(mockServer.lastRequest.method).to.equal('get');
+                expect(mockServer.lastRequest.url).to.equal(configUrl);
+
+                const configResponse: IConfigResponse = JSON.parse(
+                    mockServer.lastRequest.response
+                );
+
+                expect(configResponse).to.be.ok;
+                expect(configResponse.appName).to.equal('Test App');
+                expect(configResponse.kitConfigs).to.be.ok;
```
rmi22186, line 178: You are missing an expect for the data plan here.

rmi22186 (changes requested):
Great start to moving this over so quickly.  I have several comments, but thanks for adding all those tests!

src/configAPIClient.ts
alexs-mparticle: I would normally agree, but I want to focus this migration on a direct translation to typescript and fleshing out type definitions rather than making changes to core functionality or converting functions to classes to avoid any potential side effects.

Also, as we are using similar API clients elsewhere in the SDK, I would rather combine these clients into a single API client class as part of a future pull request.

src/configAPIClient.ts
alexs-mparticle: I made some adjustments based on your feedback. Take a look and let me know if you think there's anything that should be optional rather than required or if we need to extend what I have.

We probably need to sync up on `DataPlanResult` but I think I'll put my notes in your comment where you directly reference it.

test/src/tests-config-api-client.ts
alexs-mparticle: I think I'm missing something here, since the relevant code on [lines 118-127](https://github.com/mParticle/mparticle-web-sdk/blob/refactor/SQDSDKS-5010-config-api-client/src/configAPIClient.ts#L118) reference `dataPlan`, not `dataPlanResult`

src/configAPIClient.ts
rmi22186: SGTM

src/sdkRuntimeModels.ts
```diff
@@ -158,10 +158,13 @@ export interface MParticleWebSDK {
     isIOS?: boolean;
 }
 
+// Used in cases where server requires booleans as strings
+export type BooleanStringLowerCase = 'false' | 'true';
+export type BooleanStringUpperCase = 'False' | 'True';
```
rmi22186, line 163: This is technically `TitleCase` Uppercase would be `FALSE` and `TRUE`

rmi22186 (changes requested):
final nit

rmi22186 (approved):

PR #383, opened 2023-01-18
feat: Add Support for Batch Persistence when Offline 

alexs-mparticle:
## Instructions

 ## Summary
 - As events are queued, they are placed into local storage for later processing. Once the prepareAndUpload method is triggered, events are removed from local storage and converted to Batches, then placed into the batch vault.
 - This pull request also refactors parts of the BatchUploader to better handle asynchronous uploads. Currently, a single asynchronous request is made that iterates through a list of batches, yet each batch is individually uploaded to the server. Loops inside of promises are not recommended in JavaScript as they could lead a longer turnaround time to process the response or other unexpected results. By breaking up the single upload promise into individual promises, each upload is a single process that returns quicker requires less processing.
 - Also in this PR are updates to our event conversion logic so that we can make sure that consent state is not lost when deserializing events and batches when storing in local storage.
 - Lastly, Beacon Uploader tests are moved out of the Batch Uploader test suite as there was an issue with race condition in the test runner.

 ## Testing Plan

 - Closes https://go.mparticle.com/work/SQDSDKS-4716

### DIFF
--- a/src/batchUploader.ts
+++ b/src/batchUploader.ts
@@ -8,18 +8,36 @@ import {
 import { convertEvents } from './sdkToEventsApiConverter';
 import Types from './types';
 import { isEmpty } from './utils';
+import Vault from './vault';
 
+/**
+ * BatchUploader contains all the logic to store/retrieve events and batches
+ * to/from persistence, and upload batches to mParticle. It queues events as
+ * they come in, and at set intervals, retrieves events from persistence and
+ * turns them into batches. It then attempts to upload them to mParticle.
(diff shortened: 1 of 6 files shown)


rmi22186 (approved):

alexs-mparticle:
Closing in favor of #386 so we can wrap this feature in a feature flag

alexs-mparticle:
Closing in favor of #386 as this wraps the feature with a feature flag and is more reduced in scope.

PR #384, opened 2023-01-30
fix: use currentUser consent state as backup if missing in last event

alexs-mparticle:
## Instructions

 ## Summary
 - In cases where the event is missing a consent state, we should use the current user's consent state

 ## Testing Plan

 - Closes https://go.mparticle.com/work/SQDSDKS-5047
 - Broken out of #383 to split out unrelated functionality in that feature branch

### DIFF
--- a/src/sdkToEventsApiConverter.ts
+++ b/src/sdkToEventsApiConverter.ts
@@ -49,8 +49,7 @@ export function convertEvents(
 
     let currentConsentState: SDKConsentState = null;
 
-    // If the event is missing consent state for some reason,
-    // get consent state from last event or user
+    // Add the consent state from either the Last Event or the user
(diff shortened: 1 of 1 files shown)


src/sdkToEventsApiConverter.ts
```diff
@@ -44,6 +47,16 @@ export function convertEvents(
         return null;
     }
 
+    let currentConsentState: SDKConsentState = null;
+
+    // If the event is missing consent state for some reason,
+    // get consent state from last event or user
```
rmi22186, line 53:
```suggestion
    // Add the consent state from either the Last Event or the user
```

rmi22186 (changes requested):
small nit for comment clarification

rmi22186 (approved):

PR #385, opened 2023-01-30
refactor: Modify BatchUploader to use promises for individual uploads

alexs-mparticle:
## Instructions

 ## Summary
This pull request refactors parts of the BatchUploader to better handle asynchronous uploads. Currently, a single asynchronous request is made that iterates through a list of batches, yet each batch is individually uploaded to the server. Loops inside of promises are not recommended in JavaScript as they could lead a longer turnaround time to process the response or other unexpected results. By breaking up the single upload promise into individual promises, each upload is a single process that returns quicker requires less processing.

Also, Beacon Uploader tests are moved out of the Batch Uploader test suite as there was an issue with race condition in the test runner.

 ## Testing Plan

 - Closes https://go.mparticle.com/work/SQDSDKS-5048
 - Breaks out non-persistence related functionality from #383

### DIFF
--- a/src/batchUploader.ts
+++ b/src/batchUploader.ts
@@ -9,17 +9,35 @@ import { convertEvents } from './sdkToEventsApiConverter';
 import Types from './types';
 import { isEmpty } from './utils';
 
+/**
+ * BatchUploader contains all the logic to upload batches to mParticle.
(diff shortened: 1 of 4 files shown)


rmi22186 (approved):
Tiny comment update otherwise LGTM

rmi22186 (approved):

PR #386, opened 2023-02-01
feat: Add Feature Flag Support for Offline Batching

alexs-mparticle:
## Instructions

 ## Summary
- As events are queued, they are placed into local storage for later processing. Once the prepareAndUpload method is triggered, events are removed from local storage and converted to Batches, then placed into the batch vault.
 - Implements a Feature Flag so we can control offline batching release to users via a server config

 ## Testing Plan

 - Closes https://go.mparticle.com/work/SQDSDKS-5049
 - Reintroduces approved code from #373

### DIFF
--- a/src/batchUploader.ts
+++ b/src/batchUploader.ts
@@ -1,4 +1,5 @@
 import { Batch } from '@mparticle/event-models';
+import Constants from './constants';
 import {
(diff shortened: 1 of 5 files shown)


alexs-mparticle:
Closed in favor of #402

PR #390, opened 2023-02-13
refactor: Create Unit tests and TypeScript Interfaces for SessionManager

alexs-mparticle:
## Instructions

 ## Summary
 - Create Unit tests and TypeScript Interfaces for SessionManager

 ## Testing Plan
 - Not necessary as this PR only contains updated tests and interfaces that are not compiled.

 - Closes https://go.mparticle.com/work/SQDSDKS-5091

### DIFF
--- a/src/sessionManager.js
+++ b/src/sessionManager.js
@@ -101,6 +101,7 @@ function SessionManager(mpInstance) {
 
             cookies = mpInstance._Persistence.getPersistence();
 
+            // TODO: https://go.mparticle.com/work/SQDSDKS-5684
             if (!cookies) {
                 return;
             }
--- a/src/sessionManager.js
+++ b/src/sessionManager.js
@@ -142,6 +142,10 @@ function SessionManager(mpInstance) {
                 }
             }
         } else {
+            // At this moment, an AbandonedEndSession is defined when on of three things occurs:
+            // - the SDK's store is not enabled because mParticle.setOptOut was called
+            // - the devToken is undefined
+            // - webviewBridgeEnabled is set to false
             mpInstance.Logger.verbose(
                 Messages.InformationMessages.AbandonEndSession
             );
--- a/src/sessionManager.js
+++ b/src/sessionManager.js
@@ -35,6 +35,7 @@ function SessionManager(mpInstance) {
         }
     };
 
+    // TODO: Rename to getSessionId
     this.getSession = function() {
         return mpInstance._Store.sessionId;
     };
--- a/test/src/tests-session-manager.ts
+++ b/test/src/tests-session-manager.ts
@@ -86,7 +86,7 @@ describe('SessionManager', () => {
             expect(mpInstance._Store.sessionId).to.equal('TEST-UNIQUE-ID');
         });
 
-        it('resumes the previous session if session ID exists and dateLastSent is within the timeoutWindow', () => {
+        it('resumes the previous session if session ID exists and dateLastSent is within the timeout window', () => {
             const sessionTimeoutInMilliseconds = 8 * 60000; // 11 minutes
 
             mParticle.init(apiKey, window.mParticle.config);
--- a/test/src/tests-session-manager.ts
+++ b/test/src/tests-session-manager.ts
@@ -574,7 +574,7 @@ describe.only('SessionManager', () => {
             expect(startNewSessionSpy.called).to.equal(true);
         });
 
-        it('should NOT call startNewSession if Store.sessionId and Persistence is null', () => {
+        it('should NOT call startNewSession if Store.sessionId and Persistence are null', () => {
             mParticle.init(apiKey, window.mParticle.config);
             const mpInstance = mParticle.getInstance();
 
--- a/test/src/tests-session-manager.ts
+++ b/test/src/tests-session-manager.ts
@@ -387,9 +387,7 @@ describe('SessionManager', () => {
                 );
             });
 
-            // QUESTION: What exactly defines an "AbandonedEndSession"?
-            // FIXME: Test only passes if specific cases are set
-            it.skip('should log an AbandonedEndSession message if webviewBridgeEnabled is not enabled', () => {
+            it('should log AbandonEndSession message if webviewBridgeEnabled is false but Store.isEnabled is true and devToken is undefined', () => {
                 mParticle.init(apiKey, window.mParticle.config);
 
                 const mpInstance = mParticle.getInstance();
@@ -399,23 +397,20 @@ describe('SessionManager', () => {
 
                 const consoleSpy = sinon.spy(mpInstance.Logger, 'verbose');
 
-                // FIXME: This test only passes if devToken is undefined
-                //        If it is set, webviewBridgeEnabled is ignored
+                mpInstance._Store.isEnabled = true;
                 mpInstance._Store.webviewBridgeEnabled = false;
                 mpInstance._Store.devToken = undefined;
 
                 mpInstance._SessionManager.endSession();
 
-                // Should log initial StartingEndSession and AbandonEndSession messagesk
+                // Should log initial StartingEndSession and AbandonEndSession messages
                 expect(consoleSpy.getCalls().length).to.equal(2);
                 expect(consoleSpy.lastCall.firstArg).to.equal(
                     Messages.InformationMessages.AbandonEndSession
                 );
             });
 
-            // QUESTION: What exactly defines an "AbandonedEndSession"?
-            // FIXME: Test only passes if specific cases are set
-            it.skip('should log an AbandonedEndSession message if devToken is not set', () => {
+            it('should log NoSessionToEnd message if webviewBridgeEnabled and Store.isEnabled are true but devToken is undefined', () => {
                 mParticle.init(apiKey, window.mParticle.config);
 
                 const mpInstance = mParticle.getInstance();
@@ -425,20 +420,20 @@ describe('SessionManager', () => {
 
                 const consoleSpy = sinon.spy(mpInstance.Logger, 'verbose');
 
-                // FIXME: Test only passes if devToken is undefined and webviewBridgeEnabled is false
+                mpInstance._Store.isEnabled = true;
                 mpInstance._Store.devToken = undefined;
-                mpInstance._Store.webviewBridgeEnabled = false;
+                mpInstance._Store.webviewBridgeEnabled = true;
 
                 mpInstance._SessionManager.endSession();
 
-                // Should log initial StartingEndSession and AbandonEndSession messagesk
+                // Should log initial StartingEndSession and NoSessionToEnd messages
                 expect(consoleSpy.getCalls().length).to.equal(2);
                 expect(consoleSpy.lastCall.firstArg).to.equal(
-                    Messages.InformationMessages.AbandonEndSession
+                    Messages.InformationMessages.NoSessionToEnd
                 );
             });
 
-            it('should priortize cookie sessionId over Store.sessionId', () => {
+            it('should prioritize cookie sessionId over Store.sessionId', () => {
                 mParticle.init(apiKey, window.mParticle.config);
 
                 const mpInstance = mParticle.getInstance();
@@ -481,6 +476,9 @@ describe('SessionManager', () => {
 
                 mpInstance._SessionManager.endSession();
 
+                // We are verifying that the session has not ended, and therefore the
+                // session ID should still be the same, as opposed to null, which
+                // is assigned when the session actually ends
                 expect(mpInstance._Store.sessionId).to.equal('fake-session-id');
 
                 // When session is not timed out, setSessionTimer is called to keep track
--- a/test/src/tests-session-manager.ts
+++ b/test/src/tests-session-manager.ts
@@ -387,7 +387,6 @@ describe('SessionManager', () => {
             const mpInstance = mParticle.getInstance();
 
             // Manually clear global timer since it gets set during init
-            // window.clearTimeout(mpInstance._Store.globalTimer);
             mpInstance._Store.globalTimer = null;
             expect(mpInstance._Store.globalTimer).to.eq(null);
 
--- a/test/src/tests-session-manager.ts
+++ b/test/src/tests-session-manager.ts
@@ -9,7 +9,7 @@ import {
     MessageType,
     MILLISECONDS_IN_ONE_MINUTE,
 } from './config';
-import { IdentityApiData } from '@mparticle/web-sdk';
+import { IdentityApiData, endSession } from '@mparticle/web-sdk';
 import Constants from '../../src/constants';
 
 const { Messages } = Constants;
@@ -43,7 +43,7 @@ describe('SessionManager', () => {
         ]);
     });
 
-    afterEach(function() {
+    afterEach(function () {
         sandbox.restore();
         clock.restore();
         mockServer.restore();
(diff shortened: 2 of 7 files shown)


test/src/tests-session-manager.ts
```diff
@@ -0,0 +1,607 @@
...
+            expect(identifySpy.getCall(0).args[0]).to.eql(identityApiData);
+            expect(identifySpy.getCall(0).args[1]).to.eql(callbackSpy);
+            expect(callbackSpy.called).to.equal(true);
+            expect(mpInstance._Store.identifyCalled).to.eql(true);
+            expect(mpInstance._Store.SDKConfig.identityCallback).to.eql(null);
+        });
+    });
+
+    describe('#endSession', () => {
+        // TODO: Set up scaffolding for session
```
rmi22186, line 248: Does this still need to be done?

test/src/tests-session-manager.ts
```diff
@@ -0,0 +1,607 @@
...
+        it.skip('should end a session', () => {
```
rmi22186, line 249: Should this be unskipped?

test/src/tests-session-manager.ts
```diff
@@ -0,0 +1,607 @@
...
+
+            mpInstance._SessionManager.endSession();
+
+            expect(mpInstance._Store.sessionId).to.equal(null);
+            expect(mpInstance._Store.dateLastEventSent).to.equal(null);
+            expect(mpInstance._Store.sessionAttributes).to.eql({});
+
+            // Persistence isn't necessary for this feature, but we should test
+            // to see that it is called in case this ever needs to be refactored
+            expect(persistenceSpy.called).to.equal(true);
```
rmi22186, line 262: 👍 love it

test/src/tests-session-manager.ts
```diff
@@ -0,0 +1,607 @@
...
+        it.skip('should end a session', () => {
+            mParticle.init(apiKey, window.mParticle.config);
+            const mpInstance = mParticle.getInstance();
+            const persistenceSpy = sinon.spy(mpInstance._Persistence, 'update');
+
+            mpInstance._SessionManager.endSession();
+
+            expect(mpInstance._Store.sessionId).to.equal(null);
+            expect(mpInstance._Store.dateLastEventSent).to.equal(null);
+            expect(mpInstance._Store.sessionAttributes).to.eql({});
```
rmi22186, line 258: I think to be more clear, these should be shown to have actual values prior to line 254.

test/src/tests-session-manager.ts
```diff
@@ -0,0 +1,607 @@
...
+
+            // Persistence isn't necessary for this feature, but we should test
+            // to see that it is called in case this ever needs to be refactored
+            expect(persistenceSpy.called).to.equal(true);
+        });
+
+        it('should do nothing when a session does not exist', () => {});
+        it('should do nothing when persistence cookies are missing', () => {});
+        it('should nullify session attributes when override is used', () => {});
+        it('should use the cookie sessionId over Store.sessionId', () => {});
```
rmi22186, line 284: Should there be tests here?

test/src/tests-session-manager.ts
```diff
@@ -0,0 +1,607 @@
...
+
+        it('should NOT end session if session has not timed out', () => {
```
rmi22186, line 286: Perhaps add a comment that 20 minutes ago is less than the default session time out of 30.

test/src/tests-session-manager.ts
```diff
@@ -0,0 +1,607 @@
...
+            mpInstance._SessionManager.endSession();
+
+            expect(mpInstance._Store.sessionId).to.equal('fake-session-id');
+
+            // When session is not timed out, setSessionTimer is called to keep track
+            // of current session timeout
+            expect(timerSpy.getCalls().length).to.equal(1);
+        });
+
+        it('should end session if session times out', () => {
```
rmi22186, line 316: I see what you are doing here, but I think the approach is a bit off.  By calling `endSession` you are not letting the session time out on its own, but instead forcing a session to end.

Instead, you should initialize mParticle, check to see if `sessionId`, `dateLastEventSet` exist, and then move the clock forward (`clock.tick(500000)`) or whatever number it is that exceeds the default session timeout. Then you can call the expects you have on lines 336-342 again.

You can keep stubbing persistence as needed though in line 327.

test/src/tests-session-manager.ts
```diff
@@ -0,0 +1,607 @@
...
+
+            // Start Timer
+            mpInstance._SessionManager.setSessionTimer();
+
+            // Progress 29 minutes to make sure end session has not fired
+            clock.tick(29 * 60000);
+            expect(endSessionSpy.called).to.equal(false);
+
+            // Progress one minutes to make sure end session fires
+            clock.tick(30 * 60000);
```
rmi22186, line 363: This isn't progressing one more minute, it's progressing 30 more minutes.  Still works, but the comment is off, or the code is off.
```suggestion
            clock.tick(1 * 60000);
```

test/src/tests-session-manager.ts
```diff
@@ -0,0 +1,607 @@
...
+        });
+
+        it('should set a global timer', () => {
+            mParticle.init(apiKey, window.mParticle.config);
+            const mpInstance = mParticle.getInstance();
+
+            // Manually clear global timer since it gets set during init
+            // window.clearTimeout(mpInstance._Store.globalTimer);
+            mpInstance._Store.globalTimer = null;
+            expect(mpInstance._Store.globalTimer).to.eq(null);
```
rmi22186, line 377: Should line 375 be uncommented, and should line 376 be deleted?

The way I see it, 376 sets a null, then 377 checks if it's null, which is a bit superfluous.  If you uncomment line 375 and delete 376, does the test still pass?

test/src/tests-session-manager.ts
```diff
@@ -0,0 +1,607 @@
...
+                'setSessionTimer'
+            );
+
+            mpInstance._SessionManager.resetSessionTimer();
+            expect(startNewSessionSpy.called).to.equal(false);
+            expect(clearSessionTimeoutSpy.called).to.equal(false);
+            expect(setSessionTimerSpy.called).to.equal(false);
+        });
+
+        it('should reset session timer', () => {
```
rmi22186, line 440: this is a bit generic of a test description.  Should it be `should call clearSessionTimeout and setSessionTimer`?

test/src/tests-session-manager.ts
```diff
@@ -0,0 +1,607 @@
...
+        it('should NOT call startNewSession if sessionId is defined and Persistence is undefined', () => {
+            mParticle.init(apiKey, window.mParticle.config);
+            const mpInstance = mParticle.getInstance();
+
+            // Session Manager relies on persistence check sid (Session ID)
+            // However, if persistence is undefined, this will not create a
+            // new session
+            sinon.stub(mpInstance._Persistence, 'getPersistence').returns(null);
+
+            mpInstance._Store.sessionId = null;
```
rmi22186, line 519: `sessionId` is `null` here, but the test description says sessionId should be defined.  Should you set this to a string?

test/src/tests-session-manager.ts
```diff
@@ -0,0 +1,607 @@
...
+            mpInstance._Store.sessionId = undefined;
+
+            mpInstance._SessionManager.startNewSessionIfNeeded();
+
+            mpInstance._Store.sessionId = 'sid-from-persistence';
+
+            expect(startNewSessionSpy.called).to.equal(true);
+        });
+
+        it('should NOT call startNewSession if Store.sessionId and Persistence is null', () => {
```
rmi22186, line 573:
```suggestion
        it('should NOT call startNewSession if Store.sessionId and Persistence are null', () => {
```

rmi22186 (changes requested):
Great work adding all this test coverage!  I have a bunch of comments. Happy to discuss over zoom after you do a quick review yourself.

src/sessionManager.js
```diff
@@ -35,6 +35,7 @@ function SessionManager(mpInstance) {
         }
     };
 
+    // TODO: Rename to getSessionId
```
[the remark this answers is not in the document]
alexs-mparticle: This is more relevant to #707, so comments will be addressed there.

test/src/tests-session-manager.ts
```diff
@@ -0,0 +1,607 @@
...
+            generateUniqueIdSpy.returns('test-unique-id');
+
+            mParticle.init(apiKey, window.mParticle.config);
+            const mpInstance = mParticle.getInstance();
+
+            mpInstance._Store.sessionId = '';
+            expect(mpInstance._Store.sessionId).to.be.empty;
+
+            mpInstance._SessionManager.initialize();
+            expect(mpInstance._Store.sessionId).to.equal('TEST-UNIQUE-ID');
```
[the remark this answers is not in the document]
alexs-mparticle: For posterity: `startNewSession` capitalizes whatever comes back from `generateUniqueId`. For reasons.

test/src/tests-session-manager.ts
alexs-mparticle: To be handled as part of https://go.mparticle.com/work/SQDSDKS-5614

test/src/tests-session-manager.ts
alexs-mparticle: There was a typo in the test suite label. it should have been `should NOT call startNewSession if sessionId is undefined and Persistence is undefined`

test/src/tests-session-manager.ts
alexs-mparticle: In my testing, `window.clearTimeout(mpInstance._Store.globalTimer);` doesn't actually clear out the timer. I think that's why I commented it out in the first place and forgot to delete it.

test/src/tests-session-manager.ts
rmi22186: Tiiiiny nit: I would say "not defined", since you are setting to null, which is not defined, and not `undefined`.  Or change the `null`s to `undefined` and then the test description is fine.

rmi22186 (changes requested):
Mostly nits. We are close!

test/src/tests-session-manager.ts
```diff
@@ -0,0 +1,622 @@
...
+            mpInstance._SessionManager.endSession();
+
+            expect(mpInstance._Store.sessionId).to.equal(null);
+            expect(mpInstance._Store.dateLastEventSent).to.equal(null);
+            expect(mpInstance._Store.sessionAttributes).to.eql({});
+
+            // Persistence isn't necessary for this feature, but we should test
+            // to see that it is called in case this ever needs to be refactored
+            expect(persistenceSpy.called).to.equal(true);
+        });
```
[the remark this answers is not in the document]
alexs-mparticle: I think I understand your reasoning, but my intention is to have these tests be as true "unit tests" as we possibly can with the SDK. I think testing the behavior of ticking the clock forward is very much an integration test, and possibly should be moved to a separate section of this test suite. However, I see value in highlighting expected behavior of calling `endSession` in all it's permutations.

I think for now I will separate Unit and Integration tests, and add the clock ticking test as an Integration test, since it does not directly rely on calling `endSession`. We can then group similar tests for clarity in the future.

test/src/tests-session-manager.ts
```diff
@@ -0,0 +1,860 @@
...
+                const mpInstance = mParticle.getInstance();
+                sinon
+                    .stub(mpInstance._Persistence, 'getPersistence')
+                    .returns({ gs: {} });
+
+                const consoleSpy = sinon.spy(mpInstance.Logger, 'verbose');
+
+                mpInstance._SessionManager.endSession();
+
+                // Should log initial StartingEndSession and NoSessionToEnd messagesk
```
rmi22186, line 341:
```suggestion
                // Should log initial StartingEndSession and NoSessionToEnd messages
```

test/src/tests-session-manager.ts
```diff
@@ -0,0 +1,860 @@
...
+                expect(consoleSpy.getCalls().length).to.equal(2);
+                expect(consoleSpy.lastCall.firstArg).to.equal(
+                    Messages.InformationMessages.NoSessionToEnd
+                );
+            });
+
+            it('should log an AbandonedEndSession message if unable to log', () => {
```
rmi22186, line 348: it's a little bit clear the use of log at the end of this test statement.  maybe the following is more explicit
```suggestion
            it('should log an AbandonedEndSession message if SDK canLog() returns false', () => {
```

test/src/tests-session-manager.ts
```diff
@@ -0,0 +1,860 @@
...
+                    .stub(mpInstance._Persistence, 'getPersistence')
+                    .returns({ gs: {} });
+
+                sinon.stub(mpInstance._Helpers, 'canLog').returns(false);
+
+                const consoleSpy = sinon.spy(mpInstance.Logger, 'verbose');
+
+                mpInstance._SessionManager.endSession();
+
+                // Should log initial StartingEndSession and AbandonEndSession messagesk
```
rmi22186, line 362:
```suggestion
                // Should log initial StartingEndSession and AbandonEndSession messages
```

test/src/tests-session-manager.ts
```diff
@@ -0,0 +1,860 @@
...
+                expect(consoleSpy.getCalls().length).to.equal(2);
+                expect(consoleSpy.lastCall.firstArg).to.equal(
+                    Messages.InformationMessages.AbandonEndSession
+                );
+            });
+
+            it('should log an AbandonedEndSession message if Store is not enabled', () => {
```
rmi22186, line 369: Wording is a bit confusing here. it's not that Store is or isn't enabled, it's that the flag isEnabled is `false`.  It looks like isEnabled is only set when `optOut` is called. The flag itself is definitely not well named, and maybe we should address that in a future refactor to be more clear as to what is going on (is SDK logging enabled or not).

```suggestion
            it('should log an AbandonedEndSession message if Store.isEnabled is false', () => {
```

test/src/tests-session-manager.ts
```diff
@@ -0,0 +1,860 @@
...
+                mpInstance._SessionManager.endSession();
+
+                // Should log initial StartingEndSession and AbandonEndSession messagesk
+                expect(consoleSpy.getCalls().length).to.equal(2);
+                expect(consoleSpy.lastCall.firstArg).to.equal(
+                    Messages.InformationMessages.AbandonEndSession
+                );
+            });
+
+            // QUESTION: What exactly defines an "AbandonedEndSession"?
```
rmi22186, line 390: Searching through out codebase, we have things similar messages like `AbandonLogEvent`, `AbandonAliasUsers`, `AbandonStartSession`, and `AbandonEndSession`.  They all happen if logging is disabled or developer token is not sent.  

Although there is also `LoggingDisabled`, which we should audit and determine what the difference is.

test/src/tests-session-manager.ts
```diff
@@ -0,0 +1,860 @@
...
+                mpInstance._SessionManager.endSession();
+
+                // Should log initial StartingEndSession and AbandonEndSession messagesk
+                expect(consoleSpy.getCalls().length).to.equal(2);
+                expect(consoleSpy.lastCall.firstArg).to.equal(
+                    Messages.InformationMessages.AbandonEndSession
+                );
+            });
+
+            it('should priortize cookie sessionId over Store.sessionId', () => {
```
rmi22186, line 441: I think it's fine to keep this unit test, but I think in real life scenarios, these are always in sync.  If you found a case where it isn't, then that's a bug.

test/src/tests-session-manager.ts
```diff
@@ -0,0 +1,860 @@
...
+                twentyMinutesAgo.setMinutes(now.getMinutes() - 20);
+
+                mParticle.init(apiKey, window.mParticle.config);
+                const mpInstance = mParticle.getInstance();
+                const timerSpy = sinon.spy(
+                    mpInstance._SessionManager,
+                    'setSessionTimer'
+                );
+
+                // Session Manager relies on persistence to determine last time seen (LES)
```
rmi22186, line 473:
```suggestion
                // Session Manager relies on persistence to determine last event sent (LES) time
```

test/src/tests-session-manager.ts
```diff
@@ -0,0 +1,860 @@
...
+                sinon.stub(mpInstance._Persistence, 'getPersistence').returns({
+                    gs: {
+                        les: twentyMinutesAgo,
+                        sid: 'fake-session-id',
+                    },
+                });
+
+                mpInstance._SessionManager.endSession();
+
+                expect(mpInstance._Store.sessionId).to.equal('fake-session-id');
```
rmi22186, line 484: I was a little thrown off by this, but I realize that if the session ended, the `sessionId` would be `null`.  Maybe add that as a comment as to what you are testing here, or add an `expect` right before this line to say that the `sessionId` doesn't equal `null`.

test/src/tests-session-manager.ts
```diff
@@ -0,0 +1,860 @@
...
+
+                // When session is not timed out, setSessionTimer is called to keep track
+                // of current session timeout
+                expect(timerSpy.getCalls().length).to.equal(1);
```
rmi22186, line 488: Can you add a test for what the message should be here?

test/src/tests-session-manager.ts
```diff
@@ -0,0 +1,860 @@
...
+                // Init will set dateLastEventSent to now, but endSession relies on the persistence layer
+                expect(mpInstance._Store.dateLastEventSent).to.eql(now);
+                expect(mpInstance._Store.sessionAttributes).to.eql({});
+
+                const persistenceUpdateSpy = sinon.spy(
+                    mpInstance._Persistence,
+                    'update'
+                );
+
+                // Session Manager relies on persistence to determine last time seen (LES)
```
rmi22186, line 517:
```suggestion
                // Session Manager relies on persistence to determine last event seen (LES) time
```

test/src/tests-session-manager.ts
```diff
@@ -0,0 +1,860 @@
...
+                );
+
+                // Start Timer
+                mpInstance._SessionManager.setSessionTimer();
+
+                // Progress 29 minutes to make sure end session has not fired
+                clock.tick(29 * MILLISECONDS_IN_ONE_MINUTE);
+                expect(endSessionSpy.called).to.equal(false);
+
+                // Progress one minutes to make sure end session fires
```
rmi22186, line 573:
```suggestion
                // Progress one minute to make sure end session fires
```

test/src/tests-session-manager.ts
```diff
@@ -0,0 +1,860 @@
...
+                    mpInstance._SessionManager,
+                    'startNewSession'
+                );
+
+                mpInstance._SessionManager.startNewSessionIfNeeded();
+                expect(startNewSessionSpy.called).to.equal(false);
+            });
+        });
+    });
+    describe('Integration Tests', () => {
```
rmi22186, line 819:
```suggestion
    });

    describe('Integration Tests', () => {
```

rmi22186 (changes requested):
mostly minor nits

test/src/tests-session-manager.ts
```diff
@@ -0,0 +1,860 @@
...
+                    .stub(mpInstance._Persistence, 'getPersistence')
+                    .returns(null);
+
+                const consoleSpy = sinon.spy(mpInstance.Logger, 'verbose');
+
+                mpInstance._SessionManager.endSession();
+
+                // This path currently goes through the !cookies path,
+                // which doesn't really return anything except for logging the
+                // initial StartingEndSession message
```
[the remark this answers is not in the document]
alexs-mparticle: My comment was based off the fact that this code path is literally: `if (!cookies) return;`, and the only indication something actually happens is that it logs a console message at the very beginning of the method.

rmi22186 (approved):

rmi22186:
besides ensuring tests pass, and the [1 code smell that is legitimate](https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&pullRequest=390&id=mParticle_mparticle-web-sdk&open=AYopcKydstFJzATpKBi2), this can be merged to `development`

PR #391, opened 2023-02-14
fix: removeUserAttribute and forwarder filters fix

mmustafa-tse:
## Instructions

 ## Summary
 - User attributes when removes are not respecting the filters set which leads some forwarder to set them as null even though they're filtered. The fix here is to not fire removeUserAttribute for a filtered UA.

 ## Testing Plan
- Was tested locally via local overrides and passed unit testing

## References
- Closes https://mparticle-eng.atlassian.net/browse/PRODRDMP-5414

### DIFF
--- a/src/forwarders.js
+++ b/src/forwarders.js
@@ -404,8 +404,8 @@ export default function Forwarders(mpInstance, kitBlocker) {
         mpInstance._Store.activeForwarders.forEach(function(forwarder) {
             const forwarderFunction = forwarder[functionNameKey];
             if (
-                forwarderFunction &&
-                mpInstance._Helpers.isFilteredUserAttribute(
+                !forwarderFunction ||
+                !mpInstance._Helpers.isFilteredUserAttribute(
                     key,
                     forwarder.userAttributeFilters
                 )
--- a/src/forwarders.js
+++ b/src/forwarders.js
@@ -3,6 +3,12 @@ import filteredMparticleUser from './filteredMparticleUser';
 
 export default function Forwarders(mpInstance, kitBlocker) {
     var self = this;
+
+    const UserAttributeActionTypes = {
+        setUserAttribute: 'setUserAttribute',
+        removeUserAttribute: 'removeUserAttribute'
+    }
+
     this.initForwarders = function(userIdentities, forwardingStatsCallback) {
         var user = mpInstance.Identity.getCurrentUser();
         if (
@@ -387,14 +393,14 @@ export default function Forwarders(mpInstance, kitBlocker) {
         }
     };
 
-    this.onHandleForwarderUserAttributes = function(key, value, functionName) {
+    this.onHandleForwarderUserAttributes = function(key, value, functionNameKey) {
         if (kitBlocker && kitBlocker.isAttributeKeyBlocked(key)) {
             return;
         }
 
         if (mpInstance._Store.activeForwarders.length) {
             mpInstance._Store.activeForwarders.forEach(function(forwarder) {
-                var forwarderFunction = forwarder[functionName];
+                var forwarderFunction = forwarder[functionNameKey];
                 if (
                     forwarderFunction &&
                     forwarder.userAttributeFilters &&
@@ -406,9 +412,9 @@ export default function Forwarders(mpInstance, kitBlocker) {
                     try {
                         var result;
 
-                        if (functionName === 'setUserAttribute') {
+                        if (functionNameKey === UserAttributeActionTypes.setUserAttribute) {
                             result = forwarder.setUserAttribute(key, value);
-                        } else if (functionName === 'removeUserAttribute') {
+                        } else if (functionNameKey === UserAttributeActionTypes.removeUserAttribute) {
                             result = forwarder.removeUserAttribute(key);
                         }
 
--- a/src/forwarders.js
+++ b/src/forwarders.js
@@ -6,8 +6,8 @@ export default function Forwarders(mpInstance, kitBlocker) {
 
     const UserAttributeActionTypes = {
         setUserAttribute: 'setUserAttribute',
-        removeUserAttribute: 'removeUserAttribute'
-    }
+        removeUserAttribute: 'removeUserAttribute',
+    };
 
     this.initForwarders = function(userIdentities, forwardingStatsCallback) {
         var user = mpInstance.Identity.getCurrentUser();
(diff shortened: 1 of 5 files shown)


src/forwarders.js
```diff
@@ -387,23 +387,30 @@ export default function Forwarders(mpInstance, kitBlocker) {
         }
     };
 
-    this.callSetUserAttributeOnForwarders = function(key, value) {
+    this.onHandleForwarderUserAttributes = function(key, value, functionName) {
```
alexs-mparticle, line 390: Probably a nitpick for me, but I think calling this `functionNameKey` or `functionNameAsKey` is more descriptive since we're using it as a lookup.

src/forwarders.js
```diff
@@ -387,23 +387,30 @@ export default function Forwarders(mpInstance, kitBlocker) {
...
                     !mpInstance._Helpers.inArray(
                         forwarder.userAttributeFilters,
                         mpInstance._Helpers.generateHash(key)
                     )
                 ) {
                     try {
-                        var result = forwarder.setUserAttribute(key, value);
+                        var result;
+
+                        if (functionName === 'setUserAttribute') {
```
alexs-mparticle, line 409: `setUserAttribute` and `removeUserAttribute` should be made into constants so we can know if we made a typo.

alexs-mparticle (changes requested):

src/forwarders.js
```diff
@@ -387,23 +393,40 @@ export default function Forwarders(mpInstance, kitBlocker) {
         }
     };
 
-    this.callSetUserAttributeOnForwarders = function(key, value) {
+    this.onHandleForwarderUserAttributes = function(
```
rmi22186, line 396: Nit - I feel like this should be `handleForwarderUserAttributes` instead of `onHandleForwarderUserAttributes`.  `on` something is more an effect, almost like a callback (but not quite), or something that is not directly associated with it the verb it is `on`-ing if that makes sense.  So in our kit code, it is `onHandle` or `onUserIdentified` because the intent when I call `mParticle.....setUserAttribute` is to change it in our system...and then the the kit has the downstream effect from the core.  it's a bit nuanced (and perhaps my opinion) but in the core SDK, i don't think we use `on` anything....otherwise we'd see `on-Verb` a lot more.

alexs-mparticle: Agreed. The `on` prefix for me is usually associated with UI functionality, which feels misleading in this context.

mmustafa-tse: Sounds good, will implement changes, thanks for feedback!

mmustafa-tse: @rmi22186 and @alexs-mparticle, I was able to take off one unnecessary `if` statement and make a helper function to reduce the messy lines to check if attribute is filtered as requested, and changed `onHandleForwarderUserAttributes` to `handleForwarderUserAttributes`. Let me know if anything else is needed.

src/forwarders.js
```diff
@@ -387,33 +393,46 @@ export default function Forwarders(mpInstance, kitBlocker) {
...
+            const forwarderFunction = forwarder[functionNameKey];
+            if (
+                forwarderFunction &&
+                mpInstance._Helpers.isFilteredUserAttribute(
+                    key,
+                    forwarder.userAttributeFilters
+                )
+            ) {
+                try {
+                    var result;
```
alexs-mparticle, line 414:
```suggestion
                    let result;
```

src/helpers.js
```diff
@@ -387,6 +387,14 @@ export default function Helpers(mpInstance) {
         return filteredUserAttributes;
     };
 
+    this.isFilteredUserAttribute = function(userAttributeKey, filterList) {
+        const hashedUserAttribute = self.generateHash(userAttributeKey);
+        if (filterList && !self.inArray(filterList, hashedUserAttribute)) {
+            return true;
+        }
+        return false;
```
alexs-mparticle, line 395: Since you're returning a boolean for this, you can just return the whole conditional as the result, which will be a boolean.

```suggestion
        return filterList && !self.inArray(filterList, hashedUserAttribute));
```

src/forwarders.js
```diff
@@ -387,33 +393,46 @@ export default function Forwarders(mpInstance, kitBlocker) {
...
-                        forwarder.userAttributeFilters,
-                        mpInstance._Helpers.generateHash(key)
-                    )
-                ) {
-                    try {
-                        var result = forwarder.setUserAttribute(key, value);
+        mpInstance._Store.activeForwarders.forEach(function(forwarder) {
+            const forwarderFunction = forwarder[functionNameKey];
+            if (
+                forwarderFunction &&
```
alexs-mparticle, line 407: I think you can break out of this nested if by just returning after checking for the inverse:

```
if (
  !forwarderFunction &&
  !mpInstance._Helpers.isFilteredUserAttribute(
    key,
    forwarder.userAttributeFilters
  )
) {
  return;
}

**Review by alexs-mparticle — CHANGES_REQUESTED** [review]

**Comment by mmustafa-tse** [comment]
@alexs-mparticle changes added, thanks for the feedback!

File: src/forwarders.js (hunk 6 above)
Comment (mmustafa-tse) [reply src/forwarders.js]: @alexs-mparticle  also, I changed the operator here from `&&` to `||` since if it fails in either case we should continue with the next forwarder, correct me if I'm wrong.

**Review by alexs-mparticle — APPROVED** [review]
LGTM. Once @rmi22186 approves, we will prepare this for release.

**Review by rmi22186 — CHANGES_REQUESTED** [review]
Some small nits, but this is looking great Mo!

File: test/src/tests-forwarders.js [hunk 7]
```diff
@@ -1229,6 +1229,66 @@ describe('forwarders', function() {
...
+        mParticle.init(apiKey, window.mParticle.config);
+
+        // force filtered UA in mock forwarder (due to filtering affecting setUserAttribute) and test init
+        window.MockForwarder1.instance.userAttributes['weight'] = 150
+        mParticle.Identity.getCurrentUser().removeUserAttribute('weight');
+        mParticle.userAttributesFilterOnInitTest.should.have.property(
+            'weight'
+        );
+
+        mParticle.init(apiKey, window.mParticle.config);
```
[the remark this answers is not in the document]
mmustafa-tse: This is needed to test filter user attributes from forwarder on init for removeUserAttribute

mmustafa-tse:
@rmi22186 thanks for the feedback, implemented changes

rmi22186 (approved):
Looks awesome, @mmustafa-tse !   Great work on fixing this bug and leaving the code better than it was before.

🚢

PR #393, opened 2023-02-16
test: Fix flakey session date test

rmi22186:
## Instructions

 ## Summary
The previous test format would fail because it was looking for the difference in 2 unix timestamps and this comparison can be flakey depending on how quickly the test is running (if it runs fast and there is no difference in timestamp then it fails, and if it runs slowly there is a difference in timestamp and so the test passes).  After reviewing the test, it was not testing the proper things.  It should instead check to see what properties exists on the `actualEventObject` when it is a Session End. It also mutates the store, which we should test.

 ## Testing Plan
 - Ran unit tests over and over and it passed every time

 - Closes https://go.mparticle.com/work/SQDSDKS-5031

### DIFF
--- a/test/src/tests-serverModel.ts
+++ b/test/src/tests-serverModel.ts
@@ -625,8 +625,8 @@ describe('ServerModel', () => {
                 'currentSessionMPIDs'
             ).to.eql(['testMPID']);
 
-            // A SessionEnd event appends sessionLength
-            expect(actualEventObject).to.have.property('sessionLength');
+            // A SessionEnd event appends SessionLength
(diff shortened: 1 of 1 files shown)


alexs-mparticle (approved):
LGTM. I would rerun tests a few times to verify that it does not flake out.

rmi22186:
> LGTM. I would rerun tests a few times to verify that it does not flake out.

Yup I have re-run it several times locally.  It seems to have only ever failed in an action during release, and by removing that specific test where it compared actual timestamps, I'm confident in this test fix.

PR #394, opened 2023-02-23
fix: Verify Batches and events are uploaded in order of creation

alexs-mparticle:
## Instructions

 ## Summary
 - Events and Batches should be created and uploaded in the order they were created.
 - Write regression tests to verify this logic does not change drastically in any future code commits
 - Fix upload logic to interrupt transmission reattempts in case of any network interruption.

 ## Testing Plan

 - Closes https://go.mparticle.com/work/SQDSDKS-5104

### DIFF
--- a/src/batchUploader.ts
+++ b/src/batchUploader.ts
@@ -9,17 +9,35 @@ import { convertEvents } from './sdkToEventsApiConverter';
 import Types from './types';
 import { isEmpty } from './utils';
 
+/**
+ * BatchUploader contains all the logic to upload batches to mParticle.
+ * It queues events as they come in and at set intervals turns them into batches.
+ * It then attempts to upload them to mParticle.
+ *
+ * These uploads happen on an interval basis using window.fetch or XHR
+ * requests, depending on what is available in the browser.
+ *
+ * Uploads can also be triggered on browser visibility/focus changes via an
+ * event listener, which then uploads to mPartice via the browser's Beacon API.
+ */
+
 export class BatchUploader {
-    //we upload JSON, but this content type is required to avoid a CORS preflight request
+    // We upload JSON, but this content type is required to avoid a CORS preflight request
     static readonly CONTENT_TYPE: string = 'text/plain;charset=UTF-8';
     static readonly MINIMUM_INTERVAL_MILLIS: number = 500;
     uploadIntervalMillis: number;
-    pendingEvents: SDKEvent[];
-    pendingUploads: Batch[];
+    eventsQueuedForProcessing: SDKEvent[];
+    batchesQueuedForProcessing: Batch[];
     mpInstance: MParticleWebSDK;
     uploadUrl: string;
     batchingEnabled: boolean;
+    private uploader: AsyncUploader;
 
+    /**
+     * Creates an instance of a BatchUploader
+     * @param {MParticleWebSDK} mpInstance - the mParticle SDK instance
+     * @param {number} uploadInterval - the desired upload interval in milliseconds
+     */
     constructor(mpInstance: MParticleWebSDK, uploadInterval: number) {
         this.mpInstance = mpInstance;
         this.uploadIntervalMillis = uploadInterval;
@@ -28,8 +46,8 @@ export class BatchUploader {
         if (this.uploadIntervalMillis < BatchUploader.MINIMUM_INTERVAL_MILLIS) {
             this.uploadIntervalMillis = BatchUploader.MINIMUM_INTERVAL_MILLIS;
         }
-        this.pendingEvents = [];
-        this.pendingUploads = [];
+        this.eventsQueuedForProcessing = [];
+        this.batchesQueuedForProcessing = [];
 
         const { SDKConfig, devToken } = this.mpInstance._Store;
         const baseUrl = this.mpInstance._Helpers.createServiceUrl(
@@ -38,15 +56,20 @@ export class BatchUploader {
         );
         this.uploadUrl = `${baseUrl}/events`;
 
-        setTimeout(() => {
-            this.prepareAndUpload(true, false);
-        }, this.uploadIntervalMillis);
+        this.uploader = window.fetch
+            ? new FetchUploader(this.uploadUrl)
+            : new XHRUploader(this.uploadUrl);
+
+        this.triggerUploadInterval(true, false);
         this.addEventListeners();
     }
 
+    // Adds listeners to be used trigger Navigator.sendBeacon if the browser
+    // loses focus for any reason, such as closing browser tab or minimizing window
     private addEventListeners() {
         const _this = this;
 
+        // visibility change is a document property, not window
         document.addEventListener('visibilitychange', () => {
             _this.prepareAndUpload(false, _this.isBeaconAvailable());
         });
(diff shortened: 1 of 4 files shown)


test/src/tests-batchUploader.ts
```diff
@@ -89,6 +89,612 @@ describe('batch uploader', () => {
...
+
+            // // Force time to pass so that batch processing occurs
+            // clock.tick(1000);
+
+            // Batch 2
+            window.mParticle.logEvent('Test Event 1');
+            window.mParticle.logEvent('Test Event 2');
+            window.mParticle.logEvent('Test Event 3');
+
+            // Force an upload of events
```
rmi22186, line 568: I'd change this comment here and elsewhere to be a bit more specific because the upload process is very nuanced. 
 Something like the below:

```suggestion
            // Manually initiate the upload process - turn event into batches and upload the batch
```

test/src/tests-batchUploader.ts
```diff
@@ -89,6 +89,612 @@ describe('batch uploader', () => {
...
+            };
+
+            // debugger;
+
+            window.mParticle._resetForTests(MPConfig);
+            window.mParticle.init(apiKey, window.mParticle.config);
+            // Generates Batch 1 with Session Start + AST
+
+            // Adds a custom event to Batch 1
+            window.mParticle.logEvent('Test Event');
```
rmi22186, line 642: I'm just thinking about this and only commenting here, but this is relevant for everywhere in this file.  We shouldn't have `Test Event`, then `Test Event 1, 2, 3, 4, 5, 6`.  We should have `Test Event 1-7`.

rmi22186 (changes requested):
Looks good. Mostly nits about naming.

test/src/tests-batchUploader.ts
```diff
@@ -89,6 +89,612 @@ describe('batch uploader', () => {
...
+                eventsV3: '100',
+                eventBatchingIntervalMillis: 1000,
+            };
+
+            window.mParticle._resetForTests(MPConfig);
+            window.mParticle.init(apiKey, window.mParticle.config);
+
+            window.mParticle.logEvent('Test Event');
+
+            mockServer.requests.length.should.equal(1);
```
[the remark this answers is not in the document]
alexs-mparticle: At this point, the request should be the initial SessionStart/AST as part of the init and Custom Event. I don't believe there is an Identity call at this point when I test locally.

test/src/tests-batchUploader.ts
```diff
@@ -89,6 +89,612 @@ describe('batch uploader', () => {
...
+            // in this exact order
+            expect(batch1.events.length).to.equal(3);
+            expect(batch1.events[0].event_type).to.equal('session_start');
+            expect(batch1.events[1].event_type).to.equal(
+                'application_state_transition'
+            );
+            expect(batch1.events[2].data.event_name).to.equal('Test Event');
+
+            // // Force time to pass so that batch processing occurs
+            // clock.tick(1000);
```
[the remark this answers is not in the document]
alexs-mparticle: It should be removed. Forgot to kill it when I was cleaning up. Good catch.

test/src/tests-batchUploader.ts
alexs-mparticle: I think `Test Event 0` might be more appropriate. :laugh:

test/src/tests-batchUploader.ts
```diff
@@ -84,16 +84,603 @@ describe('batch uploader', () => {
...
+                now: new Date().getTime(),
+            });
+        });
+
+        afterEach(() => {
+            window.fetchMock.restore();
+            clock.restore();
+        });
+
+        it('should organize events and batches in the order they are processed and maintain that order when uploading', (done) => {
```
rmi22186, line 430: nitpick - When I read this test, it only ensures the order of events, not batches.  Each batch is uploaded individually and then the events are checked. The order of the batches aren't ever checked.  You do that in the ensuing tests, not this one.
```suggestion
        it('should organize events in the order they are processed and maintain that order when uploading', (done) => {
```

test/src/tests-batchUploader.ts
```diff
@@ -84,16 +84,603 @@ describe('batch uploader', () => {
...
+                eventBatchingIntervalMillis: 1000,
+            };
+
+            window.mParticle._resetForTests(MPConfig);
+            window.mParticle.init(apiKey, window.mParticle.config);
+
+            window.mParticle.logEvent('Test Event 0');
+
+            // Force time to pass so that batch processing occurs
+            clock.tick(1000);
```
rmi22186, line 448: I would stick to either doing `mParticle.upload` or `clock.tick` throughout all the tests to avoid confusion as to why you are doing one and not the other.  Do a full find/replace in this file.

test/src/tests-batchUploader.ts
```diff
@@ -84,16 +84,603 @@ describe('batch uploader', () => {
...
+                'application_state_transition'
+            );
+            expect(batch1.events[2].data.event_name).to.equal('Test Event 0');
+
+            // Log a second batch of events
+            window.mParticle.logEvent('Test Event 1');
+            window.mParticle.logEvent('Test Event 2');
+            window.mParticle.logEvent('Test Event 3');
+
+            // Force an upload of events
```
rmi22186, line 471: do a find/replace in this file to change instances of this to what i suggested to be updated elsewhere in the doc: `Manually initiate the upload process - turn event into batches and upload the batch`.

rmi22186 (changes requested):
a couple more small nits

test/src/tests-batchUploader.ts
```diff
@@ -85,16 +84,597 @@ describe('batch uploader', () => {
...
+                        .batchesQueuedForProcessing;
+
+                expect(batchQueue.length).to.equal(3);
+
+                expect(batchQueue[0].events[0].event_type).to.equal(
+                    'session_start'
+                );
+                expect(batchQueue[0].events[1].event_type).to.equal(
+                    'application_state_transition'
+                );
```
rmi22186, line 567: This should include one more event:
```
expect(batchQueue[0].events[0].data.event_name).to.equal(
    'Test Event 0'
);
```

test/src/tests-batchUploader.ts
```diff
@@ -85,16 +84,597 @@ describe('batch uploader', () => {
...
+                overwriteRoutes: false,
+            });
+
+            window.mParticle.config.flags = {
+                eventsV3: '100',
+                eventBatchingIntervalMillis: 1000,
+            };
+
+            // debugger;
+
```
rmi22186, line 610:
```suggestion
```

src/batchUploader.ts
```diff
@@ -169,47 +207,59 @@ export class BatchUploader {
...
-        if (remainingUploads && remainingUploads.length) {
-            this.pendingUploads.unshift(...remainingUploads);
+
+        // Batches that do not successfully upload are added back to the process queue
+        // in the order they were created so that we can attempt re-transmission in
+        // the same sequence. This is to prevent any potential data corruption.
+        if (!isEmpty(batchesThatDidNotUpload)) {
+            // FIXME: This unshift will reverse the order of batches on a retry
+            // this.batchesQueuedForProcessing.unshift(...batchesThatDidNotUpload);
+            this.batchesQueuedForProcessing.push(...batchesThatDidNotUpload);
```
rmi22186, line 246: I suggest you revert this to unshift to finish this PR, then explain the Sam the situation.  I think this single line coded change should be a PR on its own as a bugfix. This way this PR can then be entitled something like : `refactor: Revise variable names for clarity, add tests for coverage`, and the next one is fixing this bug once you discussed w Sam.  This unblocks you from a release this week.

rmi22186 (changes requested):
minor nits

rmi22186 (approved):

PR #395, opened 2023-03-01
feat: Support sideloaded kits

rmi22186:
## Instructions

 ## Summary
This PR introduces an API change in order to support sideloaded kits.  This is supported on both NPM and CDN.  When the customer imports a custom kit and uses the following CDN:

```
mParticle.config.sideloadedKits = [SideloadedKit];
```
The sideloaded kit is processed and added to activeForwarders array so that all mParticle methods (ie. logPageView, logEvent, login, etc) will map to it. 

Note that as this is the first milestone, there is no filtering logic, so I added defaults to [each forwarder setting here.](https://github.com/mParticle/mparticle-web-sdk/pull/395/files#diff-5f6c74306b4617bc6153000c0fc1fec37efa885c93f4cd1806f867b56bda5984R561-R585).

 ## Testing Plan
* Unit tests written
* [Simple integration written](https://github.com/mparticle-integrations/mparticle-javascript-integration-example/tree/feat/side-loaded-kit) and new API tested incorporating this via both CDN and NPM.  Loads successfully with proper callbacks firing
image

 - Closes https://go.mparticle.com/work/ SQDSDKS-5113

### DIFF
--- a/src/forwarders.js
+++ b/src/forwarders.js
@@ -59,7 +59,6 @@ export default function Forwarders(mpInstance, kitBlocker) {
                         user ? user.getAllUserAttributes() : {},
                         forwarder.userAttributeFilters
                     );
-
                     if (!forwarder.initialized) {
                         forwarder.logger = mpInstance.Logger;
                         forwarder.init(
@@ -552,44 +551,60 @@ export default function Forwarders(mpInstance, kitBlocker) {
                     config.isSandbox ===
                         mpInstance._Store.SDKConfig.isDevelopmentMode
                 ) {
-                    newForwarder = new forwarders[name].constructor();
-
-                    newForwarder.id = config.moduleId;
-                    newForwarder.isSandbox = config.isDebug || config.isSandbox;
-                    newForwarder.hasSandbox = config.hasDebugString === 'true';
-                    newForwarder.isVisible = config.isVisible;
-                    newForwarder.settings = config.settings;
-
-                    newForwarder.eventNameFilters = config.eventNameFilters;
-                    newForwarder.eventTypeFilters = config.eventTypeFilters;
-                    newForwarder.attributeFilters = config.attributeFilters;
-
-                    newForwarder.screenNameFilters = config.screenNameFilters;
-                    newForwarder.screenNameFilters = config.screenNameFilters;
-                    newForwarder.screenAttributeFilters =
-                        config.screenAttributeFilters;
-
-                    newForwarder.userIdentityFilters =
-                        config.userIdentityFilters;
-                    newForwarder.userAttributeFilters =
-                        config.userAttributeFilters;
-
-                    newForwarder.filteringEventAttributeValue =
-                        config.filteringEventAttributeValue;
-                    newForwarder.filteringUserAttributeValue =
-                        config.filteringUserAttributeValue;
-                    newForwarder.eventSubscriptionId =
-                        config.eventSubscriptionId;
-                    newForwarder.filteringConsentRuleValues =
-                        config.filteringConsentRuleValues;
-                    newForwarder.excludeAnonymousUser =
-                        config.excludeAnonymousUser;
+                    newForwarder = this.returnConfiguredKit(
+                        forwarders[name],
+                        config
+                    );
 
                     mpInstance._Store.configuredForwarders.push(newForwarder);
                     break;
                 }
             }
         }
+    };
+
+    // kits can be included via mParticle UI, or vi
+    this.configureSideloadedKit = function(sideloadedKit) {
+        mpInstance._Store.configuredForwarders.push(
+            this.returnConfiguredKit(sideloadedKit)
+        );
+    };
+
+    this.returnConfiguredKit = function(forwarder, config = {}) {
+        const newForwarder = new forwarder.constructor();
+        newForwarder.id = config.moduleId;
+
+        // TODO: isSandbox, hasSandbox is never used in any kit or in core SDK.
+        // isVisibleInvestigate is only used in 1 place. It is always true if
+        // it is sent to JS. Investigate further to determine if these can be removed.
+        // https://go.mparticle.com/work/SQDSDKS-5156
+        newForwarder.isSandbox = config.isDebug || config.isSandbox;
+        newForwarder.hasSandbox = config.hasDebugString === 'true';
+        newForwarder.isVisible = config.isVisible || true;
+        newForwarder.settings = config.settings || {};
+
+        newForwarder.eventNameFilters = config.eventNameFilters || [];
+        newForwarder.eventTypeFilters = config.eventTypeFilters || [];
+        newForwarder.attributeFilters = config.attributeFilters || [];
+
+        newForwarder.screenNameFilters = config.screenNameFilters || [];
+        newForwarder.screenAttributeFilters =
+            config.screenAttributeFilters || [];
+
+        newForwarder.userIdentityFilters = config.userIdentityFilters || [];
+        newForwarder.userAttributeFilters = config.userAttributeFilters || [];
+
+        newForwarder.filteringEventAttributeValue =
+            config.filteringEventAttributeValue || {};
+        newForwarder.filteringUserAttributeValue =
+            config.filteringUserAttributeValue || {};
+        newForwarder.eventSubscriptionId = config.eventSubscriptionId || null;
+        newForwarder.filteringConsentRuleValues =
+            config.filteringConsentRuleValues || {};
+        newForwarder.excludeAnonymousUser =
+            config.excludeAnonymousUser || false;
+
+        return newForwarder;
     };
 
     this.configurePixel = function(settings) {
@@ -610,6 +625,7 @@ export default function Forwarders(mpInstance, kitBlocker) {
             );
         } else {
             try {
+                // Process MP configured Kits
                 if (
                     Array.isArray(config.kitConfigs) &&
                     config.kitConfigs.length
@@ -617,6 +633,30 @@ export default function Forwarders(mpInstance, kitBlocker) {
                     config.kitConfigs.forEach(function(kitConfig) {
                         self.configureForwarder(kitConfig);
                     });
+                }
+
+                // Process sideloaded kits
+                // TODO: Sideloading kits currently requires the use of a register method
+                // which requires an object on which to be registered.
+                // In the future, when all kits are moved to the config rather than
+                // there being a separate process for MP configured kits and
+                // sideloaded kits, this will need to be refactored.
+                if (Array.isArray(config.sideloadedKits)) {
+                    var registeredSideloadedKits = { kits: {} };
+
+                    // First register each kit's constructor onto registeredSideloadedKits,
+                    // which is typed { kits: Dictionary<constructor> }.
+                    // The constructors are keyed by the name of the kit.
+                    config.sideloadedKits.forEach(function(sideloadedKit) {
+                        sideloadedKit.register(registeredSideloadedKits);
+                    });
+
+                    // Then configure each kit
+                    for (var registeredKit in registeredSideloadedKits.kits) {
+                        var kitConstructor =
+                            registeredSideloadedKits.kits[registeredKit];
+                        self.configureSideloadedKit(kitConstructor);
+                    }
                 }
 
                 if (
--- a/src/forwarders.js
+++ b/src/forwarders.js
@@ -563,7 +563,7 @@ export default function Forwarders(mpInstance, kitBlocker) {
         }
     };
 
-    // kits can be included via mParticle UI, or vi
+    // kits can be included via mParticle UI, or via sideloaded kit config API
     this.configureSideloadedKit = function(kitConstructor) {
         mpInstance._Store.configuredForwarders.push(
             this.returnConfiguredKit(kitConstructor)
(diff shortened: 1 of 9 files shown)


src/forwarders.js
```diff
@@ -572,6 +543,50 @@ export default function Forwarders(mpInstance, kitBlocker) {
         }
     };
 
+    // kits can be included via mParticle UI, or vi
```
alexs-mparticle, line 546: Vi? like VIM? 😆

src/forwarders.js
```diff
@@ -572,6 +543,50 @@ export default function Forwarders(mpInstance, kitBlocker) {
...
+    this.configureSideloadedKit = function(sideloadedKit) {
+        mpInstance._Store.configuredForwarders.push(
+            this.returnConfiguredKit(sideloadedKit)
+        );
+    };
+
+    this.returnConfiguredKit = function(forwarder, config = {}) {
```
alexs-mparticle, line 553: I've been thinking about this for a while, but I wonder if we should create a Forwarder class or factory that takes a config and constructor, and returns an instance of a forwarder.

src/forwarders.js
```diff
@@ -599,6 +615,30 @@ export default function Forwarders(mpInstance, kitBlocker) {
                     });
                 }
 
+                // Process sideloaded kits
+                // TODO: Sideloading kits currently requires the use of a register method
```
alexs-mparticle, line 619: You should create a sustained engineering ticket so we can follow up on this.

test/src/tests-forwarders.js
```diff
@@ -2,12 +2,15 @@ import Utils from './utils';
...
 import { apiKey, MPConfig, testMPID, MessageType } from './config';
+import { expect } from 'chai';
+
 
 var getEvent = Utils.getEvent,
     getForwarderEvent = Utils.getForwarderEvent,
     setLocalStorage = Utils.setLocalStorage,
     forwarderDefaultConfiguration = Utils.forwarderDefaultConfiguration,
     MockForwarder = Utils.MockForwarder,
+    SideloadedKit = Utils.sideloadedKit,
```
alexs-mparticle, line 13: Nitpick: I would call this `Utils.mockSideLoadedKit`

src/forwarders.js
rmi22186: will finish my thought

src/forwarders.js
rmi22186: will make a stand in sustained eng ticket, while ben fleshes out actual ticket since there are a lot of web specific intricacies

test/src/tests-forwarders.js
```diff
@@ -2513,4 +2527,178 @@ describe('forwarders', function() {
...
+
+                mParticle.logEvent('foo', mParticle.EventType.Navigation);
+                const sideloadedKit1Event = window.SideloadedKit11.instance.receivedEvent;
+                const sideloadedKit2Event = window.SideloadedKit22.instance.receivedEvent;
+
+                sideloadedKit1Event.should.have.property('EventName', 'foo');
+                sideloadedKit2Event.should.have.property('EventName', 'foo');
+            });
+
+            it('should invoke side loaded identify call', function() {
```
alexs-mparticle, line 2607:
```suggestion
            it('should invoke sideloaded identify call', function() {
```

test/src/tests-forwarders.js
```diff
@@ -2513,4 +2527,178 @@ describe('forwarders', function() {
...
+
+                mParticle.init(apiKey, window.mParticle.config);
+                window.SideloadedKit11.instance.should.have.property('setUserIdentityCalled', true);
+                window.SideloadedKit22.instance.should.have.property('setUserIdentityCalled', true);
+
+                window.SideloadedKit11.instance.should.have.property('onUserIdentifiedCalled', true);
+                window.SideloadedKit22.instance.should.have.property('onUserIdentifiedCalled', true);
+            });
+
+            it('should invoke side loaded set/removeUserAttribute call', function() {
```
alexs-mparticle, line 2628:
```suggestion
            it('should invoke sideloaded set/removeUserAttribute call', function() {
```

test/src/tests-forwarders.js
```diff
@@ -2513,4 +2527,178 @@ describe('forwarders', function() {
...
+                mParticle.Identity.getCurrentUser().removeUserAttribute('gender');
+
+                window.SideloadedKit11.instance.should.have.property('setUserAttributeCalled', true);
+                window.SideloadedKit22.instance.should.have.property('setUserAttributeCalled', true);
+
+                window.SideloadedKit11.instance.should.have.property('removeUserAttributeCalled', true);
+                window.SideloadedKit22.instance.should.have.property('removeUserAttributeCalled', true);
+            });
+
+            it('should invoke side loaded logout call', function() {
```
alexs-mparticle, line 2646:
```suggestion
            it('should invoke sideloaded logout call', function() {
```

test/src/tests-forwarders.js
```diff
@@ -2513,4 +2527,178 @@ describe('forwarders', function() {
...
+                window.mParticle.config.sideloadedKits = sideloadedKits;
+
+                mParticle.init(apiKey, window.mParticle.config);
+                mParticle.Identity.logout();
+
+                window.SideloadedKit11.instance.should.have.property('onLogoutCompleteCalled', true);
+                window.SideloadedKit22.instance.should.have.property('onLogoutCompleteCalled', true);
+            });
+
+            it('should invoke side loaded login call', function() {
```
alexs-mparticle, line 2660:
```suggestion
            it('should invoke sideloaded login call', function() {
```

test/src/tests-forwarders.js
```diff
@@ -2513,4 +2527,178 @@ describe('forwarders', function() {
...
+                window.mParticle.config.sideloadedKits = sideloadedKits;
+
+                mParticle.init(apiKey, window.mParticle.config);
+                mParticle.Identity.login({userIdentities: {customerid: 'abc'}});
+
+                window.SideloadedKit11.instance.should.have.property('onLoginCompleteCalled', true);
+                window.SideloadedKit22.instance.should.have.property('onLoginCompleteCalled', true);
+            });
+
+            it('should invoke side loaded modify call', function() {
```
alexs-mparticle, line 2674:
```suggestion
            it('should invoke sideloaded modify call', function() {
```

test/src/tests-forwarders.js
```diff
@@ -2513,4 +2527,178 @@ describe('forwarders', function() {
...
+                window.mParticle.config.sideloadedKits = sideloadedKits;
+
+                mParticle.init(apiKey, window.mParticle.config);
+                mParticle.Identity.modify({userIdentities: {customerid: 'abc'}});
+
+                window.SideloadedKit11.instance.should.have.property('onModifyCompleteCalled', true);
+                window.SideloadedKit22.instance.should.have.property('onModifyCompleteCalled', true);
+            });
+
+            it('should invoke side loaded modify call', function() {
```
alexs-mparticle, line 2688:
```suggestion
            it('should invoke sideloaded modify call', function() {
```

alexs-mparticle (changes requested):

src/forwarders.js
```diff
@@ -599,6 +615,30 @@ export default function Forwarders(mpInstance, kitBlocker) {
...
+                // which requires an object on which to be registered.
+                // In the future, when all kits are moved to the config rather than
+                // there being a separate process for MP configured kits and
+                // sideloaded kits, this will need to be refactored.
+                if (Array.isArray(config.sideloadedKits)) {
+                    const registeredSideloadedKits = { kits: {} };
```
rmi22186, line 625: something like 
```suggestion
                    const emptySideloadedKits = { kits: {} };
                    ...
                    const reisteredSideloadei = {...emptysideloadedkits}

```

src/forwarders.js
```diff
@@ -572,6 +543,50 @@ export default function Forwarders(mpInstance, kitBlocker) {
         }
     };
 
+    // kits can be included via mParticle UI, or vi
+    this.configureSideloadedKit = function(sideloadedKit) {
+        mpInstance._Store.configuredForwarders.push(
+            this.returnConfiguredKit(sideloadedKit)
+        );
+    };
```
rmi22186, line 551:
```suggestion
    this.configureSideloadedKit = function(kitConstructor) {
        mpInstance._Store.configuredForwarders.push(
            this.returnConfiguredKit(kitConstructor)
        );
    };
```

src/forwarders.js
```diff
@@ -599,6 +615,30 @@ export default function Forwarders(mpInstance, kitBlocker) {
...
+                        sideloadedKit.register(registeredSideloadedKits);
+                    });
+
+                    // Then configure each kit
+                    for (const registeredKit in registeredSideloadedKits.kits) {
+                        const kitConstructor =
+                            registeredSideloadedKits.kits[registeredKit];
+                        self.configureSideloadedKit(kitConstructor);
+                    }
+                }
```
alexs-mparticle, line 640:
```suggestion
                    for (const registeredKitKey in registeredSideloadedKits.kits) {
                        const kitConstructor =
                            registeredSideloadedKits.kits[registeredKitKey];
                        self.configureSideloadedKit(kitConstructor);
                    }
                }
```

src/forwarders.js
```diff
@@ -590,6 +605,7 @@ export default function Forwarders(mpInstance, kitBlocker) {
             );
         } else {
             try {
+                // Process MP configured Kits
```
alexs-mparticle, line 608: Consider splitting this function up into separate functions, such as configure, register and configure pixel, or at the very least break up the flow so that each "unit" is in its own try/catch block with unique error cases. I think this would make it easier to debug and possibly refactor later.

test/src/tests-forwarders.js
```diff
@@ -2513,4 +2527,178 @@ describe('forwarders', function() {
...
+
+                mParticle.init(apiKey, window.mParticle.config);
+                const activeForwarders = mParticle.getInstance()._getActiveForwarders();
+                expect(activeForwarders.length, 'active forwarders length').to.equal(sideloadedKits.length);
+                expect(activeForwarders[0].name, '1st active forwarder ').to.deep.equal(sideloadedKit1.name);
+                expect(activeForwarders[1].name, '2nd active forwarder').to.deep.equal(sideloadedKit2.name);
+            });
+
+            it('should add sideloaded kits along with configured forwarders from server to the active forwarders', function() {
+                const sideloadedKit1 = new SideloadedKit('SideloadedKit1', 1);
```
rmi22186, line 2558: add comment to say it's simulating a real MP kit

test/src/tests-forwarders.js
```diff
@@ -2513,4 +2527,178 @@ describe('forwarders', function() {
...
+            it('should send event to sideloaded kits', function() {
+                const sideloadedKit1 = new SideloadedKit('SideloadedKit1', 1);
+                const sideloadedKit2 = new SideloadedKit('SideloadedKit2', 2);
+                const sideloadedKits = [sideloadedKit1, sideloadedKit2];
+                
+                window.mParticle.config.sideloadedKits = sideloadedKits;
+                mParticle.init(apiKey, window.mParticle.config);
+
+                mParticle.logEvent('foo', mParticle.EventType.Navigation);
+                const sideloadedKit1Event = window.SideloadedKit11.instance.receivedEvent;
```
rmi22186, line 2600: Fix this 11, 22

src/forwarders.js
rmi22186: https://go.mparticle.com/work/SQDSDKS-5161

src/forwarders.js
rmi22186: https://mparticle-eng.atlassian.net/browse/SQDSDKS-5162

src/forwarders.js
rmi22186: good call

src/forwarders.js
```diff
@@ -589,33 +604,78 @@ export default function Forwarders(mpInstance, kitBlocker) {
...
-                        self.configurePixel(pixelConfig);
-                    });
+            self.initForwarders(
+                mpInstance._Store.SDKConfig.identifyRequest.userIdentities,
+                forwardingStatsCallback
+            );
+        }
+    };
+
+    this.processMpConfiguredKits = function(config) {
```
alexs-mparticle, line 618: I think `processImportedKits` or `processConfiguredKits` sounds more appropriate to me, as they are imported from the config, mpConfigured kits sounds vague.

If the intention is to denote that they came from the mparticle config, then perhaps `processKitsFromMPConfig` might be more appropriate.

src/forwarders.js
```diff
@@ -589,33 +604,78 @@ export default function Forwarders(mpInstance, kitBlocker) {
...
-            try {
-                if (
-                    Array.isArray(config.kitConfigs) &&
-                    config.kitConfigs.length
-                ) {
-                    config.kitConfigs.forEach(function(kitConfig) {
-                        self.configureForwarder(kitConfig);
-                    });
-                }
+            this.processMpConfiguredKits(config);
```
alexs-mparticle, line 607: I think we need a short comment to explain this process, ideally to explain briefly the flow so that we can remember next time we refactor. It should either go here, or maybe in the function that calls `processForwarders`. Your choice.

alexs-mparticle (changes requested):

src/forwarders.js
rmi22186: hm, so there is the rare use case where a company is npm installing the kits, but is overriding the config settings with their own.  so really this is for any kit configured from our UI.  i'll do `processUIConfiguredKits`

rmi22186: or I'll do `processUIEnabledKits` actually

alexs-mparticle (approved):

einsteinx2 (approved):
Went over this with @rmi22186 over a Zoom call and agree on the approach. It's ultimately very similar to iOS and I think the concession of not using instances makes sense in the Web SDK case and is fine for this MVP.

alexs-mparticle (approved):

PR #396, opened 2023-03-01
refactor: Simplify Vault for Offline Storage

alexs-mparticle:
## Instructions

 ## Summary
 - Simplify Vault Implementation to be used as a simple local storage wrapper with a simplified API.

 ## Testing Plan

 - Closes https://go.mparticle.com/work/SQDSDKS-5146

### DIFF
--- a/src/vault.ts
+++ b/src/vault.ts
@@ -6,7 +6,7 @@ interface IVaultOptions {
     offlineStorageEnabled?: boolean;
 }
 
-export default class Vault<StorableItem extends Dictionary> {
+export default class Vault<StorableItem> {
     public contents: StorableItem;
     private readonly _storageKey: string;
     private logger?: Logger;
@@ -59,7 +59,7 @@ export default class Vault<StorableItem extends Dictionary> {
      */
     public purge(): void {
         this.contents = null;
-        this.removeItem();
+        this.removeFromLocalStorage();
     }
 
     private saveToLocalStorage(items: StorableItem): void {
@@ -77,12 +77,12 @@ export default class Vault<StorableItem extends Dictionary> {
     private getFromLocalStorage(): StorableItem | null {
         // TODO: Handle cases where Local Storage is unavailable
         // https://go.mparticle.com/work/SQDSDKS-5022
-        const itemString = window.localStorage.getItem(this._storageKey);
+        const item = window.localStorage.getItem(this._storageKey);
 
-        return itemString ? JSON.parse(itemString) : null;
+        return item ? JSON.parse(item) : null;
     }
 
-    private removeItem(): void {
+    private removeFromLocalStorage(): void {
         window.localStorage.removeItem(this._storageKey);
     }
 }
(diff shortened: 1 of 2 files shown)


src/vault.ts
```diff
@@ -121,21 +74,15 @@ export default class Vault<StorableItem extends Dictionary> {
...
 
-        return itemString ? JSON.parse(itemString) : {};
+        return itemString ? JSON.parse(itemString) : null;
     }
 
-    private removeItems(): void {
-        if (!this.offlineStorageEnabled) {
-            return;
-        }
+    private removeItem(): void {
```
alexs-mparticle, line 85: Should be `removeFromLocalStorage`

src/vault.ts
```diff
@@ -7,28 +7,19 @@ interface IVaultOptions {
 }
 
 export default class Vault<StorableItem extends Dictionary> {
```
rmi22186, line 9: Does this still need to extend dictionary?  I removed it locally, ran the tests and the tests still pass.

src/vault.ts
```diff
@@ -121,21 +74,15 @@ export default class Vault<StorableItem extends Dictionary> {
...
-            return this.contents;
-        }
+    private getFromLocalStorage(): StorableItem | null {
         // TODO: Handle cases where Local Storage is unavailable
         // https://go.mparticle.com/work/SQDSDKS-5022
         const itemString = window.localStorage.getItem(this._storageKey);
 
-        return itemString ? JSON.parse(itemString) : {};
+        return itemString ? JSON.parse(itemString) : null;
     }
```
rmi22186, line 83: I think the following would be more TS-y. 

```
const item: string = window.localStorage.getItem(this._storageKey);

return item ? JSON.parse(item) : null;
```

test/src/tests-vault.ts
```diff
@@ -1,329 +1,168 @@
...
 
-    describe('#storeItems', () => {
-        it('should store an array of batches as a hashmap', () => {
-            const batch1 = {
-                foo: 'bar',
+        it('should store batches in the order they were added', () => {
+            const storageKey = 'test-batch-save-order';
+
+            const batch1: Partial<Batch> = {
+                mpid: 'bar',
```
rmi22186, line 52: Very minor nit.  There should be some sort of order to each mpid and source_request_id for each batch. I know we aren't necessarily referencing those items.  But I might do something like `source_request_id_1/2/3/4/5`, and `mpid1/2/3/4/5` rather than them being totally random. It creates a bit more order and is easier to read.

rmi22186 (changes requested):
Mostly minor nits.  Looks great overall.

test/src/tests-vault.ts
alexs-mparticle: I think my intention was to show that the order wasn't being alphabetized, which is an issue with JS objects.

rmi22186: Oh - let me rephrase - it has more to do with readability so we would know which mpid and which source request id was part of which batch if we ever needed to test those.  I know we're not testing them specifically in this test, but we should adhere to that practice to make it easier to grok that `mpid2` is with `batch2`, `mpid1` is `batch1`, rather than how it is now with mpid of `bar` with `batch1`, mpid `baz` with `batch2`.

rmi22186 (approved):

PR #435, opened 2023-03-14
ci: Manually build at each job instead of downloading the archive

rmi22186:
## Instructions

 ## Summary
I noticed that with the new CI steps merged from this [PR](https://go.mparticle.com/work/SQDSDKS-5163) that tests for commonjs and modulejs were failing in GHA, even though they were passing locally. 

https://github.com/mParticle/mparticle-web-sdk/actions/runs/4416514762/jobs/7741341885 

https://github.com/mParticle/mparticle-web-sdk/actions/runs/4416514762/jobs/7741341109 

Upon additional testing, it seems that there is an issue with downloading from the archive in each of these individual jobs. When we manually build the dist/ file, the tests pass, as shown in this build https://github.com/mParticle/mparticle-web-sdk/actions/runs/4417052866/jobs/7742205773

 ## Testing Plan

 - Closes https://go.mparticle.com/work/SQDSDKS-5205

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -27,9 +27,6 @@ jobs:
             - name: Lint with Prettier
               run: npm run prettier
 
-            - name: Build Files
-              run: npm run build
-
             - name: Archive Bundle
               uses: actions/upload-artifact@v3
               with:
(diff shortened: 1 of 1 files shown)


.github/workflows/pull-request.yml
```diff
@@ -162,10 +162,8 @@ jobs:
             - name: Run NPM CI
               run: npm ci
 
-            - name: Download Local Bundle
-              uses: actions/download-artifact@v3
-              with:
-                  name: bundle-local
+            - name: Build Files
```
alexs-mparticle, line 165: I think each test should build its own files, rather than running the entire build. Would make it easier to troubleshoot in the future if a single task is off.

alexs-mparticle (approved):

PR #459, opened 2023-03-21
refactor: Add interfaces for kits

rmi22186:
## Instructions

 ## Summary
- Kits go through a few different states in our SDK when initializing, and each one has a different interface. I named these 1. unregistered,  2. registered, and 3. configured.  This PR adds a few interfaces and updates a few others.

 ## Testing Plan

 - Closes https://go.mparticle.com/work/SQDSDKS-5160=

### DIFF
--- a/src/configAPIClient.ts
+++ b/src/configAPIClient.ts
@@ -22,8 +22,8 @@ export interface IKitConfigs {
     isDebugString: BooleanStringLowerCase;
     hasDebugString: BooleanStringLowerCase;
     settings: Dictionary;
-    screenNameFilters: number[]; // TODO: STILL NEED VERIFICATION FROM A REAL CONFIG
-    screenAttributeFilters: number[];  // TODO: STILL NEED VERIFICATION FROM A REAL CONFIG
+    screenNameFilters: number[];
+    screenAttributeFilters: number[];
     userIdentityFilters: number[];
     userAttributeFilters: number[];
     eventNameFilters: number[];
--- a/src/configAPIClient.ts
+++ b/src/configAPIClient.ts
@@ -22,15 +22,16 @@ export interface IKitConfigs {
     isDebugString: BooleanStringLowerCase;
     hasDebugString: BooleanStringLowerCase;
     settings: Dictionary;
-    screenNameFilters: number[];
-    screenAttributeFilters: number[];
+    screenNameFilters: number[]; // TODO: STILL NEED VERIFICATION FROM A REAL CONFIG
+    screenAttributeFilters: number[];  // TODO: STILL NEED VERIFICATION FROM A REAL CONFIG
     userIdentityFilters: number[];
     userAttributeFilters: number[];
     eventNameFilters: number[];
     eventTypeFilters: number[];
     attributeFilters: number[];
-    filteringEventAttributeValue: Dictionary;
+    filteringEventAttributeValue: IFilteringEventAttributeValue;
     filteringConsentRuleValues: Dictionary;
+    filteringUserAttributeValue: IFilteringUserAttributeValue;
     consentRegulationFilters: number[];
     consentRegulationPurposeFilters: number[];
     messageTypeFilters: number[];
@@ -38,6 +39,25 @@ export interface IKitConfigs {
     eventSubscriptionId: number;
     excludeAnonymousUser: boolean;
 }
+
+export interface IFilteringAttributeValue {
+    eventAttributeName: string;
+    eventAttributeValue: string;
+    includeOnMatch: boolean
+}
+
+export interface IFilteringEventAttributeValue extends IFilteringAttributeValue {}
+export interface IFilteringUserAttributeValue extends IFilteringAttributeValue {}
+export interface IFilteringConsentRulesValue {
+    includeOnMatch: boolean;
+    values: IConsentRuleValue[];
+}
+
+export interface IConsentRuleValue {
+    consentPurpose: string;
+    hasConsented: boolean
+}
+
 
 export interface IPixelConfig {
     name: string;
--- a/src/forwarders.interfaces.ts
+++ b/src/forwarders.interfaces.ts
@@ -1,0 +1,41 @@
+import { SDKEvent } from './sdkRuntimeModels';
+import { Dictionary } from './utils';
+import { IKitConfigs } from './configAPIClient';
+
+// The state of the kit when accessed via window.KitName via CDN
+// or imported as an NPM package
+export interface UnregisteredKit {
+    register(config): void;
+}
+
+// The state of the kit after being added to forwarderConstructors in the CDN 
+// or after registered to SDKConfig.kits via NPM
+export interface RegisteredKit {
+    constructor(): void;
+    name: string;
+    getId(): number;
+}
+
+// The state of the kit aftering being configured. This is what the kit looks like when acted on.
+export interface ConfiguredKit
+    extends Omit<IKitConfigs, 'isDebugString' | 'hasDebugString'> {
+    common: Dictionary;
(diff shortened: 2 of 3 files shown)


src/configAPIClient.ts
```diff
@@ -39,6 +40,25 @@ export interface IKitConfigs {
     excludeAnonymousUser: boolean;
 }
 
+export interface IFilteringAttributeValue {
```
alexs-mparticle, line 43: It looks like both interfaces are the exact same. Why do we need to differentiate? Can we just use `IFilteringAttributeValue` for both `filteringEventAttributeValue` and `filteringUserAttributeValue`?

src/configAPIClient.ts
```diff
@@ -39,6 +40,25 @@ export interface IKitConfigs {
...
+    eventAttributeName: string;
+    eventAttributeValue: string;
+    includeOnMatch: boolean
+}
+
+export interface IFilteringEventAttributeValue extends IFilteringAttributeValue {}
+export interface IFilteringUserAttributeValue extends IFilteringAttributeValue {}
+export interface IFilteringConsentRulesValue {
```
alexs-mparticle, line 51: I don't see where this is being used.

src/forwarders.interfaces.ts
```diff
@@ -0,0 +1,41 @@
...
+    name: string;
+    getId(): number;
+}
+
+// The state of the kit aftering being configured. This is what the kit looks like when acted on.
+export interface ConfiguredKit
+    extends Omit<IKitConfigs, 'isDebugString' | 'hasDebugString'> {
+    common: Dictionary;
+    init(settings, service, testMode, trackerId, userAttributes, userIdentities, appVersion, appName, customFlags, clientId): void;
+    onIdentifyComplete(user, filteredIdentityRequest): void;
```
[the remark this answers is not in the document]
rmi22186: I reviewed these and will update.  I also made this bug here which is tangential to this - https://mparticle-eng.atlassian.net/browse/SQDSDKS-5221

src/forwarders.interfaces.ts
```diff
@@ -0,0 +1,45 @@
...
+    name: string;
+    getId(): number;
+}
+
+// The state of the kit aftering being configured. This is what the kit looks like when acted on.
+export interface ConfiguredKit
+    extends Omit<IKitConfigs, 'isDebugString' | 'hasDebugString'> {
+    common: Dictionary;
+    init(settings, service, testMode, trackerId, userAttributes, userIdentities, appVersion, appName, customFlags, clientId): string;
+    onIdentifyComplete(user, filteredIdentityRequest): string | KitMappedMethodFailure;
```
alexs-mparticle, line 24: for these signatures, `user` should be `user: MPUser`. But what is `filteredIdentityRequest`?

Also, based on usage, shouldn't the return type be `boolean` not `string`?

Lastly, can you provide a link to a sample of how this is used, to make it easier for me to grok how this is implemented?

src/forwarders.interfaces.ts
```diff
@@ -0,0 +1,45 @@
...
+    onModifyComplete(user, filteredIdentityRequest): string | KitMappedMethodFailure;
+    onUserIdentified(user): string | KitMappedMethodFailure;
+    process(event: SDKEvent): string;
+    setOptOut(isOptingOut: boolean): string | KitMappedMethodFailure;
+    removeUserAttribute(key:string): string;
+    setUserAttribute(key: string, value:string): void;
+
+    // TODO: Convert type to enum during Identity migration
+    // https://go.mparticle.com/work/SQDSDKS-5218
+    setUserIdentity(id: string, type: number): void;
```
alexs-mparticle, line 36: Nitpick, instead of `string` and `number`, I would make a new type called 
`IdentityValue` and `IdentityType` until we can get the enum values working the way we want.

I find this is helpful so we can remember what the value should be, rather than any old string and number.

src/forwarders.interfaces.ts
```diff
@@ -0,0 +1,45 @@
...
+    init(settings, service, testMode, trackerId, userAttributes, userIdentities, appVersion, appName, customFlags, clientId): string;
+    onIdentifyComplete(user, filteredIdentityRequest): string | KitMappedMethodFailure;
+    onLoginComplete(user, filteredIdentityRequest): string | KitMappedMethodFailure;
+    onLogoutComplete(user, filteredIdentityRequest): string | KitMappedMethodFailure;
+    onModifyComplete(user, filteredIdentityRequest): string | KitMappedMethodFailure;
+    onUserIdentified(user): string | KitMappedMethodFailure;
+    process(event: SDKEvent): string;
+    setOptOut(isOptingOut: boolean): string | KitMappedMethodFailure;
+    removeUserAttribute(key:string): string;
+    setUserAttribute(key: string, value:string): void;
```
alexs-mparticle, line 32: Based on usage, this returns a result.

src/forwarders.interfaces.ts
rmi22186: I like it

src/forwarders.interfaces.ts
rmi22186: Weird. I thought I updated this when I changed removeUserAttribute.  Good catch.

src/forwarders.interfaces.ts
```diff
@@ -0,0 +1,61 @@
...
+    hasSandbox: boolean;
+}
+export interface KitMappedMethodFailure {
+    error: string
+}
+
+export type UserIdentityId = string;
+export type UserIdentityType = number;
+
+export type prepareForwardingStats = (forwarder: ConfiguredKit, event: SDKEvent) => void;
```
alexs-mparticle, line 61:
```suggestion
export type forwardingStatsCallback = (forwarder: ConfiguredKit, event: SDKEvent) => void;
```

src/forwarders.interfaces.ts
```diff
@@ -0,0 +1,61 @@
...
+    init(
+        settings: Dictionary<unknown>,
+        service: prepareForwardingStats,
+        testMode: boolean,
+        trackerId: string,
+        userAttributes: UserAttributes,
+        userIdentities: SDKUserIdentity,
+        appVersion: string,
+        appName:string,
+        customFlags:SDKEventCustomFlags,
```
alexs-mparticle, line 34:
```suggestion
        customFlags: SDKEventCustomFlags,
```

src/forwarders.interfaces.ts
```diff
@@ -0,0 +1,61 @@
...
+        clientId:string): string;
```
alexs-mparticle, line 35:
```suggestion
        clientId: string): string;
```

src/forwarders.interfaces.ts
```diff
@@ -0,0 +1,61 @@
...
+    onIdentifyComplete(user: MParticleUser, filteredIdentityRequest: IdentityApiData): string | KitMappedMethodFailure;
+    onLoginComplete(user: MParticleUser, filteredIdentityRequest: IdentityApiData): string | KitMappedMethodFailure;
+    onLogoutComplete(user: MParticleUser, filteredIdentityRequest: IdentityApiData): string | KitMappedMethodFailure;
+    onModifyComplete(user: MParticleUser, filteredIdentityRequest: IdentityApiData): string | KitMappedMethodFailure;
+    onUserIdentified(user: MParticleUser): string | KitMappedMethodFailure;
+    process(event: SDKEvent): string;
+    setOptOut(isOptingOut: boolean): string | KitMappedMethodFailure;
+    removeUserAttribute(key:string): string;
```
alexs-mparticle, line 43:
```suggestion
    removeUserAttribute(key: string): string;
```

alexs-mparticle (changes requested):

alexs-mparticle (approved):

PR #500, opened 2023-04-18
fix: Remove legacy migrations module

alexs-mparticle:
## Instructions

 ## Summary
 - Remove legacy migrations module
 - This is legacy code that was written to support and older version of our cookie and Events APIs. As all users have been migrated to v3 of our Events API, we no longer need to migrate.

 ## Testing Plan
 - Run automated tests to verify SDK continues to work
 - Test locally to make sure there are no errors.

 - Closes https://go.mparticle.com/work/SQDSDKS-2733

### DIFF
--- a/src/mockBatchCreator.ts
+++ b/src/mockBatchCreator.ts
@@ -63,7 +63,6 @@ export default class _BatchValidator {
                 context: null,
                 configurationLoaded: false,
                 identityCallInFlight: false,
-                migratingToIDSyncCookies: false,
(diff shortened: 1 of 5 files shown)


rmi22186:
Quick note about: `This is legacy code that was written to support and older version of our cookie and Events APIs. As all users have been migrated to v3 of our Events API, we no longer need to migrate`

v3 events API actually has nothing to do with this.  It is all related to the older version of the cookie.

rmi22186 (approved):

PR #501, opened 2023-04-18
refactor: Update Persistence Tests to TypeScript

alexs-mparticle:
## Instructions

 ## Summary
 - Migrate Persistence Tests to Typescript as part of a phased approach of migrating Persistence to TypeScript

 ## Testing Plan
 - N/A as this is only changing automated tests

 - Closes https://go.mparticle.com/work/SQDSDKS-5169

### DIFF
--- a/src/persistence.interfaces.ts
+++ b/src/persistence.interfaces.ts
@@ -98,7 +98,7 @@ export interface IPersistence {
         expires: string,
         domain: string,
         maxCookieSize: number
-    ): string | undefined;
+    ): string;
     findPrevCookiesBasedOnUI(identityApiData: IdentityApiData): void;
     encodePersistence(persistance: IPersistenceMinified): string;
     decodePersistence(persistance: IPersistenceMinified): string;
--- a/src/persistence.interfaces.ts
+++ b/src/persistence.interfaces.ts
@@ -101,7 +101,7 @@ export interface IPersistence {
     ): string;
     findPrevCookiesBasedOnUI(identityApiData: IdentityApiData): void;
     encodePersistence(persistence: IPersistenceMinified): string;
-    decodePersistence(persistanceString: string): string;
+    decodePersistence(persistenceString: string): string;
     replaceCommasWithPipes(string: string): string;
     replacePipesWithCommas(string: string): string;
     replaceApostrophesWithQuotes(string: string): string;
--- a/src/persistence.interfaces.ts
+++ b/src/persistence.interfaces.ts
@@ -100,7 +100,7 @@ export interface IPersistence {
         maxCookieSize: number
     ): string;
     findPrevCookiesBasedOnUI(identityApiData: IdentityApiData): void;
-    encodePersistence(persistance: IPersistenceMinified): string;
+    encodePersistence(persistence: IPersistenceMinified): string;
     decodePersistence(persistanceString: string): string;
     replaceCommasWithPipes(string: string): string;
     replacePipesWithCommas(string: string): string;
--- a/src/persistence.interfaces.ts
+++ b/src/persistence.interfaces.ts
@@ -101,7 +101,7 @@ export interface IPersistence {
     ): string;
     findPrevCookiesBasedOnUI(identityApiData: IdentityApiData): void;
     encodePersistence(persistance: IPersistenceMinified): string;
-    decodePersistence(persistance: IPersistenceMinified): string;
+    decodePersistence(persistanceString: string): string;
     replaceCommasWithPipes(string: string): string;
     replacePipesWithCommas(string: string): string;
     replaceApostrophesWithQuotes(string: string): string;
--- a/src/persistence.interfaces.ts
+++ b/src/persistence.interfaces.ts
@@ -98,7 +98,7 @@ export interface IPersistence {
         expires: string,
         domain: string,
         maxCookieSize: number
-    ): void;
+    ): string | undefined;
     findPrevCookiesBasedOnUI(identityApiData: IdentityApiData): void;
     encodePersistence(persistance: IPersistenceMinified): string;
     decodePersistence(persistance: IPersistenceMinified): string;
--- a/test/src/tests-persistence.ts
+++ b/test/src/tests-persistence.ts
@@ -1921,7 +1921,7 @@ describe('migrations and persistence-related', () => {
         done();
     });
 
-    it('should save to persistence a device id set viamParticle.config', done => {
+    it('should save to persistence a device id set via mParticle.config', done => {
         mParticle._resetForTests(MPConfig);
         mParticle.config.deviceId = 'foo-guid';
         mParticle.init(apiKey, mParticle.config);
--- a/test/src/tests-persistence.ts
+++ b/test/src/tests-persistence.ts
@@ -1921,7 +1921,7 @@ describe('migrations and persistence-related', () => {
         done();
     });
 
-    it('should save to persistance a device id set viamParticle.config', done => {
+    it('should save to persistence a device id set viamParticle.config', done => {
         mParticle._resetForTests(MPConfig);
         mParticle.config.deviceId = 'foo-guid';
         mParticle.init(apiKey, mParticle.config);
--- a/test/src/tests-persistence.ts
+++ b/test/src/tests-persistence.ts
@@ -727,7 +727,7 @@ describe('migrations and persistence-related', () => {
         mParticle._resetForTests(MPConfig);
         mParticle.config.maxCookieSize = 700;
 
-        const cookies: Partial<IPersistenceMinified> = {
+        const cookies: IPersistenceMinified = {
             gs: {
                 csm: ['mpid1', 'mpid2', 'mpid3'],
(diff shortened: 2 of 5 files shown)


src/persistence.interfaces.ts
```diff
@@ -98,7 +98,7 @@ export interface IPersistence {
         expires: string,
         domain: string,
         maxCookieSize: number
-    ): void;
+    ): string | undefined;
```
jared-dickman, line 101: i feel like `| undefined` may not be needed here and is implicit with the compiler

test/src/tests-persistence.ts
```diff
@@ -444,21 +473,21 @@ describe('migrations and persistence-related', function() {
         done();
     });
 
-    it('should add new MPID to cookies when returned MPID does not match anything in cookies, and have empty UI and UA', function(done) {
-        mParticle._resetForTests(MPConfig);
+    it('should add new MPID to cookies when returned MPID does not match anything in cookies, and have empty UI and UA', done => {
+        window.mParticle._resetForTests(MPConfig);
```
jared-dickman, line 477: could this go in a beforeEach or similar?

test/src/tests-persistence.ts
```diff
@@ -639,74 +681,75 @@ describe('migrations and persistence-related', function() {
...
 
-    it('should remove MPID as keys if the cookie size is beyond the setting', function(done) {
-        mParticle._resetForTests(MPConfig);
-        mParticle.config.maxCookieSize = 700;
+    it('should remove MPID as keys if the cookie size is beyond the setting', done => {
+        window.mParticle._resetForTests(MPConfig);
+        window.mParticle.config.maxCookieSize = 700;
 
-        var cookies = {
+        const cookies: Partial<IPersistenceMinified> = {
```
jared-dickman, line 742: **praise** types!

test/src/tests-persistence.ts
```diff
@@ -1695,49 +1747,62 @@ describe('migrations and persistence-related', function() {
...
+        window.mParticle
             .getInstance()
             ._Persistence.getFirstSeenTime('previous_set')
             .should.equal(100);
         done();
     });
 
-    it('should properly set setLastSeenTime()', function(done) {
+    it('should properly set setLastSeenTime()', done => {
+        debugger;
```
jared-dickman, line 1799: **issue** `debugger;`

test/src/tests-persistence.ts
```diff
@@ -1695,49 +1747,62 @@ describe('migrations and persistence-related', function() {
...
+
+        const mParticle = window.mParticle;
```
jared-dickman, line 1801: **thought** why do this in this test and not _all_ the others? (if its not too late) would create a lot less lines of changes

jared-dickman:
overall looks good to me! mostly just syntax changes in the tests

test/src/tests-persistence.ts
alexs-mparticle: Possibly. We have some tests that don't need to reset the SDK, but that might require restructuring this test file. I'd rather save this for a future refactor than introduce more scope creep.

test/src/tests-persistence.ts
alexs-mparticle: Good callout

test/src/tests-persistence.ts
```diff
@@ -963,44 +997,48 @@ describe('migrations and persistence-related', function() {
...
-        Should(cookiesResult['mpid2']).not.be.ok();
-        Should(cookiesResult['mpid3']).be.ok();
+        expect(cookiesResult['mpid1']).to.not.be.ok;
+        expect(cookiesResult['mpid2']).to.not.be.ok;
+        expect(cookiesResult['mpid3']).be.ok;
         cookiesResult.gs.csm[0].should.equal('mpid3');
         cookiesResult.should.have.property('mpid3');
-        mParticle.maxCookieSize = 3000;
+        // FIXME: Should this be in the config or not?
+        mParticle.config.maxCookieSize = 3000;
```
rmi22186, line 1031: The only thing I can thin of is this is done to revert the maxCookieSize back to the default size of 3000 since it is set to 400 on line 960.  Perhaps adding a comment here would have saved us this confusion, so I'd say add one.  Then again it gets set in the next few tests as well.  If tests pass if you remove this, then I'd say it's safe to remove.

test/src/tests-persistence.ts
```diff
@@ -1695,12 +1729,13 @@ describe('migrations and persistence-related', function() {
         });
 
         setCookie(workspaceCookieName, cookies);
-        mParticle.useCookieStorage = true;
+        // FIXME: Should this be in configs or global?
```
rmi22186, line 1732: I think perhaps it should be in a before clause for a separate section.  The persistence tests definitely need to be organized better with more `describe` blocks.  The lack of them right now makes things confusing as to when one portion of related tests ends and another begins - this is an issue overall with how tests are set up and not just persistence.  Something we should 100% do better in jest.

test/src/tests-persistence.ts
```diff
@@ -1877,57 +1897,64 @@ describe('migrations and persistence-related', function() {
             cu: 'current',
         });
         setCookie(workspaceCookieName, cookies, true);
-        mParticle.useCookieStorage = true;
+        // FIXME: config or global?
```
rmi22186, line 1900: ditto

rmi22186 (changes requested):
Overall this looks good. The `Should(foo)` --> expect(foo)` makes sense to me, but I wasn't sure why there were a couple places you updated a `(foo).should` --> `expect(foo)` (old line 1454/new line 1492 as an example).  Did this fail even if `Should` wasn't being used?

test/src/tests-persistence.ts
alexs-mparticle: The placement of this assignment at the end of the test is weird to me, but there's also an inconstancy in the SDK interface. Technically `maxCookieSize` is not a property of `MParticleSDK`, but in the original version of this diff, it is. My question is whether or not it _should_ be.

alexs-mparticle: The placement of this assignment at the end of the test is weird to me, but there's also an inconstancy in the SDK interface. Technically `maxCookieSize` is not a property of `MParticleSDK`, but in the original version of this diff, it is. My question is whether or not it _should_ be.

test/src/tests-persistence.ts
```diff
@@ -734,29 +765,32 @@ describe('migrations and persistence-related', function() {
                 },
             },
         };
-        var expires = new Date(
+        const expires = new Date(
             new Date().getTime() + 365 * 24 * 60 * 60 * 1000
-        ).toGMTString();
-        var cookiesWithExpiration = mParticle
+        ).toString();
```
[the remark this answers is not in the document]
alexs-mparticle: Decision over Zoom: `toUTCString()`

src/persistence.interfaces.ts
```diff
@@ -98,10 +98,10 @@ export interface IPersistence {
         expires: string,
         domain: string,
         maxCookieSize: number
-    ): void;
+    ): string;
     findPrevCookiesBasedOnUI(identityApiData: IdentityApiData): void;
     encodePersistence(persistance: IPersistenceMinified): string;
```
rmi22186, line 103:
```suggestion
    encodePersistence(persistence: IPersistenceMinified): string;
```

src/persistence.interfaces.ts
```diff
@@ -98,10 +98,10 @@ export interface IPersistence {
...
-    decodePersistence(persistance: IPersistenceMinified): string;
+    decodePersistence(persistanceString: string): string;
```
rmi22186, line 104:
```suggestion
    decodePersistence(persistenceString: string): string;
```

test/src/tests-persistence.ts
```diff
@@ -1877,57 +1896,64 @@ describe('migrations and persistence-related', function() {
...
+        expect(
             mParticle.getInstance()._Persistence.getFirstSeenTime('previous')
-        ).equal(null);
+        ).to.equal(null);
 
         done();
     });
 
-    it('should save to persistance a device id set with setDeviceId', function(done) {
+    it('should save to persistance a device id set with setDeviceId', done => {
```
rmi22186, line 1910:
```suggestion
    it('should save to persistence a device id set with setDeviceId', done => {
```

test/src/tests-persistence.ts
```diff
@@ -1877,57 +1896,64 @@ describe('migrations and persistence-related', function() {
...
+        mParticle
+            .getInstance()
+            ._Persistence.getLocalStorage()
+            .gs.das.should.equal('foo-guid');
 
         done();
     });
 
-    it('should save to persistance a device id set via mParticle.config', function(done) {
+    it('should save to persistance a device id set viamParticle.config', done => {
```
rmi22186, line 1924:
```suggestion
    it('should save to persistence a device id set viamParticle.config', done => {
```

rmi22186 (approved):
typos, otherwise LGTM

PR #516, opened 2023-04-25
refactor: Move util methods out of persistence

alexs-mparticle:
## Instructions

 ## Summary

 ## Testing Plan

 - Closes https://go.mparticle.com/work/SQDSDKS-5338


src/utils.ts
```diff
@@ -134,6 +140,16 @@ const parseStringOrNumber = (
...
+const replaceCommasWithPipes = (value: string) => value.replace(/,/g, '|');
+
+const replacePipesWithCommas = (value: string) => value.replace(/\|/g, ',');
+
+const replaceApostrophesWithQuotes = (value: string) =>
+    value.replace(/\'/g, '"');
+
+const replaceQuotesWithApostrophes = (value: string) =>
+    value.replace(/\"/g, "'");
+
```
rmi22186, line 152:
```suggestion
const replaceCommasWithPipes = (value: string): string => value.replace(/,/g, '|');

const replacePipesWithCommas = (value: string): string => value.replace(/\|/g, ',');

const replaceApostrophesWithQuotes = (value: string): string =>
    value.replace(/\'/g, '"');

const replaceQuotesWithApostrophes = (value: string): string =>
    value.replace(/\"/g, "'");

```

rmi22186 (approved):

PR #507, opened 2023-04-21
refactor: Remove non-batching event code path (v2 events endpoint)

rmi22186:
## Instructions
 1. PR target branch should be against `development`
 2. PR title name should follow this format: https://github.com/mParticle/mparticle-workflows/blob/main/.github/workflows/pr-title-check.yml
 3. PR branch prefix should follow this format: https://github.com/mParticle/mparticle-workflows/blob/main/.github/workflows/pr-branch-check-name.yml

 ## Summary
I had to rebase development into the approved PR from https://github.com/mParticle/mparticle-web-sdk/pull/494 due to the potential conflicts.  Fortunately there were no conflicts, but there were a few failing tests. The first commit has already been approved (its commit hash is just difference because of the rebase). The other changes I made can easily be reviewd in commit hash `43a3b20`:
* In order to reduce the flakiness of each test, and make each test module more silo'd, I decided to add `beforeEach/afterEach` clauses where I set and reset the fetchMock each time.  This reduces some of the need to go to test-main.js to figure out what is happening (although it is less DRY, I think it's an improvement).
* The test `should trigger beacon on page visibilitychange events` was flakey and so I updated it to use lastCall instead of getCalls()[0] to always identify the correct call.
* Replaced a string with `url.events` for DRY-ness.

 ## Testing Plan
 - [x] Was this tested locally? If not, explain why.

 ## Reference Issue (For mParticle employees only.  Ignore if you are an outside contributor)
 - Closes https://go.mparticle.com/work/SQDSDKS-5315

PR #518, opened 2023-04-26
refactor: Force tests onto v3 endpoint, set up test utils, fix test-core-sdk.js

rmi22186:
## Instructions

 ## Summary
This is the first PR in a larger effort to break up https://github.com/mParticle/mparticle-web-sdk/pull/507 into more bite sized pieces. That PR has been approved, but since PR 507 touches all test files, it constantly needed to be rebased.  To avoid this, rather than remove the V2 path, which all tests depend on, I will update all tests to be on the V3 path and then update a few test files in each PR.  This PR starts with refactoring a lot of the test-utils functions (identical to the changes [here](https://github.com/mParticle/mparticle-web-sdk/pull/507/files)).  In order to confirm these are all working, I am refactoring `tests-core-sdk.js` first.

Each subsequent PR will uncomment a few `test/src/` files, until all are uncommented.  Then a final PR will remove the V2 code path.

 ## Testing Plan

 - Closes https://go.mparticle.com/work/SQDSDKS-5341

### DIFF
--- a/test/src/tests-main.js
+++ b/test/src/tests-main.js
@@ -18,6 +18,10 @@ beforeEach(function() {
         kitConfigs: [],
         requestConfig: false,
         isDevelopmentMode: false,
+        flags: {
+            eventBatchingIntervalMillis: 0,
+            eventsV3: 100
+        }
     };
 
     mParticle._resetForTests(MPConfig);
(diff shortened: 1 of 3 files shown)


test/src/tests-main.js
```diff
@@ -29,31 +33,31 @@ afterEach(function() {
...
-import './tests-runtimeToBatchEventsDTO';
-import './tests-apiClient';
-import './tests-mparticle-instance-manager';
-import './tests-queue-public-methods';
-import './tests-validators';
-import './tests-utils';
-import './tests-store';
-import './tests-config-api-client';
-import './tests-vault';
+// import './tests-temp-session-bug-fix';
```
alexs-mparticle, line 36: Nitpick. I think I understand at this moment why you're commenting this block of code out, but we should write a comment explaining why, since Future Alex and Rob are probably exhausted and braindead. Especially Future Alex.

alexs-mparticle (approved):

alexs-mparticle:
LGTM. My only gripe is that I would like us to replace `var` with `const` but I agree that it is out of scope at this point.

test/src/tests-main.js
rmi22186: i'll add it in the next one.   i expect this to move along pretty quickly at which point there won't be anything commented anymore.

PR #519, opened 2023-04-26
refactor: Update tests for batchUploader, beaconUpload, session bugfix, kit-blocking, persistence

rmi22186:
## Instructions

 ## Summary
Update tests for batchUploader, beaconUpload, and session bugfix

 ## Testing Plan

 - Closes https://go.mparticle.com/work/SQDSDKS-5342


test/src/tests-batchUploader.ts
```diff
@@ -671,7 +672,7 @@ describe('batch uploader', () => {
             done();
         });
 
-        it('should purge events from Session Storage upon Batch Creation', (done) => {
+         it('should purge events from Session Storage upon Batch Creation', (done) => {
```
alexs-mparticle, line 675: For some reason, there's a stray space here.

```suggestion
        it('should purge events from Session Storage upon Batch Creation', (done) => {
```

test/src/tests-batchUploader.ts
```diff
@@ -862,7 +863,7 @@ describe('batch uploader', () => {
 
             // First upload is successful
             window.fetchMock.post(urls.eventsV3, 200, {
-                overwriteRoutes: false,
+                overwriteRoutes: true,
```
alexs-mparticle, line 866: Why is this being changed to true?

test/src/tests-temp-session-bug-fix.js
```diff
@@ -303,4 +303,4 @@ describe('session bug fix test', function() {
 
         done();
     });
-});
+});
```
alexs-mparticle, line 306: I think you should undo this change.

alexs-mparticle (changes requested):

test/src/tests-batchUploader.ts
rmi22186: Weird.  for some reason it was failing when on true so i switched it...but seems to be working again.  I'll revert

alexs-mparticle (approved):

PR #528, opened 2023-04-27
refactor: Remove non-batching event code path (v2 events endpoint)

rmi22186:
## Instructions

 ## Summary
Now that all tests have been refactored to use the V3 events endpoint instead of the V2 events endpoint, we can remove all references to the V2 events endpoint from the source code.  

`test/src/tests-temp-session-bug-fix.js` has been removed because V2 is being removed - There was [a bug](https://github.com/mParticle/mparticle-web-sdk/pull/360) that was fixed in September that introduced these duplicated tests in order to test the v3 endpoint - the tests they were duplicated from were still testing the V2 endpoing.  Now that ALL tests have been moved to V3, these are 100% duplicated tests and can be removed.

 ## Testing Plan

 - Closes https://go.mparticle.com/work/SQDSDKS-5361


test/integrations/requirejs/test-requirejs.js
```diff
@@ -2,6 +2,17 @@ describe('Require.JS Pages', function() {
     it('loads mParticle properly', function() {
         var server = new MockHttpServer();
         server.start();
+        window.fetchMock.post('https://jssdks.mparticle.com/v3/JS/test_key/events', 200);
```
alexs-mparticle, line 5: Shouldn't this be using the `urls.events` variable?

test/src/tests-temp-session-bug-fix.js
```diff
@@ -1,306 +0,0 @@
-import Utils from './utils';
```
alexs-mparticle, line 1: 🎉

test/integrations/requirejs/test-requirejs.js
rmi22186: Yes, but I cant link to an expanded resolved comment, so I'll paste a screenshot here:
image

alexs-mparticle: Can you include a link to the Sus Eng ticket as a todo?

alexs-mparticle (approved):

test/integrations/requirejs/test-requirejs.js
rmi22186: done!

alexs-mparticle (approved):

PR #531, opened 2023-04-28
refactor: Move generateHash methods into utility class

rmi22186:
## Instructions

 ## Summary
This is step 1 in enabling a user to add filters to sideloaded kits.  This refactors instances where we use the `generateHash` for different filters to a helper class.  The original code is refactored to use this helper class.

Note that ecommerce and screenview (page view on web) are all filtered using the regular event filtering logic, and so we do not have hashing methods that relate to commerce or screenviews as was originally stated in the tech spec.

 ## Testing Plan

 - Closes https://go.mparticle.com/work/SQDSDKS-5349


rmi22186:
Looks like I'm having an import issue in the tests.  Was running locally fine.  I'll report back when this is ready again.

src/kitFilterHelpers.ts
```diff
@@ -0,0 +1,56 @@
+import { generateHash } from "./utils";
```
alexs-mparticle, line 1: Should we consider moving `generateHash` out of utils and into this class?

src/kitFilterHelpers.ts
```diff
@@ -0,0 +1,56 @@
...
+// TODO: EventType/EventTypeEnum exists in differnet forms between @types/eventmodel/dataplanningnode.  determine the differences and consolidate if possible
+import { EventTypeEnum } from "./types.interfaces";
+import { UserIdentityType } from "./forwarders.interfaces";
+import { IdentityType } from "@mparticle/web-sdk";
+
+
+export default class KitFilterHelpers {
```
alexs-mparticle, line 8: Nitpick about the name of this class. Is it really a helper and is it limited to only kits? Unless this is the naming convention we're using across all SDKs, maybe we call it `HashUtility`?

src/kitFilterHelpers.ts
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    static hashGDPRPurpose(purpose: string): number {
+        const GDPRConsentHashPrefix = '1';
+
+        return generateHash(GDPRConsentHashPrefix + purpose)
+    }
+
+    static hashCCPAPurpose(){
+        const CCPAPurpose = 'data_sale_opt_out' as const;
```
alexs-mparticle, line 43: I think this already exists elsewhere and you should import it from there.

alexs-mparticle (changes requested):

src/kitFilterHelpers.ts
rmi22186: `getRampNumber` which is in Utils also uses `generateHash`, so it would be awkward to import a class into utils file.

src/kitFilterHelpers.ts
rmi22186: @einsteinx2 @markvdouw - We should discuss this since it is a cross platform naming topic.

src/types.interfaces.ts
```diff
@@ -0,0 +1,36 @@
...
+    Search,
+    Transaction,
+    UserContent,
+    UserPreference,
+    Social,
+    Other,
+    Media,
+}
+
+export enum IdentityType {
```
alexs-mparticle, line 14: To avoid duplication, does it make sense to remove this from types.js? If it's easier, we can make this a sustained eng ticket and work on it separately.

test/src/tests-utils.ts
```diff
@@ -330,4 +330,11 @@ describe('Utils', () => {
             expect(isEmpty(undefined)).to.equal(true);
         });
     });
+
+    describe('hashing', () => {
+        it('hashing utility class should have certain functions on it', () => {
```
alexs-mparticle, line 335: Is this something you for a different PR or did is this still a work in progress?

src/kitFilterHelpers.ts
```diff
@@ -0,0 +1,54 @@
...
+        return userIdentity;
+    }
+
+    static hashGDPRPurpose(purpose: string): number {
+        const GDPRConsentHashPrefix = '1';
+
+        return generateHash(GDPRConsentHashPrefix + purpose)
+    }
+
+    static hashCCPA(){
```
alexs-mparticle, line 42: Nitpick: For consistency, since we have `hashGDPRPurpose`, this should be `hashCCPAPurpose`

src/kitFilterHelpers.ts
```diff
@@ -0,0 +1,54 @@
...
+        const CCPAHashPrefix = '2';
+        return generateHash(CCPAHashPrefix + CCPAPurpose);
+    }
+
+    // Forwarding filters
+    static hashEventAttributeKeyForForwarding(eventName: string): number {
+        return generateHash(eventName);
+    }
+    static hashEventAttributeValueForForwarding(customAttributeName: string): number {
```
alexs-mparticle, line 51:
```suggestion
    }

    static hashEventAttributeValueForForwarding(customAttributeName: string): number {
```

src/kitFilterHelpers.ts
rmi22186: [I actually originally had this, but I removed](https://github.com/mParticle/mparticle-web-sdk/pull/531/commits/9908880408d1062b2421ceedac07feacd39d5653) `purpose` because ccpa only has one purpose, and we pass a purpose into the GDPR, but it seems inconsistent to not pass a purpose into CCPA.  I could go either way.  input from @einsteinx2 and @markvdouw might be valuable as well.

src/kitFilterHelpers.ts
```diff
@@ -0,0 +1,54 @@
...
+        return generateHash(GDPRConsentHashPrefix + purpose)
+    }
+
+    static hashCCPA(){
+        const CCPAHashPrefix = '2';
+        return generateHash(CCPAHashPrefix + CCPAPurpose);
+    }
+
+    // Forwarding filters
+    static hashEventAttributeKeyForForwarding(eventName: string): number {
```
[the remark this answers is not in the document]
rmi22186: They are different in that they are for a slightly different filtering purpose, it's for conditional forwarding.  I can clarify the `// Forwarding` comment above to make that more clear.

src/forwarders.js
```diff
@@ -110,12 +111,12 @@ export default function Forwarders(mpInstance, kitBlocker) {
...
                     if (userAttributes.hasOwnProperty(attrName)) {
-                        attrHash = mpInstance._Helpers
-                            .generateHash(attrName)
-                            .toString();
-                        valueHash = mpInstance._Helpers
-                            .generateHash(userAttributes[attrName])
-                            .toString();
+                        attrHash = FilterUtilities.hashUserAttributeKey(
+                            attrName
+                        ).toString();
```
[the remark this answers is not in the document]
alexs-mparticle: I wasn't suggesting modifying `generateHash` but rather the return value of `KitFilterHelpers`'s methods.

src/utils.ts
```diff
@@ -1,3 +1,4 @@
+import { EventTypeEnum } from "./types.interfaces";
```
alexs-mparticle, line 1: You can remove this now that you are no longer using this enum.

src/forwarders.js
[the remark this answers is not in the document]
rmi22186: Just to provide a bit of background (and for future devs, bc I think it's a bit confusing either way we do it bc the backend is inconsistent :(). For some reason, the "normal" filtering requires a comparison of numbers, and the "conditional filtering" requires a string comparison.  So the 2 options are:
1. In the KitFilterHelper's methods, have 2 of the ~10 methods return a string, while the 8 others return a number and comment as to why.  In this case the return values are inconsistent, (which is probably okay despite all being labeled `hashFoo`.  
2. Keep as is if we think it's important for all `hash` methods to return the same, and use the toString() where it is being used.

The more I think of it, I think option 1 (your recommendation) is correct.  But I wanted to share that this is a bit nuanced, and backend driven. This way, the helper is doing everything required to make a proper comparison, and we can comment in the code one place why it is this way.

image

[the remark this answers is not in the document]
alexs-mparticle: I appreciate the explanation and I think it's great to have this context, especially for posterity.

I made my comment initially since I think in some uses of the hash functions, you're converting to a string, so I just assumed that returning a string in those cases would simplify things.

I still think you and I line up on option 1, but option 2 at least grants us some level of "type safety" without converting this file to TypeScript since we know this usage is converting to string. At least with a comment, we can understand why we're doing this conversion.

src/helpers.js
```diff
@@ -311,8 +312,8 @@ export default function Helpers(mpInstance) {
         if (userIdentitiesObject && Object.keys(userIdentitiesObject).length) {
             for (var userIdentityName in userIdentitiesObject) {
                 if (userIdentitiesObject.hasOwnProperty(userIdentityName)) {
-                    var userIdentityType = Types.IdentityType.getIdentityType(
-                        userIdentityName
+                    var userIdentityType = FilterUtilities.hashUserIdentity(
```
alexs-mparticle, line 315:
```suggestion
                    const userIdentityType = FilterUtilities.hashUserIdentity(
```

alexs-mparticle (changes requested):

src/kitFilterHelpers.ts
einsteinx2: Let's go with `KitFilterHelper` singular

src/types.interfaces.ts
rmi22186: Yea I'd say this is out of scope.  It's also an issue because `IdentityType` is extended in a bizarre way where it's not just types, but methods on these types.  It needs to be completely refactored when we migrate the Identity module.  I added that to the description to discuss here: go.mparticle.com/work/SQDSDKS-2736

alexs-mparticle (approved):

src/consent.ts
```diff
@@ -144,25 +145,21 @@ export default function Consent(this: IConsent, mpInstance: MParticleWebSDK) {
...
                             gdprConsentState[purpose].Consented;
                     }
                 }
             }
             const CCPAConsentState = consentState.getCCPAConsentState();
             if (CCPAConsentState) {
-                purposeHash = mpInstance._Helpers
-                    .generateHash(CCPAHashString)
-                    .toString();
+                purposeHash = KitFilterHelper.hashConsentPurposeConditionalForwarding(CCPAHashPrefix, CCPAPurpose);
```
einsteinx2, line 162: I know we mentioned that this whole thing could also just be a constant, but I kind of like this as it matches the logic for GDPR and may be easier to understand than a "magic" hash value.

src/helpers.js
```diff
@@ -306,13 +307,13 @@ export default function Helpers(mpInstance) {
         userIdentitiesObject,
         filterList
     ) {
-        var filteredUserIdentities = {};
+        const filteredUserIdentities = {};
```
einsteinx2, line 310: yay for `const`!

src/forwarders.js
```diff
@@ -1,6 +1,7 @@
 import Types from './types';
 import filteredMparticleUser from './filteredMparticleUser';
 import { isEmpty } from './utils';
+import FilterUtilities from './kitFilterHelper';
```
einsteinx2, line 4: We should rename `FilterUtilities` in all places to `KitFilterHelper` to match the class name and avoid confusion.

einsteinx2 (changes requested):
Looks good, only change I request is to replace `FilterUtilities` with `KitFilterHelper` so it matches the class name.

src/consent.ts
rmi22186: Yea, so we are doing this so that there is not a need for separate `hashCCPA` and `hashGDPR` methods.  It's used internally so no issues here.  It's more a matter of how will the public API look - but that can be discussed in the future when the ticket is done.

src/forwarders.js
rmi22186: Good catch.

PR #545, opened 2023-05-01
test: Increase timeout for batch uploader tests

rmi22186:
## Instructions

 ## Summary
Due to the async nature of the batch uploader, we need a setTimeout for several of our batch uploader tests.  We should extend this to `500` to reduce the flakiness of this test.

 ## Testing Plan

 - Closes https://go.mparticle.com/work/SQDSDKS-5373

### DIFF
--- a/test/karma.config.js
+++ b/test/karma.config.js
@@ -36,6 +36,11 @@ if (DEBUG === 'true') {
 
 module.exports = function(config) {
     config.set({
+        client: {
+            mocha: {
+                timeout: 10000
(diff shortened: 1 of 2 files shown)


alexs-mparticle (changes requested):

alexs-mparticle (approved):

rmi22186:
@alexs-mparticle this actually isn't working from my local testing. Closing out and we'll need to figure something else out wrt this flakey test.

PR #550, opened 2023-05-02
test: Reorder tests to avoid flaky tests

rmi22186:
## Instructions

 ## Summary
Reordering these to the front seem to reduce flakiness.  Once we migrate to jest, the order shouldn't matter.  

 ## Testing Plan

 - Closes https://go.mparticle.com/work/SQDSDKS-5373

### DIFF
--- a/test/src/tests-main.js
+++ b/test/src/tests-main.js
@@ -31,8 +31,8 @@ afterEach(function() {
     window.fetchMock.restore();
 });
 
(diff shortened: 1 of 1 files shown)


alexs-mparticle (approved):

rmi22186:
@alexs-mparticle The failure here, and further testing on our internal repo showed that the test reordering in different way is still flakey. Closing this.

PR #559, opened 2023-05-08
refactor: Implement Next Gen Stand Alone Interfaces and Classes

alexs-mparticle:
## Instructions

 ## Summary
 - First pass implementation of stand alone interfaces and classes for Next Gen Web SDK Patterns.

 ## Testing Plan
 - N/A - This pull request only implements initial classes and interfaces for review. Functionality will not be integrated with the core sdk at this point.

 - Closes https://go.mparticle.com/work/SQDSDKS-5376

### DIFF
--- a/src/NextGen/core/baseMPMessage.ts
+++ b/src/NextGen/core/baseMPMessage.ts
@@ -1,0 +1,3 @@
+export class BaseMPMessage {
+    public name: string;
+}
--- a/src/NextGen/core/component.ts
+++ b/src/NextGen/core/component.ts
@@ -1,0 +1,3 @@
+export interface IMParticleComponent {
+    // TODO: Should we just define the component type in here?
+}
--- a/src/NextGen/dataRepository.ts
+++ b/src/NextGen/dataRepository.ts
@@ -4,7 +4,7 @@ import { BaseMPMessage } from './core/baseMPMessage';
 export class MParticleDataRepository {
     public internalState: BaseMPMessage[] = [];
 
-    public insertCommerceEvent(data: BaseMPMessage) {
+    public insertCommerceEvent(data: BaseMPMessage): void {
         this.internalState.push(data);
     }
 
--- a/src/NextGen/dataRepository.ts
+++ b/src/NextGen/dataRepository.ts
@@ -1,0 +1,26 @@
+import { Batch } from '@mparticle/event-models';
+import { BaseMPMessage } from './core/baseMPMessage';
+
+export class MParticleDataRepository {
+    public internalState: BaseMPMessage[] = [];
+
+    public insertCommerceEvent(data: BaseMPMessage) {
+        this.internalState.push(data);
+    }
+
+    public insertBreadcrumb(data: BaseMPMessage): void {
+        this.internalState.push(data);
+    }
+
+    public getEventsByType(): BaseMPMessage[] {
+        return [];
+    }
+
+    public getBatch(source_request_id: string): Batch {
+        return null;
+    }
+
+    public getBatches(): Batch[] {
+        return [];
+    }
+}
--- a/src/NextGen/identity/identity.ts
+++ b/src/NextGen/identity/identity.ts
@@ -1,0 +1,146 @@
+import { MPID, User } from '@mparticle/web-sdk';
+import { IMParticleComponent } from '../core/component';
+import { Context } from '@mparticle/event-models';
+
+export interface IdentityApiResult {
+    user: User;
+    previousUser: User;
+
+    getUser(): User;
+    getPreviousUser(): User;
+}
+
+export enum MParticleClientErrorCodes {
+    noHttpCoverage = -1,
+    activeIdentityRequest = -2,
+    activeSession = -3,
+    validationIssue = -4,
+    nativeIdentityRequest = -5,
+    loggingDisabledOrMissingAPIKey = -6,
+}
+
+export enum MParticleHttpErrorCodes {
+    badRequest = 400,
+    forbidden = 401,
+    tooManyRequests = 429,
+    serverError = 500,
+}
+
+export interface IdentityAPIServerErrorResult {
+    errorCode: MParticleClientErrorCodes;
+}
+
+export interface IdentityAPIHttpErrorResult {
+    errorCode: number;
+}
+
+export type IMPIdentityApiResponse =
+    | IdentityApiResult
+    | IdentityAPIHttpErrorResult
+    | IdentityAPIServerErrorResult;
+
+// TODO: See if we can use a utility class to make this cleaner
+interface IKnownIdentities {
+    device_application_stamp: string;
+    other?: string;
+    customerid?: string;
+    facebook?: string;
+    twitter?: string;
+    google?: string;
+    microsoft?: string;
+    yahoo?: string;
+    email?: string;
+    facebookcustomaudienceid?: string;
+    other2?: string;
+    other3?: string;
+    other4?: string;
+    other5?: string;
+    other6?: string;
+    other7?: string;
+    other8?: string;
+    other9?: string;
+    other10?: string;
+    phone_number_2?: string;
+    phone_number_3?: string;
+    mobile_number?: string;
+}
+
+export interface IdentityApiRequest {
+    client_sdk: {
+        platform: string;
+        sdk_vendor: string;
+        sdk_version: string;
+    };
+    context: Context;
+    environment: 'development' | 'production';
+    request_id: string;
+    request_timestamp_ms: number;
+    previous_mpid: MPID | null;
+    known_identities: IKnownIdentities;
+}
+
+export interface IMParticleIdentity extends IMParticleComponent {
+    login(request: IdentityApiRequest): Promise<IMPIdentityApiResponse>;
+    logout(request: IdentityApiRequest): Promise<IMPIdentityApiResponse>;
+    getUser(mpid: MPID): Promise<User>;
+    getUsers(): Promise<User[]>;
+}
+
+export interface InternalIdentity extends IMParticleIdentity {
+    login(request: IdentityApiRequest): Promise<IMPIdentityApiResponse>;
+    logout(request: IdentityApiRequest): Promise<IMPIdentityApiResponse>;
+    modify(request: IdentityApiRequest): Promise<IMPIdentityApiResponse>;
+    identify(request: IdentityApiRequest): Promise<IMPIdentityApiResponse>;
+}
+
+export interface IdentityListener {}
+
+export class IdentityImplementation implements InternalIdentity {
+    constructor() {}
+
+    // TODO: Temporary implementation to test signature
+    private generateFakeAPIResult = (): IMPIdentityApiResponse => ({
+        user: null,
+        previousUser: null,
+        getUser: () => {
+            return null;
+        },
+        getPreviousUser: () => {
+            return null;
+        },
+    });
+
+    public identify(
+        request: IdentityApiRequest
+    ): Promise<IMPIdentityApiResponse> {
+        return Promise.resolve(this.generateFakeAPIResult());
+    }
+
+    public modify(
+        request: IdentityApiRequest
+    ): Promise<IMPIdentityApiResponse> {
+        // FIXME: Forcing a rejected promise for demonstration purposes
+        //        This should be replaced with a real implementation eventually
+        return Promise.reject({
+            errorCode: MParticleClientErrorCodes.activeSession,
+        });
+    }
+
+    public logout(
+        request: IdentityApiRequest
+    ): Promise<IMPIdentityApiResponse> {
+        return Promise.resolve(this.generateFakeAPIResult());
+    }
+
+    public login(request: IdentityApiRequest): Promise<IMPIdentityApiResponse> {
+        return Promise.resolve(this.generateFakeAPIResult());
+    }
+
+    public getUser(mpid: string): Promise<User> {
+        return Promise.resolve(null);
+    }
+
+    public getUsers(): Promise<User[]> {
+        return Promise.resolve(null);
+    }
+}
--- a/test/jest/NextGen/dataRepository.spec.ts
+++ b/test/jest/NextGen/dataRepository.spec.ts
@@ -1,0 +1,54 @@
+import { BaseMPMessage } from '../../../src/NextGen/core/baseMPMessage';
+import { MParticleDataRepository } from '../../../src/NextGen/dataRepository';
+
+describe('Next Gen', () => {
+    describe('Data Repository', () => {
+        describe('#insertCommerceEvent', () => {
+            const dataRepo = new MParticleDataRepository();
+            const event: BaseMPMessage = {
+                name: 'Session Start Test',
+            };
+
+            expect(dataRepo.insertCommerceEvent).toBeDefined();
+
+            expect(dataRepo.internalState.length).toBe(0);
+
+            dataRepo.insertCommerceEvent(event);
+
+            expect(dataRepo.internalState.length).toBe(1);
+        });
+
+        describe('#insertBreadcrumb', () => {
+            it('should update its own internal state', () => {
+                const dataRepo = new MParticleDataRepository();
+                const event: BaseMPMessage = {
+                    name: 'Breadcrumb test',
+                };
+
+                expect(dataRepo.insertBreadcrumb).toBeDefined();
+
+                dataRepo.insertBreadcrumb(event);
+
+                expect(dataRepo.internalState.length).toBe(1);
+            });
+        });
+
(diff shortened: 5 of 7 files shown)


markvdouw (commented):
Nice

src/NextGen/identity/identityCallback.ts
```diff
@@ -0,0 +1,8 @@
+import { IdentityApiResult } from './identity';
+import { mParticleCallback } from '../mParticleCallback';
+
+// TODO: This already exists in @mparticle/web-sdk
+export class IdentityCallback extends mParticleCallback<
+    IdentityApiResult,
```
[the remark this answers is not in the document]
alexs-mparticle: IIRC, on Native, the Identity callback contains an `IdentityApiResult` which returns two users, current and previous. However, on Web, our identity callback returns `IdentityResult` which contains the current and previous users, as well as HTTP status codes and additional body meta data.

I believe I pointed this out in identity.ts, but it's something we should discuss as part of our streamlining conversation.

src/NextGen/dataRepository.ts
```diff
@@ -0,0 +1,14 @@
+// QUESTION: Should we adopt BaseMessage like Native?
```
rmi22186, line 1: Can you remind me the differences?

src/NextGen/dataRepository.ts
```diff
@@ -0,0 +1,14 @@
...
+import { BaseEvent } from '../sdkRuntimeModels';
+
+export class mParticleDataRepository {
+    // For demonstration purposes. This will likely go away as we spec this out
```
rmi22186, line 5: Does `This will likely go away` refer just to the next line, or the general structure of the DataRepository?

src/NextGen/identity/identity.ts
```diff
@@ -0,0 +1,31 @@
...
+import { IdentityCallback } from './identityCallback';
+
+// QUESTION: What is the diff between this and IdentityApiData?
+// TODO: sync this with android class
+export interface IdentityApiResult {}
+export interface IdentityApiRequest {}
+
+// QUESTION: Should we mirror the way Android handles HTTP codes, or do it the way we've been doing it?
+// https://docs.mparticle.com/developers/sdk/android/idsync/#idsync-status-codes
+// https://docs.mparticle.com/developers/sdk/web/idsync/#error-handling
```
rmi22186, line 12: We should do a sync with Native to see where iOS's lives.  The more we can have in common, the better.  One improvement we need to make with web error handling is that there are results that are not http codes, https://github.com/mParticle/mparticle-web-sdk/blob/master/src/constants.ts#L151-L157

There are just error codes, which should live somewhere else as well.  It's unclear how these errors are represented (if at all) on native.

src/NextGen/identity/identity.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export interface IMparticleIdentity extends IMParticleComponent {
+    // QUESTION: Should these be callbacks or promises that eventually return something?
```
rmi22186, line 15: I'd argue we should move as much to promises as possible.

src/NextGen/identity/identity.ts
```diff
@@ -0,0 +1,31 @@
...
+    login(callback: IdentityCallback): void;
+    // login(): IdentityApiResult;
+    logout(callback: IdentityCallback): void;
+    // logout(): IdentityApiResult;
+    getUser(mpid: MPID, callback: IdentityCallback): User;
+    getUsers(callback: IdentityCallback): User[];
```
rmi22186, line 21: Neither of these make async requests, so I'm curious why they have callbacks.  Probably a question for @markvdouw

src/NextGen/mParticleCallback.ts
```diff
@@ -0,0 +1,16 @@
+export abstract class mParticleCallback<Success, Error> {
+    public isSuccessful: boolean = false;
+    public isCompleted: boolean = false;
+
+    // QUESTION: Why do these methods have a result param that isn't used?
```
rmi22186, line 5: I agree - seems like there should be a `this.result = result` in order to allow access to what is on the result (assuming there is anything of value there).

test/jest/NextGen/dataRepository.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+                expect(dataRepo.updateSession).toBeDefined();
+
+                expect(dataRepo.internalState.length).toBe(0);
+
+                dataRepo.updateSession(event);
+
+                expect(dataRepo.internalState.length).toBe(1);
+            });
+        });
+        describe('#insertBreadcrumb', () => {
```
rmi22186, line 24:
```suggestion
        });

        describe('#insertBreadcrumb', () => {
```

src/NextGen/dataRepository.ts
alexs-mparticle: A BaseMessage seems to have an mpid, session id, timestamps, attributes and some event level meta data (like Message Type) and it's somewhat similar to our BaseEvent.

For more specifics: https://github.com/mParticle/mparticle-android-sdk/blob/main/android-core/src/main/java/com/mparticle/internal/messages/BaseMPMessage.java

src/NextGen/dataRepository.ts
alexs-mparticle: Just the next line. The Kotlin Spec doesn't have an internal state, but I added one just for testing.

src/NextGen/dataRepository.ts
alexs-mparticle: From recent conversation: Web should start using Messages for "Storage" in the Data Repository. Create a new ticket.

src/NextGen/identity/identity.ts
alexs-mparticle: From conversation: Prefer async/await, fail back to callbacks if we can.

src/NextGen/identity/identity.ts
```diff
@@ -0,0 +1,59 @@
+import { MPID, User } from '@mparticle/web-sdk';
+import { IMParticleComponent } from '../core/component';
+import { IdentityCallback } from './identityCallback';
+
+// QUESTION: What is the diff between this and IdentityApiData?
```
BrandonStalnaker, line 5: I'd say IdentityApiRequest is roughly equivalent to what I assume IdentityApiData is? On mobile a request class usually contains the post data like what Identities are being changed https://github.com/mParticle/mparticle-apple-sdk/blob/main/mParticle-Apple-SDK/Include/MPIdentityApiRequest.h

src/NextGen/identity/identity.ts
BrandonStalnaker: Here's the iOS section on these error codes. https://docs.mparticle.com/developers/sdk/ios/idsync/#error-handling

iOS error handling is done very similarly to android by using a callback

src/NextGen/identity/identity.ts
BrandonStalnaker: Promises is a new concept to me (thought it seems like its a thing in Swift though it kinda still looks like a handler to me https://www.swiftbysundell.com/articles/under-the-hood-of-futures-and-promises-in-swift/ )

src/NextGen/identity/identity.ts
jasonmgeorge: +1 on aligning implementations where we can

src/NextGen/mParticleCallback.ts
einsteinx2: So as we discussed on the call, I strongly believe that we should use function/lambda callbacks on Web and iOS as it's much more idiomatic. The way this part of the architecture was designed, using anonymous subclasses is a _very_ Java/Kotlin platform idiomatic thing to do that doesn't translate well to the other languages. On Swift for example a single function/lambda callback with a boolean success parameter and then an optional error parameter is idiomatic and standard. If something similar to that is standard on TS/JS, we should do that instead of defining a `MParticleCallback` class. Instead we can typedef `MParticleCallback` to the callback function signature, which again would be idiomatic on iOS at least and I think also on Web.

src/NextGen/core/baseMPMessage.ts
```diff
@@ -0,0 +1,3 @@
+export class BaseMPMessage {
```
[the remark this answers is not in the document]
alexs-mparticle: Cool. I have a follow up ticket to implement `BaseMPMessage`, so we can hash that out there.

src/NextGen/core/component.ts
```diff
@@ -0,0 +1,4 @@
+export interface IMParticleComponent {
+    // TODO: Should we just define the component type in here?
```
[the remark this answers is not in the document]
alexs-mparticle: When implementing many of these new interfaces, I built a scaffold of the mediator to see how pieces will interact. I suspect we may still need some sort of `component` interface with some property like `type` to make it easier for Web to determine if a component is for `event logging` vs `identity`.

I agree, we can hash that out when that time comes.

src/NextGen/identity/identity.ts
alexs-mparticle: Making them Promises for now. We should make it a point to test backwards compatibility once they have been fully implemented.

src/NextGen/identity/identity.ts
```diff
@@ -0,0 +1,59 @@
...
+    // QUESTION: Should these be callbacks or promises that eventually return something?
+    login(callback: IdentityCallback): void;
+    // login(): IdentityApiResult;
+    logout(callback: IdentityCallback): void;
+    // logout(): IdentityApiResult;
+    getUser(mpid: MPID, callback: IdentityCallback): User;
+    getUsers(callback: IdentityCallback): User[];
+}
+
+export interface InternalIdentity extends IMparticleIdentity {
```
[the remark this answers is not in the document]
alexs-mparticle: I love that it spells IMP 😈

src/NextGen/identity/identity.ts
rmi22186: Yes everything transpiles.  We just have to ensure we have the proper test coverage for browsers when it comes to the transpiling.

src/NextGen/mParticleCallback.ts
alexs-mparticle: @einsteinx2 and I have decided to remove `MParticleCallback` and the related interfaces as the kotlin implementation is not idiomatic to Web or iOS. Also, as stated above, async/await should be the solution going forward, and Web should support promises natively or via polyfill.

src/NextGen/identity/identity.ts
```diff
@@ -1,48 +1,134 @@
...
+    previousUser: User;
 
-// QUESTION: Should we mirror the way Android handles HTTP codes, or do it the way we've been doing it?
-// https://docs.mparticle.com/developers/sdk/android/idsync/#idsync-status-codes
-// https://docs.mparticle.com/developers/sdk/web/idsync/#error-handling
+    getUser(): User;
+    getPreviousUser(): User;
+}
+
+export enum MParticleServerErrorCodes {
```
rmi22186, line 13: These are actually client error codes, not server error codes

src/NextGen/identity/identity.ts
```diff
@@ -1,48 +1,134 @@
...
+    | 'other8'
+    | 'other9'
+    | 'other10'
+    | 'phone_number_2'
+    | 'phone_number_3'
+    | 'mobile_number'
+    | 'device_application_stamp';
+
+type KnownIdentities = {
+    [key in IdentityTypes]?: string;
```
rmi22186, line 67: This will actually always have a `device_application_stamp` on it at a minimum.  everything else is optional

src/NextGen/identity/identity.ts
```diff
@@ -1,48 +1,134 @@
...
-export interface IdentityApiRequest {}
+export interface IdentityApiResult {
+    user: User;
+    previousUser: User;
 
-// QUESTION: Should we mirror the way Android handles HTTP codes, or do it the way we've been doing it?
-// https://docs.mparticle.com/developers/sdk/android/idsync/#idsync-status-codes
-// https://docs.mparticle.com/developers/sdk/web/idsync/#error-handling
+    getUser(): User;
+    getPreviousUser(): User;
```
rmi22186: This is for general discussion fro the group, but I don't see why there is a need to have a `user` and `previousUser` if we have the get methods for them.

src/NextGen/identity/identity.ts
```diff
@@ -1,48 +1,134 @@
...
+    known_identities: KnownIdentities;
+}
 
 export interface IMParticleIdentity extends IMParticleComponent {
-    login(request: IdentityApiRequest): Promise<IdentityApiResult>;
-    logout(request: IdentityApiRequest): Promise<IdentityApiResult>;
+    login(request: IdentityApiRequest): Promise<IMPIdentityApiResponse>;
+    logout(request: IdentityApiRequest): Promise<IMPIdentityApiResponse>;
     getUser(mpid: MPID): Promise<User>;
     getUsers(): Promise<User[]>;
```
rmi22186: Another general question for the group - I'm still not sure why these are async calls.  the user is in whatever persistence layer the platform has.  we never make a call to the identity platform for user or users.

src/NextGen/identity/identity.ts
```diff
@@ -1,48 +1,134 @@
...
 
-    public modify(request: IdentityApiRequest): Promise<IdentityApiResult> {
-        return Promise.resolve({});
+    public modify(
+        request: IdentityApiRequest
+    ): Promise<IMPIdentityApiResponse> {
+        // return Promise.resolve(this.generateFakeAPIResult());
+        return Promise.reject({
+            errorCode: MParticleServerErrorCodes.activeSession,
+        });
```
rmi22186, line 127: should line 118 be uncommented and this be refactored to not have both reject/resolve?

src/NextGen/identity/identity.ts
```diff
@@ -1,48 +1,134 @@
...
 
 export interface IdentityListener {}
 
 export class IdentityImplementation implements InternalIdentity {
     constructor() {}
 
-    public identify(request: IdentityApiRequest): Promise<IdentityApiResult> {
-        return Promise.resolve({});
+    // TODO: Temporary implementation to test signature
+    private generateFakeAPIResult = () => ({
```
rmi22186, line 104: add type of `IMPIdentityApiResponse` to this signature

src/NextGen/dataRepository.ts
```diff
@@ -0,0 +1,26 @@
...
+export class MParticleDataRepository {
+    public internalState: BaseMPMessage[] = [];
+
+    public insertCommerceEvent(data: BaseMPMessage) {
+        this.internalState.push(data);
+    }
+
+    public insertBreadcrumb(data: BaseMPMessage): void {
+        this.internalState.push(data);
+    }
```
jasonmgeorge, line 13: I'm curious why `insertBreadcrum` explicitly returns void where `insertCommerceEvent` does not.

src/NextGen/identity/identity.ts
einsteinx2: Yep no need for callbacks then at all on web

src/NextGen/mParticleCallback.ts
einsteinx2: agreed

einsteinx2 (approved):
Reviewed on Zoom with Alex yesterday, removed the callbacks entirely, looks good for now, as we do the Swift implementation we'll communicate around potential changes for consistency (if it makes sense), but until then this is fine as-is.

jasonmgeorge (commented):
Reviewed. No additional comments from me.

src/NextGen/identity/identity.ts
alexs-mparticle: My intention was to hard code a single one of these methods to reject rather than resolve so we can visually see both use success and error cases. We can then remove the hardcoded reject/resolve once the real implementation is in place.

src/NextGen/dataRepository.ts
rmi22186: that was probably just an oversight, i added the suggestion

src/NextGen/dataRepository.ts
```diff
@@ -0,0 +1,26 @@
+import { Batch } from '@mparticle/event-models';
+import { BaseMPMessage } from './core/baseMPMessage';
+
+export class MParticleDataRepository {
+    public internalState: BaseMPMessage[] = [];
+
+    public insertCommerceEvent(data: BaseMPMessage) {
```
rmi22186, line 7:
```suggestion
    public insertCommerceEvent(data: BaseMPMessage): void {
```

rmi22186 (approved):
approved with the 1 `void` suggestion

markvdouw (approved):
LGTM

PR #564, opened 2023-05-10
feat: Add MPSideloadedKit class

rmi22186:
## Instructions

 ## Summary
 This PR creates the `MPSideloadedKit` class. This takes an unregistered kitInstance and provides public methods that allow a developer to add hash values for filters of event names/attributes/etc to filter out certain data points from being sent to the specified kit.

 ## Testing Plan

 - Closes https://go.mparticle.com/work/SQDSDKS-5350

### DIFF
--- a/src/configAPIClient.ts
+++ b/src/configAPIClient.ts
@@ -14,7 +14,7 @@ export type SDKCompleteInitCallback = (
     mpInstance: MParticleWebSDK
 ) => void;
 
-export interface IKitConfigs {
+export interface IKitConfigs extends IKitFilterSettings {
     name: string;
     moduleId: number;
     isDebug: boolean;
@@ -22,22 +22,25 @@ export interface IKitConfigs {
     isDebugString: BooleanStringLowerCase;
     hasDebugString: BooleanStringLowerCase;
     settings: Dictionary;
+    eventSubscriptionId: number;
+    excludeAnonymousUser: boolean;
+}
+
+export interface IKitFilterSettings {
+    eventTypeFilters: number[];
+    eventNameFilters: number[];
     screenNameFilters: number[];
     screenAttributeFilters: number[];
     userIdentityFilters: number[];
     userAttributeFilters: number[];
-    eventNameFilters: number[];
-    eventTypeFilters: number[];
     attributeFilters: number[];
-    filteringEventAttributeValue: IFilteringEventAttributeValue;
-    filteringUserAttributeValue: IFilteringUserAttributeValue;
-    filteringConsentRuleValues: IFilteringConsentRuleValues;
+    filteringEventAttributeValue?: IFilteringEventAttributeValue;
+    filteringUserAttributeValue?: IFilteringUserAttributeValue;
+    filteringConsentRuleValues?: IFilteringConsentRuleValues;
     consentRegulationFilters: number[];
     consentRegulationPurposeFilters: number[];
     messageTypeFilters: number[];
     messageTypeStateFilters: number[];
-    eventSubscriptionId: number;
-    excludeAnonymousUser: boolean;
 }
 
 export interface IFilteringEventAttributeValue {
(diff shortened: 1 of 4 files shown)


src/sideloadedKit.ts
```diff
@@ -0,0 +1,68 @@
...
+        const hashedIdentityType = KitFilterHelper.hashUserIdentity(userIdentity);
+        this.filterDictionary.userIdentityFilters.push(hashedIdentityType);
+    }
+
+    public addUserAttributeFilter(userAttributeKey: string) {
+        const hashedUserAttributeKey = KitFilterHelper.hashUserAttribute(userAttributeKey);
+        this.filterDictionary.userAttributeFilters.push(hashedUserAttributeKey);
+    }
+
+    // public addCommerceEventAttributeFilter() {
```
rmi22186, line 66: @einsteinx2 - I didn't see this as an array on the JS config, here is a sample. 
 https://jssdkcdns.mparticle.com/js/v2/e207c24e36a7a8478ba0fcb3707a616b/config

After updating the UI to add some ecommerce specific filters, I didn't see these pop up in any array filters after waiting for cache to bust either.  Maybe this is only an iOS/Android thing, (which would be a feature parity issue), but I'd like to see if anything pops up in the returned config for Android/iOS - I don't know how to get a sample one back though.  Maybe you can help...

src/configAPIClient.ts
```diff
@@ -14,30 +14,33 @@ export type SDKCompleteInitCallback = (
     mpInstance: MParticleWebSDK
 ) => void;
 
-export interface IKitConfigs {
+export interface IKitConfigs extends IKitFilterSettings {
```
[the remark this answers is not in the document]
rmi22186: I split this out because the [filterDictionary](https://github.com/mParticle/mparticle-web-sdk/pull/564/files#diff-a5066c42211de5a8e75e8ee953c4a6038b5bbbc49e353fe1bf034392b121c9fbR10) only contains the `IKitFilterSettings` and not all settings. Another option was to keep this as is, and do an Omit<IKitConfigs, //several items> but that seemed too verbose.

src/sideloadedKit.ts
einsteinx2: Looks like its actually a dictionary, keyed on the hash, with a 0 as the value if it's active. See our iOS unit test here: https://github.com/mParticle/mparticle-apple-sdk/blob/8ce7ad752ceeb1d9dc6d782d9c4462eb0ac114b9/UnitTests/MPKitContainerTests.m#L1257

I don't know if that's how it's done on Web though

einsteinx2:
I replied to your question, but otherwise this looks good.

markvdouw (approved):

einsteinx2 (approved):
LGTM

src/sideloadedKit.ts
```diff
@@ -0,0 +1,65 @@
+import KitFilterHelper from './kitFilterHelper';
+import { IKitFilterSettings } from './configAPIClient';
+import { UnregisteredKit } from './forwarders.interfaces';
+import { EventTypeEnum, IdentityType } from './types.interfaces';
+
+/**
```
alexs-mparticle, line 6: If you aren't going to set up comment doc, please remove this.

alexs-mparticle (changes requested):

alexs-mparticle (approved):

src/sideloadedKit.ts
```diff
@@ -0,0 +1,100 @@
+import KitFilterHelper from './kitFilterHelper';
+import {
+    IFilteringConsentRuleValues,
+    IFilteringEventAttributeValue,
+    IFilteringUserAttributeValue,
+    IKitFilterSettings,
+} from './configAPIClient';
+import { UnregisteredKit } from './forwarders.interfaces';
+import { EventTypeEnum, IdentityType } from './types.interfaces';
+export class MPSideloadedKit {
```
alexs-mparticle, line 10:
```suggestion
import { EventTypeEnum, IdentityType } from './types.interfaces';

export class MPSideloadedKit {
```

alexs-mparticle (changes requested):

PR #572, opened 2023-05-15
fix: Resolve flakey cookie sync tests

rmi22186:
## Instructions

 ## Summary
This PR mocks the create element function for `img` elements and forces the onload to trigger after 10ms.  This fixes the flakey tests and avoids the need for an actual URL.  I have also reduced the timeout we set to check the persistence layer to drastically reduce the amount of time the cookie sync tests take to run!

 ## Testing Plan


package.json
```diff
@@ -135,6 +135,7 @@
     },
     "dependencies": {
         "@babel/runtime": "^7.6.0",
+        "jest-environment-jsdom": "^29.5.0",
```
alexs-mparticle, line 138: This should already exist in `devDependencies`

alexs-mparticle (changes requested):

test/src/tests-cookie-syncing.js
```diff
@@ -39,7 +65,7 @@ describe('cookie syncing', function() {
             isProduction: true,
             settings: {},
             frequencyCap: 14,
-            pixelUrl: pixelUrl,
+            pixelUrl: '',
```
alexs-mparticle, line 68: I think this `pixelSettings` object is the same everywhere. Would it make sense to promote this to the top of the file?

test/src/tests-cookie-syncing.js
```diff
@@ -176,7 +172,7 @@ describe('cookie syncing', function() {
...
```
alexs-mparticle, line 175: Remove this object and just reference the global `pixelSettings` object.

alexs-mparticle, line 175: Remove this object and just reference the global `pixelSettings` object.

test/src/tests-cookie-syncing.js
```diff
@@ -206,7 +202,7 @@ describe('cookie syncing', function() {
             isProduction: false,
             settings: {},
             frequencyCap: 14,
-            pixelUrl: pixelUrl,
+            pixelUrl: '',
```
alexs-mparticle, line 205: Remove this object and just reference the global `pixelSettings` object.

test/src/tests-cookie-syncing.js
```diff
@@ -273,7 +269,7 @@ describe('cookie syncing', function() {
                     isProduction: true,
                     settings: {},
                     frequencyCap: 14,
-                    pixelUrl: pixelUrl,
+                    pixelUrl: '',
```
alexs-mparticle, line 272: Remove this object and just reference the global `pixelSettings` object.

test/src/tests-cookie-syncing.js
```diff
@@ -1228,7 +1174,7 @@ describe('cookie syncing', function() {
             isProduction: true,
             settings: {},
             frequencyCap: 14,
-            pixelUrl: pixelUrl,
+            pixelUrl: '',
```
alexs-mparticle, line 1177: Remove this object and just reference the global `pixelSettings` object.

test/src/tests-cookie-syncing.js
```diff
@@ -1253,7 +1199,7 @@ describe('cookie syncing', function() {
...
```
alexs-mparticle, line 1202: Remove this object and just reference the global `pixelSettings` object.

PR #599, opened 2023-05-30
feat: Implement Next Gen Middleware

alexs-mparticle:
## Instructions

 ## Summary
 - First pass implementation of middleware interfaces and classes for Next Gen Web SDK Patterns.

 ## Testing Plan
 - This pull request only implements initial classes and interfaces for review. Functionality will not be integrated with the core sdk at this point.

 - Closes https://go.mparticle.com/work/SQDSDKS-5366

### DIFF
--- a/src/NextGen/batchManager.ts
+++ b/src/NextGen/batchManager.ts
@@ -1,8 +1,10 @@
-import { BatchEnvironmentEnum } from '@mparticle/event-models';
-import { Batch } from '@mparticle/event-models';
+import {
+    Batch,
+    BatchEnvironmentEnum,
+    BaseEvent,
+} from '@mparticle/event-models';
 import { BaseMPMessage } from './core/baseMPMessage';
 import { isEmpty } from '../utils';
-import { BaseEvent } from '../sdkRuntimeModels';
 
 // TODO: Sync this with existing API Client(s)
 export interface IMpApiClientImplementation {
@@ -24,6 +26,7 @@ export default class BatchManager {
             return {
                 environment: BatchEnvironmentEnum.unknown,
                 mpid: 'test-mpid',
+                events: (events as unknown) as BaseEvent[],
             };
         } else {
             return null;
--- a/src/NextGen/batchManager.ts
+++ b/src/NextGen/batchManager.ts
@@ -1,0 +1,49 @@
+import { BatchEnvironmentEnum } from '@mparticle/event-models';
+import { Batch } from '@mparticle/event-models';
+import { BaseMPMessage } from './core/baseMPMessage';
+import { isEmpty } from '../utils';
+import { BaseEvent } from '../sdkRuntimeModels';
+
+// TODO: Sync this with existing API Client(s)
+export interface IMpApiClientImplementation {
+    uploadBatch(batch: Batch): void;
+}
+
+export default class BatchManager {
+    constructor(private api: IMpApiClientImplementation) {
+        this.api = api;
+    }
+
+    // QUESTION: Should this take an event or message as a param?
+    // Making an assumption that creating a batch should require a message
+    public createBatch(events: BaseMPMessage[]): Batch {
+        // QUESTION: In the existing Batch Uploader, we have a method to create new batches based on events
+        //           should this method reuse that code or will this be a much simpler create method?
+
+        if (!isEmpty(events)) {
+            return {
+                environment: BatchEnvironmentEnum.unknown,
+                mpid: 'test-mpid',
+            };
+        } else {
+            return null;
+        }
+    }
+
+    // QUESTION: We should consider making this take an array or queue of batches, as we are required to upload batches in sequence if they are queued
+    public uploadBatch(batch: Batch): void {
+        this.api.uploadBatch(batch);
+    }
+}
+
+// QUESTION: How would MpBatch differ from the existing Batch definition?
+class MpBatch {
+    // QUESTION: Why are we setting data to string? Shouldn't it be a JSON object?
+    public data: string = '';
+    public batchId: number = null;
+
+    constructor(data: string, batchId: number) {
+        this.data = data;
+        this.batchId = batchId;
+    }
+}
--- a/src/NextGen/dataHandler/dataHandler.ts
+++ b/src/NextGen/dataHandler/dataHandler.ts
@@ -2,7 +2,6 @@ import { IMParticleComponent } from '../core/component';
 
 export interface IMParticleDataHandler extends IMParticleComponent {
     // QUESTION: Shouldn't we be passing in a type for data instead of being "any"?
-    // QUESTION: Shouldn't immediateUpload be optional?
-    saveData(data: any, immediateUpload: boolean);
+    saveData(data: any, immediateUpload?: boolean);
     configure(): void;
 }
--- a/src/NextGen/dataHandler/dataHandler.ts
+++ b/src/NextGen/dataHandler/dataHandler.ts
@@ -1,0 +1,8 @@
+import { IMParticleComponent } from '../core/component';
+
+export interface IMParticleDataHandler extends IMParticleComponent {
+    // QUESTION: Shouldn't we be passing in a type for data instead of being "any"?
+    // QUESTION: Shouldn't immediateUpload be optional?
+    saveData(data: any, immediateUpload: boolean);
+    configure(): void;
+}
--- a/src/NextGen/dataHandler/dataHandlerStrategy.ts
+++ b/src/NextGen/dataHandler/dataHandlerStrategy.ts
@@ -3,7 +3,7 @@ import { BaseMPMessage } from '../core/baseMPMessage';
 import { DataHandlerType } from './dataHandlerType';
 
 export interface IMParticleDataHandlerStrategy<Input, Output> {
-    saveData(data: Input, immediateUpload: boolean): void;
+    saveData(data: Input, immediateUpload?: boolean): void;
     retrieveData(): Output[];
 
     // QUESTION: What is this trying to do?
@@ -30,7 +30,7 @@ export abstract class BaseMparticleDataHandlerStrategy<Input, Output>
     }
 
     // QUESTION: Shouldn't data be of type Input?
-    abstract saveData(data: Input, immediateUpload: boolean): void;
+    abstract saveData(data: Input, immediateUpload?: boolean): void;
     abstract retrieveData(): Output[];
     abstract type(): DataHandlerType;
 }
--- a/src/NextGen/dataHandler/dataHandlerStrategy.ts
+++ b/src/NextGen/dataHandler/dataHandlerStrategy.ts
@@ -1,0 +1,36 @@
+import BatchManager from '../batchManager';
+import { BaseMPMessage } from '../core/baseMPMessage';
+import { DataHandlerType } from './dataHandlerType';
+
+export interface IMParticleDataHandlerStrategy<Input, Output> {
+    saveData(data: Input, immediateUpload: boolean): void;
+    retrieveData(): Output[];
+
+    // QUESTION: What is this trying to do?
+    // fun I.toDto(): O?
+    // fun O.toModel(): I?
+
+    // QUESTION: Is this just a getter for type or strategy ID?
+    type(): DataHandlerType;
+}
+
+export abstract class BaseMparticleDataHandlerStrategy<Input, Output>
+    implements IMParticleDataHandlerStrategy<Input, Output> {
+    public batchManager: BatchManager;
+
+    constructor(batchManager: BatchManager) {
+        this.batchManager = batchManager;
+    }
+
+    // QUESTION: We currently queue multiple batches to upload. Should we split these into separate methods?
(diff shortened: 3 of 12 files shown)


src/NextGen/batchManager.ts
```diff
@@ -0,0 +1,52 @@
...
+                mpid: 'test-mpid',
+                events: (events as unknown) as BaseEvent[],
+            };
+        } else {
+            return null;
+        }
+    }
+
+    // QUESTION: We should consider making this take an array or queue of batches, as we are required to upload batches in sequence if they are queued
+    public uploadBatch(batch: Batch): void {
```
einsteinx2, line 37: I guess the question boils down to whether we should have implementation code that loops through the array/queue and calls this method or whether we should have this method just take an array/queue and have the logic there?

src/NextGen/batchManager.ts
```diff
@@ -0,0 +1,52 @@
...
+        this.api.uploadBatch(batch);
+    }
+}
+
+// QUESTION: How would MpBatch differ from the existing Batch definition?
+class MpBatch {
```
einsteinx2, line 43: I assume we would be replacing the existing Batch definition with this one.

src/NextGen/batchManager.ts
```diff
@@ -0,0 +1,52 @@
...
+    // QUESTION: Why are we setting data to string? Shouldn't it be a JSON object?
+    public data: string = '';
```
einsteinx2, line 45: Good question, we'll need to confirm this.

src/NextGen/dataHandler/dataHandler.ts
```diff
@@ -0,0 +1,8 @@
+import { IMParticleComponent } from '../core/component';
+
+export interface IMParticleDataHandler extends IMParticleComponent {
+    // QUESTION: Shouldn't we be passing in a type for data instead of being "any"?
+    // QUESTION: Shouldn't immediateUpload be optional?
+    saveData(data: any, immediateUpload: boolean);
```
einsteinx2, line 6: Agreed, we should definitely have a type for `data`.

For immediateUpload, is it possible to set default parameter values in TS?

src/NextGen/examples/MParticleCommerceHandler.ts
```diff
@@ -0,0 +1,40 @@
...
+        batchManager: BatchManager
+    ) {
+        super(batchManager);
+        this.dataRepository = dataRepository;
+        this.batchManager = batchManager;
+    }
+
+    // QUESTION: Should we make immideateUpload optional with a default of false?
+    // QUESTION: Why should data be any rather than an actual commerce vent?
+    public saveData(data: any, immediateUpload: boolean): void {
```
einsteinx2, line 24: agreed on immediateUpload being option with default to false as really the only time it's true is for CommerceEvents I think.

Same comment as before re: any types, we should never use any

src/NextGen/examples/MParticleCommerceHandler.ts
```diff
@@ -0,0 +1,40 @@
...
+        this.dataRepository.insertCommerceEvent(data);
+
+        if (immediateUpload) {
+            // QUESTION: I think create and upload should be split up as there might be queued batches already
+            this.createAndUploadBatch(data);
```
einsteinx2, line 29: That's a good call out, we should discuss this.

src/NextGen/dataHandler/dataHandler.ts
alexs-mparticle: Yes, we can set a default of `false` for `immediateUpload` in TS. I personally think we should do this. I'm cool with revising my PR if that's the direction we want to go.

src/NextGen/batchManager.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+
+export default class BatchManager {
+    constructor(private api: IMpApiClientImplementation) {
+        this.api = api;
+    }
+
+    // QUESTION: Should this take an event or message as a param?
+    // Making an assumption that creating a batch should require a message
+    public createBatch(events: BaseMPMessage[]): Batch {
```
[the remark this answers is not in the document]
alexs-mparticle: I mainly bring up the question because in the android spec, `createBatch` does not take any params and I think that's probably because the spec was a rough draft.

On web, we actually take an [array of events and a user object](https://github.com/mParticle/mparticle-web-sdk/blob/master/src/batchUploader.ts#L199) so that we can order the batch's events based on session, so that's another process we should standardize on.

src/NextGen/batchManager.ts
```diff
@@ -0,0 +1,52 @@
...
+        // QUESTION: In the existing Batch Uploader, we have a method to create new batches based on events
+        //           should this method reuse that code or will this be a much simpler create method?
+
+        if (!isEmpty(events)) {
```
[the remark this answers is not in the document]
alexs-mparticle: Agreed. My current implementation is geared towards fleshing out the broad strokes from the android spec and seeing where the pain points exist.

src/NextGen/batchManager.ts
alexs-mparticle: I think the answer depends on where we want to hold the "business logic" for uploading batches, since sequencing is a core piece of functionality. I think there's a disconnect on my part regarding the difference between the "BatchManager" and "UploadStrategy" we discussed.

src/NextGen/batchManager.ts
alexs-mparticle: @markvdouw Can you chime in here?

src/NextGen/examples/MParticleCommerceHandler.ts
alexs-mparticle: @markvdouw thoughts?

src/NextGen/examples/MParticleCommerceHandler.ts
```diff
@@ -0,0 +1,40 @@
...
+        dataRepository: MParticleDataRepository,
+        batchManager: BatchManager
+    ) {
+        super(batchManager);
+        this.dataRepository = dataRepository;
+        this.batchManager = batchManager;
+    }
+
+    // QUESTION: Should we make immideateUpload optional with a default of false?
+    // QUESTION: Why should data be any rather than an actual commerce vent?
```
rmi22186, line 23: probably a typo. I agree it should be specific to the event this data handler is handling.

src/NextGen/dataHandler/dataHandlerStrategy.ts
```diff
@@ -0,0 +1,36 @@
...
+    public batchManager: BatchManager;
+
+    constructor(batchManager: BatchManager) {
+        this.batchManager = batchManager;
+    }
+
+    // QUESTION: We currently queue multiple batches to upload. Should we split these into separate methods?
+    // QUESTION: How does the event get passed into the batch manager for creation?
+    protected createAndUploadBatch(event: BaseMPMessage) {
+        const batch = this.batchManager.createBatch([event]);
```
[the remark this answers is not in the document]
alexs-mparticle: Great point. We need to line up on what parameters are actually necessary to create a batch. In another spot, I mentioned that in Web, we require an array of events and a user object to create a batch, but Android does not seem to have this same requirement.

test/jest/NextGen/MParticleCommerceHandler.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+                    null
+                );
+
+                expect(commerceHandler.retrieveData()).toEqual([
+                    event,
+                    event,
+                    event,
+                ]);
+            });
+
```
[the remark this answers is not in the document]
alexs-mparticle: Great call out. We should flesh this out in a future revision, as our current spec only has commerce events as an example.

src/NextGen/batchManager.ts
markvdouw, line 37: Same here with the Repository pattern idea.

src/NextGen/examples/MParticleCommerceHandler.ts
```diff
@@ -0,0 +1,40 @@
...
+import BatchManager from '../batchManager';
+import { DataHandlerType } from '../dataHandler/dataHandlerType';
+
+export default class MParticleCommerceHandler extends BaseMparticleDataHandlerStrategy<
+    CommerceEvent,
+    BaseMPMessage
+> {
+    private dataRepository: MParticleDataRepository;
+    constructor(
+        dataRepository: MParticleDataRepository,
```
markvdouw, line 14: I made a couple of comments about the repository being injected in the BatchManager. I think that maybe we should inject it in that layer instead of here.

src/NextGen/batchManager.ts
BrandonStalnaker: I think this code should just upload the batch and the queuing should occur on the level that's calling this method.

src/NextGen/dataHandler/dataHandler.ts
```diff
@@ -0,0 +1,7 @@
+import { IMParticleComponent } from '../core/component';
+
+export interface IMParticleDataHandler extends IMParticleComponent {
+    // QUESTION: Shouldn't we be passing in a type for data instead of being "any"?
```
BrandonStalnaker, line 4: Maybe Mark was using 'any' to represent it necessarily being a different type of data depending on platform? @markvdouw ?

src/NextGen/dataHandler/dataHandler.ts
```diff
@@ -0,0 +1,7 @@
+import { IMParticleComponent } from '../core/component';
+
+export interface IMParticleDataHandler extends IMParticleComponent {
```
rmi22186, line 3: @markvdouw - can you clarify the difference between data handler and data handler strategy from a 30k foot view?  I'm having a hard time understanding the distinction.

src/NextGen/batchManager.ts
```diff
@@ -0,0 +1,52 @@
...
+} from '@mparticle/event-models';
+import { BaseMPMessage } from './core/baseMPMessage';
+import { isEmpty } from '../utils';
+
+// TODO: Sync this with existing API Client(s)
+export interface IMpApiClientImplementation {
+    uploadBatch(batch: Batch): void;
+}
+
+export default class BatchManager {
```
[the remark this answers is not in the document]
alexs-mparticle: @markvdouw I thought that was the job of the mediator, to take data from one component to another.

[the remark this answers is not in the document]
markvdouw: The mediator should be juggling with the EXTERNAL public components (Identity, Consent, Media, EventLogging, etc), that customer will basically relate to our public API. 
Though the BatchHandler can be considered one of the main components, it should, in my opinion, delegate data handling. One of the patterns widely used for data sources handling is the Repository Pattern (its basically a handler/manager for multiple data sources such as DAOs, Cache, ApiClients, etc.)

I would advice injecting a data repository in every component that needs to interact with data. 
This is THE case of data manipulation (writing into the database, reading from it/creating a batch and pass it to the apiClient so it can be updated, etc)

src/NextGen/batchManager.ts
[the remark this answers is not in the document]
markvdouw: @alexs-mparticle That's it. It was an outline to show interactions only. We can determine necessary params/dependencies

src/NextGen/dataHandler/dataHandler.ts
markvdouw: The DataHandler defines the operation that must be implemented by the class that hold the concrete strategies.
The DataHandlerStrategy defines the operations that must be implemented by each concrete strategy.

Example:
DataManager (strategies : List<DataHandlerStrategy>) implements DataHandler 

In this example makes sense that the DataHandlerStrategy defines all the operations in DataHandler PLUS any other specific operation needed in the concrete strategies (for example to map/convert data)
https://github.com/mParticle/mparticle-android-sdk/pull/367/files#diff-1e033fd6678f20a7e1dd496e9e3d1818f5241f811511b3a4891e3c811104b75a

src/NextGen/dataHandler/dataHandler.ts
markvdouw: If we want to make this component 100% agnostic for the community, we need to be able to pass ANY and then define within each strategy how to map that ANY to a specific model.
If we want to scope it within MParticle, we can define a little bit better using Generics

src/NextGen/dataHandler/dataHandlerStrategy.ts
```diff
@@ -0,0 +1,36 @@
+import BatchManager from '../batchManager';
+import { BaseMPMessage } from '../core/baseMPMessage';
+import { DataHandlerType } from './dataHandlerType';
+
+export interface IMParticleDataHandlerStrategy<Input, Output> {
+    saveData(data: Input, immediateUpload?: boolean): void;
+    retrieveData(): Output[];
+
+    // QUESTION: What is this trying to do?
```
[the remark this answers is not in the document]
markvdouw: Kind of, the advantage is that we tie the strategy to specific models, mapping and serialization steps.

markvdouw (commented):
Some extra comments/answers

PR #605, opened 2023-06-02
test: Fix Flakey Offline Storage Tests

alexs-mparticle:
## Instructions

 ## Summary
 - Noticed that an asynchronous test was firing an event but was not actually caught by the `done()` function, and it was colliding with a subsequent test.

 ## Testing Plan
 - N/A - Just changing tests

 - Closes https://go.mparticle.com/work/SQDSDKS-5383


rmi22186 (approved):
w00t!

PR #613, opened 2023-06-07
Add Sideloaded Kit flag to Batches

alexs-mparticle:
## Instructions

 ## Summary
 - Add a boolean flag to batches indicating that Sideloaded kits are in use
 - Requires a corresponding PR in the [event-models repo](https://github.com/mParticle/js-models)

 ## Testing Plan
 - Set up a sample app with a side loaded kit and verify that all batches contain `application_info` with `is_using_sideloaded_kits: true`.

 - Closes https://go.mparticle.com/work/SQDSDKS-5388

### DIFF
--- a/src/sdkToEventsApiConverter.ts
+++ b/src/sdkToEventsApiConverter.ts
@@ -56,6 +56,13 @@ export function convertEvents(
         currentConsentState = user.getConsentState();
     }
 
+    // TODO: Apply this new attribute to event-models
+    // https://go.mparticle.com/work/SQDSDKS-5458
+    interface AppInfoWithSideLoadedKitBool
+        extends EventsApi.ApplicationInformation {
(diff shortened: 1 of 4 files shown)


src/sdkToEventsApiConverter.ts
```diff
@@ -56,6 +56,13 @@ export function convertEvents(
         currentConsentState = user.getConsentState();
     }
 
+    // TODO: Apply this new attribute to event-models
```
rmi22186, line 59: I approved https://github.com/mParticle/js-models/pull/119, so I think we should release that one and then this can be removed once the package.json is updated.

rmi22186 (commented):
This PR looks great! Just the one comment about using the updated JS-models now that that is approved.

rmi22186 (approved):

PR #617, opened 2023-06-08
fix: Support suffix in kit initialization

rmi22186:
## Instructions

 ## Summary
In order to support allowing multiple versions of our client side kits to be served, we add support for initializing kits based on suffix.  When a client side kit has 2 options (ex Braze v3 and v4), prior to this code, because the `kits` (previously `forwarders`) object is keyed by name, Braze v3 and Braze v4 have the same name (Appboy), and one kit would be replaced when the second Braze kit was added.  Now we suffix the kits so that in the above example, the `kits` object would be:
```
{
  'Appboy-v3': {},
  'Appboy-v4': {}
]
```

I also renamed `forwarder` --> `kits` where I could in this method to avoid any confusion since we are aligning on `kits` instead of `forwarders`.

 ## Testing Plan
This PR was tested along with https://github.com/mparticle-integrations/mparticle-javascript-integration-braze/pull/21 and https://github.com/mparticle-integrations/mparticle-javascript-integration-braze/pull/22 to ensure it worked with multiple suffixes of Braze.

 - Closes https://go.mparticle.com/work/SQDSDKS-5444

### DIFF
--- a/src/forwarders.js
+++ b/src/forwarders.js
@@ -572,9 +572,9 @@ export default function Forwarders(mpInstance, kitBlocker) {
     };
 
     this.configureUIEnabledKit = function(configuration) {
-        var newKit = null,
-            config = configuration,
-            kits = {};
+        let newKit = null;
+        let kits = {};
+        const config = configuration;
 
         // If there are kits inside of mpInstance._Store.SDKConfig.kits, then mParticle is self hosted
         if (!isEmpty(mpInstance._Store.SDKConfig.kits)) {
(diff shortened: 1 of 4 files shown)


src/forwarders.js
```diff
@@ -572,36 +572,52 @@ export default function Forwarders(mpInstance, kitBlocker) {
...
             ) {
-                forwarders[forwarder.name] = forwarder;
+                // A suffix is added to a kitConstructor and kit config if there are multiple different
+                // versions of a client kit.  This matches the suffix in the DB.  As an example
+                // the GA4 kit has a client kit and a server side kit which has a client side
+                // component.  They share the same name/module ID in the DB, so we include a
+                // suffix to distinguish them in the kits object.
+                // If a customer wanted simultaneous GA4 client and server connections,
+                // a suffix allows the SDK to distinguish the two.
+                if (kitConstructor.suffix) {
```
alexs-mparticle, line 594: Does it make sense to check if both name and suffix exist?

rmi22186: name will always exist.  it's suffix that is optional.

rmi22186: i added suffix as an optional parameter in the IKitConfig.  Will be useful when this file is migrated to TS.

alexs-mparticle (approved):

PR #618, opened 2023-06-08
fix: Revise priority of device id when initializing state

alexs-mparticle:
## Instructions

 ## Summary
 - Change order of priority for device id when constructing SDK state.

 ## Testing Plan
 - Using a sample app, verify that sdk generates a random guid and saves it in local storage by comparing local storage value to running `mParticle.getDeviceId()` via developer console.
 - Then add `deviceId: test-device-id` to the `config` object and verify that the config device id in dev console is the same the value that was added to the config.

 - Closes https://go.mparticle.com/work/SQDSDKS-5411

### DIFF
--- a/src/persistence.js
+++ b/src/persistence.js
@@ -195,7 +195,7 @@ export default function _Persistence(mpInstance) {
                     mpInstance._Store.clientId ||
                     mpInstance._Helpers.generateUniqueId();
 
-                // For most persitence value, we prioritize localstorage/cookie values over
+                // For most persistence values, we prioritize localstorage/cookie values over
                 // Store. However, we allow device ID to be overriden via a config value and
                 // thus if it has been set before we "store in memory", we should prioritize
                 // the existing value. If neither value exist, we generate a new guid.
--- a/src/persistence.js
+++ b/src/persistence.js
@@ -197,8 +197,10 @@ export default function _Persistence(mpInstance) {
 
                 // For most persistence values, we prioritize localstorage/cookie values over
                 // Store. However, we allow device ID to be overriden via a config value and
-                // thus if it has been set before we "store in memory", we should prioritize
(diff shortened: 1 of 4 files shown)


src/persistence.js
```diff
@@ -194,10 +194,16 @@ export default function _Persistence(mpInstance) {
                     obj.gs.cgid ||
                     mpInstance._Store.clientId ||
                     mpInstance._Helpers.generateUniqueId();
+
+                // For most persitence value, we prioritize localstorage/cookie values over
```
rmi22186, line 198:
```suggestion
                // For most persistence values, we prioritize localstorage/cookie values over
```

test/src/tests-persistence.ts
```diff
@@ -1877,6 +1877,25 @@ describe('migrations and persistence-related', () => {
         done();
     });
 
+    it('should prioritize device id set via mParticle.config instead of local storage', done => {
+        mParticle._resetForTests(MPConfig);
+
+        setLocalStorage();
+
```
rmi22186, line 1884: Can you do a check for the deviceId at this point also to ensure there is a value here?

test/src/tests-persistence.ts
```diff
@@ -1877,6 +1877,25 @@ describe('migrations and persistence-related', () => {
...
+        mParticle.config.deviceId = 'guid-via-config';
```
rmi22186, line 1885: nit: Make this a const that can be reused in lines 1885, 1889, and 1894.

test/src/tests-persistence.ts
rmi22186: To clarify - mParticle.init in order to get a real guid device id. check for that to exist. then reinitialize with a device id on the config.

rmi22186 (approved):

PR #646, opened 2023-06-28
fix: Rename is_using_sideloaded_kits to sideloaded_kits_count

alexs-mparticle:
## Instructions

 ## Summary
 - Rename is_using_sideloaded_kits to sideloaded_kits_count
 - Requires https://github.com/mParticle/js-models/pull/120 to be merged

 ## Testing Plan
 - Set up a sample app with at least two sideloaded kits and verify that all batches contain application_info with`sideloaded_kits_count: X`, where `X` is the exact number of sideloaded kits that were registered.

 - Closes https://go.mparticle.com/work/SQDSDKS-5512


src/sdkToEventsApiConverter.ts
```diff
@@ -653,8 +661,9 @@ export function convertUserAttributeChangeEvent(
 export function convertUserIdentityChangeEvent(
     sdkEvent: SDKEvent
 ): EventsApi.UserIdentityChangeEvent | null {
-    const commonEventData: EventsApi.CommonEventData =
-        convertBaseEventData(sdkEvent);
+    const commonEventData: EventsApi.CommonEventData = convertBaseEventData(
+        sdkEvent
+    );
```
rmi22186, line 666: nit: I feel like these linting issues from auto-linting on save shoudln't be part of this commit, but I also don't really care that much.  I just feel like it detracts from the PR.  We can talk about that though.

rmi22186 (commented):
lgtm overall.  i would put a wait to merge, and add a commit for when https://github.com/mParticle/js-models/pull/120/files is released to update package.json.  then this is :shipit:-able

rmi22186 (approved):
LGTM

PR #778, opened 2023-09-20
fix: Update snippet to support snippet version pinning

rmi22186:
## Instructions

 ## Summary
 In order to allow CDN users to take advantage of pinning a web SDK version, we need to provide a new snippet that will add the query parameters for the new API.  

The new API will be:
mParticle.config.versions = {
     mp_sdk: '2.1.3'
};

The snippet looks for this `versions` object and appends the key and value as a query parameter.  Versions will in the future include kits.  For now only the key `mp_sdk` is supported.

 ## Testing Plan
 - Added the new snippet to an index.html file and used the new API to set the version to 2.1.3 and observed the expected key being appended.  See below for screenshot
image

 - Closes https://go.mparticle.com/work/SQDSDKS-5749


snippet.js
```diff
@@ -90,6 +90,15 @@
         }
     }
 
+    // set version query parameters
+    var versions = window.mParticle.config.versions;
+    if (versions) {
+        var versionQueryArray = [];
```
alexs-mparticle, line 96: `versionQueryArray` should be moved outside of this if block since it's being referenced on line 115.

alexs-mparticle (approved):

PR #780, opened 2023-09-22
ci: Revise dependabot to only weekly production updates

alexs-mparticle:
## Instructions

 ## Summary
 - Update Dependabot frequency to be weekly (on Sunday) instead of daily, so we can process updates on Monday before a release, if necessary.
 - Update dependency types to limit to production only.

 ## Testing Plan
 - N/A as this is a CI/CD update

 - Closes https://go.mparticle.com/work/SQDSDKS-5756

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,6 +4,7 @@ updates:
       directory: '/'
       schedule:
           interval: weekly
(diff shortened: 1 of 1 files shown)


rmi22186 (approved):
LGTM

PR #783, opened 2023-10-03
ci: POC - Add BrowserStack config files

rmi22186:
## Instructions

 ## Summary
This PR is a POC for Cross Browser Testing using Browserstack.  We test Firefox 21 on Mac OS 10.6 and Firefox 100 on Mac OS 10.15.  Tests fail on FF21 since ES6 syntax is not supported.  For now I am commenting out most tests to allow the tests to run quickly.  

 ## Testing Plan
Locally ran browserstack tests which opens cloud instances and results from tests, which show a failure on FF21, but all tests passing on FF100:

image

 - Closes https://go.mparticle.com/work/SQDSDKS-5781


test/cross-browser-testing/browserstack/browserstack.karma.config.js
```diff
@@ -0,0 +1,73 @@
+const { DEBUG } = process.env;
+
+let files = [
```
alexs-mparticle, line 3: As far as I can tell, this isn't being changed, so we can make it a `const`.

```suggestion
const files = [
```

test/cross-browser-testing/browserstack/browserstack.karma.config.js
```diff
@@ -0,0 +1,73 @@
...
+    '../../lib/mockhttprequest.js',
+    '../../lib/geomock.js',
+    '../../../node_modules/fetch-mock/dist/es5/client-bundle.js',
+    '../../../node_modules/sinon-browser-only/sinon.js',
+    '../../config.js',
+    '../../../dist/mparticle.js',
+    '../../test-bundle.js'
+];
+
+let singleRun = true;
```
alexs-mparticle, line 13: Same here. This isn't changing is it?

```suggestion
const singleRun = true;
```

test/src/tests-main.js
```diff
@@ -31,30 +31,31 @@ afterEach(function() {
     window.fetchMock.restore();
 });
 
+import '../cross-browser-testing/test-cross-browser';
 import './tests-core-sdk';
-import './tests-batchUploader';
```
alexs-mparticle, line 35: Instead of modifying `tests-main.js`, can we just import `../cross-browser-testing/test-cross-browser` into the `browserstack.karma.config` directly instead? I'm asking mainly in case we ever want to isolate some tests as "Cross Browser only" in the future.

alexs-mparticle (changes requested):

test/src/tests-main.js
rmi22186: This was more POC than anything else to get tests up and running quickly for vendor selection.  There will be a future PR where I clean things up to make things more modular and true to life for our CI/CD purposes.

package.json
```diff
@@ -36,6 +36,8 @@
         "build:webpack:module": "webpack --config test/integrations/module/webpack/webpack.config.js",
         "build:ts": "tsc -p .",
         "test": "npm run build && npm run build:test-bundle && cross-env DEBUG=false karma start test/karma.config.js",
+        "test:browserstack": "karma start test/cross-browser-testing/browserstack/browserstack.karma.config.js",
```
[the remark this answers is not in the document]
rmi22186: I think this is a good suggestion for the ticket that cleans things up once we select a vendor.  Since this is POC for browserstack, my intent was to make a `test:saucelabs` and `test:karmatest`.  I'll add this as an AC to this ticket - https://mparticle-eng.atlassian.net/browse/SQDSDKS-5797

test/cross-browser-testing/browserstack/browserstack.karma.config.js
```diff
@@ -0,0 +1,73 @@
...
+
+// Allows console logs to appear when doing npm run test:debug
+if (DEBUG === 'true') {
+    browserConsoleLogOptions = {
+        level: 'log',
+        format: '%b %T: %m',
+        terminal: true,
+    };
+    captureConsole = true;
+}
```
alexs-mparticle, line 29: I think these can be better organized for clarity.

```suggestion
let captureConsole = false;
let browserConsoleLogOptions = {};

// Allows console logs to appear when doing npm run test:debug
if (DEBUG === 'true') {
    browserConsoleLogOptions = {
        level: 'log',
        format: '%b %T: %m',
        terminal: true,
    };
    captureConsole = true;
} else {
    browserConsoleLogOptions = {
        terminal: false,
    };
}
```

PR #784, opened 2023-10-05
fix: Prevent Batches from sendBeacon from being stored offline

alexs-mparticle:
## Instructions

 ## Summary
 - We've noticed an increase in duplicate batches coming from Web SDK. This is likely due to batches being stored in Local Storage when a beacon event is fired if a user switches tabs between sessions.
 - This change should prevent batches from being stored in Local Storage during a `sendBeacon` event, and now only store batches when an upload is unsuccessful.

 ## Testing Plan

 - Closes https://go.mparticle.com/work/SQDSDKS-5835


src/batchUploader.ts
```diff
@@ -334,13 +327,8 @@ export class BatchUploader {
...
         }
 
-        // Update Offline Storage with current state of batch queue
-        if (!useBeacon && this.offlineStorageEnabled && this.batchVault) {
-            // Note: since beacon is "Fire and forget" it will empty `batchesThatDidNotUplod`
-            // regardless of whether the batches were successfully uploaded or not. We should
-            // therefore NOT overwrite Offline Storage when beacon returns, so that we can retry
-            // uploading saved batches at a later time. Batches should only be removed from
-            // Local Storage once we can confirm they are successfully uploaded.
+        if (this.offlineStorageEnabled && this.batchVault) {
```
samdozor, line 330: just checking - we shouldn't do this if !useBeacon right? why remove that clause

rmi22186: I think Sam is correct here.  We _want_ to save when useBeacon is false.

alexs-mparticle: I removed it so that regardless if beacon is fired, we purge local storage to prevent duplicates.  I initially put it in so that we avoid touching local storage if beacon is fired. I'll put it back.

rmi22186 (approved):

joelcorporan:
@alexs-mparticle, since when was this issue with duplicate events on mParticle?

alexs-mparticle:
@joelcorporan We first noticed this issue in mid to late September which is why we fast tracked this update. Were you noticing any specific issues and have you reached out to support?

PR #786, opened 2023-10-09
fix: Remove slugify dependency

alexs-mparticle:
## Instructions

 ## Summary
 - Replace Dependency on `Slugify` without using any polyfills or JS functions that are incompatible with ES5.

 ## Testing Plan
 - Using a sample app, attempt to create a config with a data plan that does not conform to our Data Plan naming format.
 - Verify that this will be rejected via an error in the Dev Console.
 - Verify that a data plan that is correct will send valid data to the live stream.

### Invalid Data Plan Slug
Screenshot 2023-10-09 at 4 20 57 PM

### Valid Data Plan Slug

Screenshot 2023-10-09 at 4 22 59 PM

 - Closes https://go.mparticle.com/work/SQDSDKS-4077

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -203,7 +203,7 @@ const isString = (value: any): boolean => typeof value === 'string';
 const isNumber = (value: any): boolean => typeof value === 'number';
 const isFunction = (fn: any): boolean => typeof fn === 'function';
 
-const toSlug = (value: any): string =>
+const toDataPlanSlug = (value: any): string =>
     // Make sure we are only acting on strings or numbers
     isStringOrNumber(value)
         ? value
@@ -212,7 +212,7 @@ const toSlug = (value: any): string =>
               .replace(/[^0-9a-zA-Z]+/g, '_')
         : '';
 
-const isDataPlanSlug = (str: string): boolean => str === toSlug(str);
+const isDataPlanSlug = (str: string): boolean => str === toDataPlanSlug(str);
 
 const isStringOrNumber = (value: any): boolean =>
     isString(value) || isNumber(value);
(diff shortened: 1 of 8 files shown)


alexs-mparticle:
Closing and reopening to re-run checks

src/utils.ts
```diff
@@ -204,9 +203,16 @@ const isString = (value: any): boolean => typeof value === 'string';
 const isNumber = (value: any): boolean => typeof value === 'number';
 const isFunction = (fn: any): boolean => typeof fn === 'function';
 
-// TODO: Refactor this to a regex
-const isDataPlanSlug = (str: string): boolean =>
-    isString(str) && str === slugify(str);
+const toSlug = (value: any): string =>
```
rmi22186, line 206: Technically these are not slugs, because slugs can have hyphens.  I think we should really call this something more specific to our purposes:

```suggestion
const toDataPlanId = (value: any): string =>
```

src/utils.ts
```diff
@@ -204,9 +203,16 @@ const isString = (value: any): boolean => typeof value === 'string';
...
+    // Make sure we are only acting on strings or numbers
+    isStringOrNumber(value)
+        ? value
+              .toString()
+              .toLowerCase()
+              .replace(/[^0-9a-zA-Z]+/g, '_')
+        : '';
+
+const isDataPlanSlug = (str: string): boolean => str === toSlug(str);
```
rmi22186, line 215:
```suggestion
const isDataPlanId = (str: string): boolean => str === toDataPlanId(str);
```

rmi22186 (changes requested):
This looks good overall. I think just renaming it would make most sense since a slug can be hyphenated

rmi22186 (changes requested):
I'd recommend also removing slugify from package.json, doing an npm install, and also commiting the package-lock.json

rmi22186 (approved):

PR #787, opened 2023-10-12
docs: Update broken links in comments

rmi22186:
## Instructions

 ## Summary
/developers/sdk/javascript was changed to /developers/sdk/web and this link update was missed in this file.

Note that http://docs.mparticle.com/developers/sdk/javascript/event-tracking#event-type does not work, but https://docs.mparticle.com/developers/sdk/web/event-tracking/#event-type does.

 ## Testing Plan
Tested link

 - Closes https://go.mparticle.com/work/SQDSDKS-5865


einsteinx2 (approved):
lgtm

PR #792, opened 2023-10-19
fix: Respect sideloaded kit filter dictionaries

rmi22186:
## Instructions

 ## Summary
Filtered dictionaries were added after supporting side loaded kits.  However, after I added the filtered dictionaries support to the sideloaded kit class, I never included it when configuring the kits in the initialization flow, so all events were sent regardless of what filters were added.  This fixes that.

 ## Testing Plan
Added local tests, tested in a local app.

 - Closes https://go.mparticle.com/work/SQDSDKS-5893

### DIFF
--- a/src/forwarders.js
+++ b/src/forwarders.js
@@ -629,61 +629,47 @@ export default function Forwarders(mpInstance, kitBlocker) {
         }
     };
 
-    // Sideloaded kits are not configured in the UI and do not have kit configurations
-    // They are automatically added to active forwarders.
+    // Unlike UI enabled kits, sideloaded kits are always added to active forwarders.
 
-    // TODO: Sideloading kits currently requires the use of a register method
+    // TODO: Sideloading kits currently require the use of a register method
     // which requires an object on which to be registered.
-    // In the future, when all kits are moved to the config rather than
+    // In the future, when all kits are moved to the mpConfig rather than
     // there being a separate process for MP configured kits and
     // sideloaded kits, this will need to be refactored.
-    this.processSideloadedKits = function(config) {
+    this.processSideloadedKits = function(mpConfig) {
         try {
-            if (Array.isArray(config.sideloadedKits)) {
-                const sideloadedKits = { kits: {} };
-                // First register each kit's constructor onto sideloadedKits,
-                // which is typed { kits: Dictionary<constructor> }.
-                // The constructors are keyed by the name of the kit.
-                config.sideloadedKits.forEach(function(sideloadedKit) {
+            if (Array.isArray(mpConfig.sideloadedKits)) {
+                const registeredSideloadedKits = { kits: {} };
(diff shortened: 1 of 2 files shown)


src/forwarders.js
```diff
@@ -629,43 +629,48 @@ export default function Forwarders(mpInstance, kitBlocker) {
...
-            if (Array.isArray(config.sideloadedKits)) {
-                const sideloadedKits = { kits: {} };
-                // First register each kit's constructor onto sideloadedKits,
-                // which is typed { kits: Dictionary<constructor> }.
-                // The constructors are keyed by the name of the kit.
-                config.sideloadedKits.forEach(function(sideloadedKit) {
+            if (Array.isArray(mpConfig.sideloadedKits)) {
+                const registeredSideloadedKits = { kits: {} };
+                const unregisteredSideloadedKits = mpConfig.sideloadedKits;
+                unregisteredSideloadedKits.forEach(function(unregisteredKit) {
```
alexs-mparticle, line 644:
```suggestion
                const unregisteredSideloadedKits = mpConfig.sideloadedKits;

                unregisteredSideloadedKits.forEach(function(unregisteredKit) {
```

src/forwarders.js
```diff
@@ -629,43 +629,48 @@ export default function Forwarders(mpInstance, kitBlocker) {
...
                     try {
-                        sideloadedKit.kitInstance.register(sideloadedKits);
+                        // Register each sideloaded kit, which adds a key of the sideloaded kit name
+                        // and a value of the sideloaded kit constructor.
+                        unregisteredKit.kitInstance.register(
+                            registeredSideloadedKits
+                        );
+                        let kitName = unregisteredKit.kitInstance.name;
```
alexs-mparticle, line 651: I don't see this being reassigned anywhere.

```suggestion
                        const kitName = unregisteredKit.kitInstance.name;
```

test/src/tests-forwarders.js
```diff
@@ -2738,6 +2738,209 @@ describe('forwarders', function() {
...
+                    const attrs = {
+                        testAttr1: 'foo',
+                        testAttr2: 'bar'
+                    }
+
+                    mParticle.logEvent('Test Event', mParticle.EventType.Navigation, attrs);
+
+                    window.SideloadedKit11.instance.receivedEvent.EventAttributes.should.have.property('testAttr2', 'bar');
+                    window.SideloadedKit11.instance.receivedEvent.EventAttributes.should.not.property('testAttr1');
+s
```
alexs-mparticle, line 2840: This stray `s` is breaking the build.

```suggestion
```

rmi22186: Whoops I saw that after the fact and didn't commit it...

alexs-mparticle (approved):

alexs-mparticle (approved):

PR #794, opened 2023-10-20
tests: Update tests to support latest fetch-mock version

rmi22186:
## Instructions

 ## Summary
As part of browser modernization, I am updating fetch mock, and sinon.  This PR is for fetch-mock.  Previously we were pulling it in via a karma config file which was setting it on the window.  This is not necessary now that we are importing it directly.  In my testing when transpiling, pulling `fetch-mock` directly did not work, I had to do `fetch-mock/esm/client`.  Might be a bug to figure out later, but since this is just for testing purposes and not in our code base, I don't think it's a big deal.

Changes include:
* `window.fetchMock` --> `fetchMock`
* `._calls` -> `calls()`
* `.calls = []` to clear all calls --> `.resetHistory`
* slight modifications to the `tests/utils.js` file because the payload changed between v7 and v9 of `fetch-mock`, 

 ## Testing Plan

 - Closes https://go.mparticle.com/work/SQDSDKS-5887


test/src/tests-batchUploader.ts
```diff
@@ -14,10 +14,12 @@ import { expect } from 'chai';
 import _BatchValidator from '../../src/mockBatchCreator';
 import Logger from '../../src/logger.js';
 import { event0, event1, event2, event3 } from '../fixtures/events';
+import fetchMock from 'fetch-mock/esm/client';
+
 declare global {
     interface Window {
         mParticle: MParticleWebSDK;
-        fetchMock: any;
+        // fetchMock: any;
```
alexs-mparticle, line 22: I think we can safely delete this.

```suggestion
```

test/src/tests-batchUploader.ts
```diff
@@ -157,19 +159,19 @@ describe('batch uploader', () => {
                     [actualBatch, eventlessBatch],
                     false
                 );
-
-                expect(window.fetchMock.calls().length).to.equal(1);
+                    debugger;
```
alexs-mparticle, line 162: Please remove the debugger.

```suggestion
```

test/src/tests-batchUploader.ts
```diff
@@ -1341,12 +1343,13 @@ describe('batch uploader', () => {
...
 
         afterEach(() => {
-            window.fetchMock.restore();
+            fetchMock.restore();
             sinon.restore();
             clock.restore();
         });
 
         it('should use custom v3 endpoint', function(done) {
+            // debugger;
```
alexs-mparticle, line 1352:
```suggestion
```

test/src/tests-persistence.ts
```diff
@@ -431,9 +432,9 @@ describe('migrations and persistence-related', () => {
     it('should save integration attributes properly on a page refresh', done => {
         mParticle.setIntegrationAttribute(128, { MCID: 'abcedfg' });
         mParticle.init(apiKey, mParticle.config);
-
+        debugger;
```
alexs-mparticle, line 435:
```suggestion
```

alexs-mparticle (changes requested):

rmi22186:
Looks like I had to rebase master and force push because of package-lock conflicts.    Relevant changes since you last looked are here https://github.com/mparticle/mparticle-web-sdk/compare/5a26facdc8a0e91ce179f6f4e6a8615c11982606...4049f578e0ecfa1c21e4cf7ede1447436a16fbc0

alexs-mparticle (changes requested):

alexs-mparticle (approved):

PR #795, opened 2023-10-25
test: Transpile tests to run older browsers in browserstack

rmi22186:
## Instructions

 ## Summary
Our testing framework needs to be transpiled down to ES5.  We currently test only on latest chrome/firefox, which supports all this syntax.  In order to test cross browser, the tests themselves have to be transpiled to be able to test our SDK. 

There are 2 rollup steps to get a transpiled test to work properly. First we use rollup and run it on test-main.js which includes all the references to each test/src file, it also imports sinon and fetch-mock where appropriate, and this rollup step creates a single distribution file of everything (`test/cross-browser-testing/CBT-tests.js`).  I was unable to get this file to transpile down to ES5 in a single step.  There was always an `undefined` error in the build when I tried to do it in one step.  However, when I added a second rollup step, which takes in `test/cross-browser-testing/CBT-tests.js` and outputs ``.  This file is then able to run properly on browserstack on older files.  There is probably more debugging that can be done to resolve this, but since it is just for testing, the return did not seem worth the additional time investment.

 ## Testing Plan
Chrome 49 runs successfully.  Chrome 48 and prior seem to fail in some tests where `setTimeout` is used.

 - Closes https://go.mparticle.com/work/SQDSDKS-5834


package.json
```diff
@@ -28,6 +28,7 @@
         "build:esm": "cross-env ENVIRONMENT=prod BUILD=esm rollup --config rollup.config.js",
         "build:stub": "cross-env ENVIRONMENT=prod BUILD=stub rollup --config rollup.config.js",
         "build:snippet": "uglifyjs snippet.js -m -o snippet.min.js",
+        "build:tests": "cross-env ENVIRONMENT=prod BUILD=tests rollup --config rollup.config.js",
```
alexs-mparticle, line 31: Not sure if you're missing an import or package somewhere. When I run this command after an `npm install` I get the following error:

```
TypeError: Cannot read properties of undefined (reading 'acorn')
```

test/cross-browser-testing/babel.config.js
```diff
@@ -0,0 +1,12 @@
+export default {
```
alexs-mparticle, line 1: Nitpick: Can we name this `babel.CBT.config.js` so that it is consistent name with the rollup config in the same file?

test/cross-browser-testing/browserstack.karma.config.js
```diff
@@ -28,39 +24,22 @@ if (DEBUG === 'true') {
 }
 
 const customLaunchers = {
-    // These 2 browsers pass because they are capable of running ES6 syntax
-    // Our tests currently have ES6 syntax in them.  They need to be transpiled
-    bs_firefox_mac_100: {
```
alexs-mparticle, line 33: It's not clear to me why we are removing the Firefox launcher. Are we only concerned about Chrome as an "older" browser?

alexs-mparticle (changes requested):
I have a few questions and nitpicks, but overall I think this looks solid. I may want to take a greater look at how we run tests so we can organize things, but I think that can wait until everything is wired together.

alexs-mparticle (changes requested):
I have a few questions and nitpicks, but overall I think this looks solid. I may want to take a greater look at how we run tests so we can organize things, but I think that can wait until everything is wired together.

package.json
rmi22186: I have the wrong command here.  Will update.

test/cross-browser-testing/browserstack.karma.config.js
rmi22186: There is another ticket in the initiative to determine what browsers to use - go.mparticle.com/work/SQDSDKS-5827.  FF would definitely make a comeback. 

There wasn't ever really an intention to remove the FIrefox one...I ended up just focusing on Chrome for some reason.

alexs-mparticle (approved):

package.json
```diff
@@ -138,6 +140,7 @@
         "webpack-cli": "^5.0.2"
     },
     "dependencies": {
-        "@babel/runtime": "^7.6.0"
+        "@babel/runtime": "^7.23.2",
+        "core-js": "^3.33.0"
```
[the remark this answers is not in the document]
rmi22186: No it doesn't.  All of the sites I read say to put it in dependencies because it is used by @babel/runtime.  However, since we aren't using it in production but only for tests, I suppose it is okay to put in dev-dependencies.  Moved it there and the test and prod builds look identical

PR #797, opened 2023-11-05
feat: Support direct url routing from config

rmi22186:
## Instructions

 ## Summary
This PR provides logic to incorporate the new `directURLRouting` feature flag.  The feature routes all external requests to the equivalent silo ie `jssdkcdn.us1.mparticle.com` as opposed to the default which is going through a generic fastly endpoint ie `jssdkcdn.mparticle.com`.  The logic is as follows:
* When `true`, set any custom endpoints passed to confirm, then set remaining endpoints to be the silo-specific url
* When `false`, set any custom endpoints passed to confirm, then set remaining endpoints to be the generic fastly URL

 
 ## Testing Plan
 * Added unit tests.
 * Also tested in a local app to see if silos were respected when feature flag is true, and if custom URLs took priority over thos still

 - Closes https://go.mparticle.com/work/SQDSDKS-5899

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -126,7 +126,7 @@ const Constants = {
         integrationDelayTimeout: 5000, // Milliseconds for forcing the integration delay to un-suspend event queueing due to integration partner errors
         maxCookieSize: 3000, // Number of bytes for cookie size to not exceed
         aliasMaxWindow: 90, // Max age of Alias request startTime, in days
-        uploadInterval: 0, // Maximum milliseconds in between batch uploads, below 500 will mean immediate upload
+        uploadInterval: '0', // Maximum milliseconds in between batch uploads, below 500 will mean immediate upload
     },
     DefaultUrls: {
         v1SecureServiceUrl: 'jssdks.mparticle.com/v1/JS/',
@@ -166,6 +166,7 @@ const Constants = {
         ReportBatching: 'reportBatching',
         EventBatchingIntervalMillis: 'eventBatchingIntervalMillis',
         OfflineStorage: 'offlineStorage',
+        DirectUrlRouting: 'directURLRouting',
     },
     DefaultInstance: 'default_instance',
     CCPAPurpose: 'data_sale_opt_out',
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -126,7 +126,7 @@ const Constants = {
         integrationDelayTimeout: 5000, // Milliseconds for forcing the integration delay to un-suspend event queueing due to integration partner errors
         maxCookieSize: 3000, // Number of bytes for cookie size to not exceed
         aliasMaxWindow: 90, // Max age of Alias request startTime, in days
-        uploadInterval: '0', // Maximum milliseconds in between batch uploads, below 500 will mean immediate upload.  This is a string because that is how the server returns it
+        uploadInterval: 0, // Maximum milliseconds in between batch uploads, below 500 will mean immediate upload.  This is a string because that is how the server returns it
     },
     DefaultBaseUrls: {
         v1SecureServiceUrl: 'jssdks.mparticle.com/v1/JS/',
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -126,7 +126,7 @@ const Constants = {
         integrationDelayTimeout: 5000, // Milliseconds for forcing the integration delay to un-suspend event queueing due to integration partner errors
         maxCookieSize: 3000, // Number of bytes for cookie size to not exceed
         aliasMaxWindow: 90, // Max age of Alias request startTime, in days
-        uploadInterval: '0', // Maximum milliseconds in between batch uploads, below 500 will mean immediate upload
+        uploadInterval: '0', // Maximum milliseconds in between batch uploads, below 500 will mean immediate upload.  This is a string because that is how the server returns it
     },
     DefaultUrls: {
         v1SecureServiceUrl: 'jssdks.mparticle.com/v1/JS/',
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -126,7 +126,7 @@ const Constants = {
         integrationDelayTimeout: 5000, // Milliseconds for forcing the integration delay to un-suspend event queueing due to integration partner errors
         maxCookieSize: 3000, // Number of bytes for cookie size to not exceed
         aliasMaxWindow: 90, // Max age of Alias request startTime, in days
-        uploadInterval: 0, // Maximum milliseconds in between batch uploads, below 500 will mean immediate upload.  This is a string because that is how the server returns it
+        uploadInterval: 0, // Maximum milliseconds in between batch uploads, below 500 will mean immediate upload.  The server returns this as a string, but we are using it as a number internally
     },
     DefaultBaseUrls: {
         v1SecureServiceUrl: 'jssdks.mparticle.com/v1/JS/',
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -128,7 +128,7 @@ const Constants = {
         aliasMaxWindow: 90, // Max age of Alias request startTime, in days
         uploadInterval: '0', // Maximum milliseconds in between batch uploads, below 500 will mean immediate upload.  This is a string because that is how the server returns it
     },
-    DefaultUrls: {
+    DefaultBaseUrls: {
         v1SecureServiceUrl: 'jssdks.mparticle.com/v1/JS/',
         v2SecureServiceUrl: 'jssdks.mparticle.com/v2/JS/',
         v3SecureServiceUrl: 'jssdks.mparticle.com/v3/JS/',
--- a/src/mockBatchCreator.ts
+++ b/src/mockBatchCreator.ts
@@ -27,7 +27,6 @@ export default class _BatchValidator {
                 createServiceUrl: mockFunction,
                 parseNumber: mockFunction,
                 isObject: mockFunction,
-                getFeatureFlag: mockFunction,
                 returnConvertedBoolean: mockFunction,
                 Validators: null,
             },
--- a/src/mockBatchCreator.ts
+++ b/src/mockBatchCreator.ts
@@ -27,6 +27,7 @@ export default class _BatchValidator {
                 createServiceUrl: mockFunction,
                 parseNumber: mockFunction,
                 isObject: mockFunction,
+                getFeatureFlag: mockFunction,
                 returnConvertedBoolean: mockFunction,
                 Validators: null,
             },
--- a/src/mp-instance.js
+++ b/src/mp-instance.js
@@ -1523,9 +1523,8 @@ function createKitBlocker(config, mpInstance) {
 
 function runPreConfigFetchInitialization(mpInstance, apiKey, config) {
     mpInstance.Logger = new Logger(config);
-    mpInstance._Store = new Store(config, mpInstance);
+    mpInstance._Store = new Store(config, mpInstance, apiKey);
     window.mParticle.Store = mpInstance._Store;
-    mpInstance._Store.devToken = apiKey || null;
     mpInstance.Logger.verbose(
         Messages.InformationMessages.StartingInitialization
     );
--- a/src/sdkRuntimeModels.ts
+++ b/src/sdkRuntimeModels.ts
@@ -246,7 +246,7 @@ export interface SDKHelpersApi {
     parseNumber?(value: string | number): number;
     generateUniqueId();
     generateHash?(value: string): string;
-    getFeatureFlag?(feature: string); // TODO: Feature Constants should be converted to enum
+    getFeatureFlag(feature: string); // TODO: Feature Constants should be converted to enum
     isDelayedByIntegration?(
         delayedIntegrations: Dictionary<boolean>,
         timeoutStart: number,
--- a/src/sdkRuntimeModels.ts
+++ b/src/sdkRuntimeModels.ts
@@ -246,7 +246,7 @@ export interface SDKHelpersApi {
     parseNumber?(value: string | number): number;
     generateUniqueId();
     generateHash?(value: string): string;
-    getFeatureFlag(feature: string); // TODO: Feature Constants should be converted to enum
+    getFeatureFlag?(feature: string); // TODO: Feature Constants should be converted to enum
     isDelayedByIntegration?(
         delayedIntegrations: Dictionary<boolean>,
         timeoutStart: number,
--- a/src/store.ts
+++ b/src/store.ts
@@ -497,6 +497,9 @@ export function processEndpoints(
         if (config.hasOwnProperty(endpointKey)) {
             endpoints[endpointKey] = config[endpointKey];
         } else {
+            if (endpointKey === 'configUrl') {
+                continue;
+            }
             var urlparts = endpoints[endpointKey].split('.');
 
             endpoints[endpointKey] = [
--- a/src/store.ts
+++ b/src/store.ts
@@ -111,6 +111,13 @@ interface WrapperSDKInfo {
     isInfoSet: boolean;
 }
 
+export interface IFeatureFlags {
+    reportBatching?: string;
+    eventBatchingIntervalMillis?: string;
+    offlineStorage?: string;
+    directURLRouting?: boolean;
+}
+
 // Temporary Interface until Store can be refactored as a class
 export interface IStore {
     isEnabled: boolean;
@@ -161,7 +168,7 @@ export default function Store(
     this: IStore,
     config: SDKInitConfig,
     mpInstance: MParticleWebSDK,
-    apiKey: string,
+    apiKey: string
 ) {
     const defaultStore: Partial<IStore> = {
         isEnabled: true,
@@ -242,7 +249,7 @@ export default function Store(
             this.SDKConfig.flags,
             apiKey
         );
-        console.log(endpoints);
+
         for (const endpointKey in endpoints) {
             this.SDKConfig[endpointKey] = endpoints[endpointKey];
         }
@@ -255,7 +262,8 @@ export default function Store(
 
         this.SDKConfig.kits = config.kits || {};
 
-        this.SDKConfig.sideloadedKits = config.sideloadedKits || [];
+        this.SDKConfig.sideloadedKits =
+            config.sideloadedKits || [];
 
         if (config.hasOwnProperty('isIOS')) {
             this.SDKConfig.isIOS = config.isIOS;
@@ -267,7 +275,8 @@ export default function Store(
         }
 
         if (config.hasOwnProperty('useCookieStorage')) {
-            this.SDKConfig.useCookieStorage = config.useCookieStorage;
+            this.SDKConfig.useCookieStorage =
+                config.useCookieStorage;
         } else {
             this.SDKConfig.useCookieStorage = false;
         }
@@ -275,7 +284,8 @@ export default function Store(
         if (config.hasOwnProperty('maxProducts')) {
             this.SDKConfig.maxProducts = config.maxProducts;
         } else {
-            this.SDKConfig.maxProducts = Constants.DefaultConfig.maxProducts;
+            this.SDKConfig.maxProducts =
+                Constants.DefaultConfig.maxProducts;
         }
 
         if (config.hasOwnProperty('maxCookieSize')) {
@@ -302,13 +312,19 @@ export default function Store(
         }
 
         if (config.hasOwnProperty('identifyRequest')) {
-            this.SDKConfig.identifyRequest = config.identifyRequest;
+            this.SDKConfig.identifyRequest =
+                config.identifyRequest;
         }
 
         if (config.hasOwnProperty('identityCallback')) {
             var callback = config.identityCallback;
-            if (mpInstance._Helpers.Validators.isFunction(callback)) {
-                this.SDKConfig.identityCallback = config.identityCallback;
+            if (
+                mpInstance._Helpers.Validators.isFunction(
+                    callback
+                )
+            ) {
+                this.SDKConfig.identityCallback =
+                    config.identityCallback;
             } else {
                 mpInstance.Logger.warning(
                     'The optional callback must be a function. You tried entering a(n) ' +
@@ -327,7 +343,8 @@ export default function Store(
         }
 
         if (config.hasOwnProperty('sessionTimeout')) {
-            this.SDKConfig.sessionTimeout = config.sessionTimeout;
+            this.SDKConfig.sessionTimeout =
+                config.sessionTimeout;
         }
 
         if (config.hasOwnProperty('dataPlan')) {
@@ -339,7 +356,8 @@ export default function Store(
             const dataPlan = config.dataPlan as DataPlanConfig;
             if (dataPlan.planId) {
                 if (isDataPlanSlug(dataPlan.planId)) {
-                    this.SDKConfig.dataPlan.PlanId = dataPlan.planId;
+                    this.SDKConfig.dataPlan.PlanId =
+                        dataPlan.planId;
                 } else {
                     mpInstance.Logger.error(
                         'Your data plan id must be a string and match the data plan slug format (i.e. under_case_slug)'
@@ -349,7 +367,8 @@ export default function Store(
 
             if (dataPlan.planVersion) {
                 if (isNumber(dataPlan.planVersion)) {
-                    this.SDKConfig.dataPlan.PlanVersion = dataPlan.planVersion;
+                    this.SDKConfig.dataPlan.PlanVersion =
+                        dataPlan.planVersion;
                 } else {
                     mpInstance.Logger.error(
                         'Your data plan version must be a number'
@@ -377,7 +396,8 @@ export default function Store(
         }
 
         if (config.hasOwnProperty('aliasMaxWindow')) {
-            this.SDKConfig.aliasMaxWindow = config.aliasMaxWindow;
+            this.SDKConfig.aliasMaxWindow =
+                config.aliasMaxWindow;
         } else {
             this.SDKConfig.aliasMaxWindow =
                 Constants.DefaultConfig.aliasMaxWindow;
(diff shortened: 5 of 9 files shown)


src/mp-instance.js
```diff
@@ -1523,9 +1523,8 @@ function createKitBlocker(config, mpInstance) {
 
 function runPreConfigFetchInitialization(mpInstance, apiKey, config) {
     mpInstance.Logger = new Logger(config);
-    mpInstance._Store = new Store(config, mpInstance);
+    mpInstance._Store = new Store(config, mpInstance, apiKey);
```
alexs-mparticle, line 1526: This feels like a big change. What's the rationale here?

src/store.ts
```diff
@@ -227,28 +237,14 @@ export default function Store(
...
-            this.SDKConfig.identityUrl = config.identityUrl;
-        }
-
-        if (config.hasOwnProperty('aliasUrl')) {
-            this.SDKConfig.aliasUrl = config.aliasUrl;
-        }
-
-        if (config.hasOwnProperty('configUrl')) {
-            this.SDKConfig.configUrl = config.configUrl;
+        const endpoints: Dictionary = processEndpoints(
```
alexs-mparticle, line 240: I think this is more clear since `Dictionary` defaults to `any`.

```suggestion
        const endpoints: Dictionary<string> = processEndpoints(
```

src/store.ts
```diff
@@ -227,28 +237,14 @@ export default function Store(
...
+            config,
+            this.SDKConfig.flags,
+            apiKey
+        );
+        console.log(endpoints);
```
alexs-mparticle, line 245: I think you forgot to remove this.

```suggestion
        console.log(endpoints);
```

src/store.ts
```diff
@@ -413,32 +409,115 @@ export default function Store(
...
-        ) {
-            this.SDKConfig.flags[Constants.FeatureFlags.ReportBatching] = false;
-        }
-        if (
-            !this.SDKConfig.flags.hasOwnProperty(
-                Constants.FeatureFlags.OfflineStorage
-            )
-        ) {
-            this.SDKConfig.flags[Constants.FeatureFlags.OfflineStorage] = 0;
+export interface IFeatureFlags {
```
alexs-mparticle, line 415: Can you please move this to the top of the file? I think all interfaces should live up there.

src/store.ts
```diff
@@ -413,32 +409,115 @@ export default function Store(
...
+    flags[OfflineStorage] = config.flags[OfflineStorage] || '0';
+    flags[DirectUrlRouting] = config.flags[DirectUrlRouting] === 'True';
+
+    return flags;
+}
+
+export function processEndpoints(
+    config: SDKInitConfig,
+    flags: IFeatureFlags,
+    apiKey: string
```
alexs-mparticle, line 452: If you're checking for the presence of `apiKey` below, you should either make this optional parameter or give it a default value of `{}`.

src/mockBatchCreator.ts
```diff
@@ -27,6 +27,7 @@ export default class _BatchValidator {
                 createServiceUrl: mockFunction,
                 parseNumber: mockFunction,
                 isObject: mockFunction,
+                getFeatureFlag: mockFunction,
```
alexs-mparticle, line 30: You can very likely remove this change and just make `getFeatureFlag` optional.

src/sdkRuntimeModels.ts
```diff
@@ -246,7 +246,7 @@ export interface SDKHelpersApi {
     parseNumber?(value: string | number): number;
     generateUniqueId();
     generateHash?(value: string): string;
-    getFeatureFlag?(feature: string); // TODO: Feature Constants should be converted to enum
+    getFeatureFlag(feature: string); // TODO: Feature Constants should be converted to enum
```
alexs-mparticle, line 249: Revert this change as per our slack conversation.

src/store.ts
```diff
@@ -413,32 +409,115 @@ export default function Store(
...
-        ) {
-            this.SDKConfig.flags[Constants.FeatureFlags.OfflineStorage] = 0;
+export interface IFeatureFlags {
+    reportBatching?: string;
+    eventBatchingIntervalMillis?: string;
+    offlineStorage?: string;
+    directURLRouting?: boolean;
+}
+
+export function processFlags(
```
alexs-mparticle, line 422: This might be a topic of conversation for a future revision, but I honestly think we should remove FeatureFlags and Endpoints from the Store and make them their own classes, or just move this processing logic outside of the file to make it smaller. Just leaving this comment as a reminder that we should chat about it.

src/store.ts
```diff
@@ -413,32 +409,115 @@ export default function Store(
...
+    flags[DirectUrlRouting] = config.flags[DirectUrlRouting] === 'True';
+
+    return flags;
+}
+
+export function processEndpoints(
+    config: SDKInitConfig,
+    flags: IFeatureFlags,
+    apiKey: string
+): Dictionary {
```
alexs-mparticle, line 453:
```suggestion
): Dictionary<string> {
```

src/store.ts
```diff
@@ -413,32 +409,115 @@ export default function Store(
...
+    // an API key is not present in a webview only mode. In this case, no endpoints are needed
+    if (!apiKey) {
+        return {};
+    }
+
+    // Set default endpoints
+    let endpoints: Dictionary = JSON.parse(
+        JSON.stringify(Constants.DefaultUrls)
```
alexs-mparticle, line 461: Might be out of scope, but can we rename this to `DefaultEndpoints`?

src/store.ts
```diff
@@ -413,32 +409,115 @@ export default function Store(
...
+    // When direct URL routing is false, update endpoints based custom urls
+    // passed to the config
+    if (!flags.directURLRouting) {
+        return processNonDirectEndpoints(endpoints, config);
+    } else {
+        return processDirectUrlEndpoints(endpoints, config, apiKey);
+    }
+}
+
+function processNonDirectEndpoints(
```
alexs-mparticle, line 473: Why can't we call this `processConfiguredEndpoints` or `processCustomEndpoints`?

src/store.ts
```diff
@@ -413,32 +409,115 @@ export default function Store(
...
+    // silo and a hyphen (ex. "us1-", "us2-", "eu1-").  us1 was the first silo,
+    // and before other silos existed, there were no prefixes and all apiKeys
+    // were us1. As such, if we split on a '-' and the resulting array length
+    // is 1, then it is an older APIkey that should route to us1.
+    const splitKey: Array<string> = apiKey.split('-');
+    let routingPrefix: string;
+
+    // when splitKey.length is greater than 1, then splitKey[0] will be
+    // us1, us2, eu1, au1, or st1, etc as new silos are added
+    routingPrefix = splitKey.length <= 1 ? 'us1' : splitKey[0];
```
alexs-mparticle, line 501: I don't think splitting this up makes it more or less readable.

```suggestion
    // when splitKey.length is greater than 1, then splitKey[0] will be
    // us1, us2, eu1, au1, or st1, etc as new silos are added
    const routingPrefix: string = splitKey.length <= 1 ? 'us1' : splitKey[0];
```

alexs-mparticle: Also, your comment doesn't explain why are we defaulting to `us1`. I would add a follow up comment explaining this.

As an alternative, make `us1` a constant called `DEFAULT_ROUTE` or `DEFAULT_SILO` and then we don't need that comment about `us1` being the default.

src/store.ts
```diff
@@ -413,32 +409,115 @@ export default function Store(
...
+    flags: IFeatureFlags,
+    apiKey: string
+): Dictionary {
+    // an API key is not present in a webview only mode. In this case, no endpoints are needed
+    if (!apiKey) {
+        return {};
+    }
+
+    // Set default endpoints
+    let endpoints: Dictionary = JSON.parse(
```
alexs-mparticle, line 460: You're never reassigning, so this can be `const`.

```suggestion
    const endpoints: Dictionary = JSON.parse(
```

src/store.ts
```diff
@@ -413,32 +409,115 @@ export default function Store(
...
+        JSON.stringify(Constants.DefaultUrls)
+    );
+
+    // When direct URL routing is false, update endpoints based custom urls
+    // passed to the config
+    if (!flags.directURLRouting) {
+        return processNonDirectEndpoints(endpoints, config);
```
alexs-mparticle, line 467: This is more of a pet peeve for me, but I don't like that these methods are mutating the endpoints object you're passing. I personally think we should have the process methods should return a new endpoints object, and then we merge it ourselves. I'm open to chat about it.

src/store.ts
```diff
@@ -413,32 +409,115 @@ export default function Store(
...
+    // were us1. As such, if we split on a '-' and the resulting array length
+    // is 1, then it is an older APIkey that should route to us1.
+    const splitKey: Array<string> = apiKey.split('-');
+    let routingPrefix: string;
+
+    // when splitKey.length is greater than 1, then splitKey[0] will be
+    // us1, us2, eu1, au1, or st1, etc as new silos are added
+    routingPrefix = splitKey.length <= 1 ? 'us1' : splitKey[0];
+
+    for (let endpointKey in endpoints) {
```
alexs-mparticle, line 503: As mentioned in my previous comment, if you don't mutate the endpoints within the process method, you can simplify this method and move this logic outside of the process method, which might make things more readable.

test/src/tests-main.js
```diff
@@ -19,7 +19,7 @@ beforeEach(function() {
         requestConfig: false,
         isDevelopmentMode: false,
         flags: {
-            eventBatchingIntervalMillis: 0,
+            eventBatchingIntervalMillis: "0",
```
alexs-mparticle, line 22: Why does this need to be a string now?

alexs-mparticle (changes requested):

src/constants.ts
```diff
@@ -126,7 +126,7 @@ const Constants = {
         integrationDelayTimeout: 5000, // Milliseconds for forcing the integration delay to un-suspend event queueing due to integration partner errors
         maxCookieSize: 3000, // Number of bytes for cookie size to not exceed
         aliasMaxWindow: 90, // Max age of Alias request startTime, in days
-        uploadInterval: 0, // Maximum milliseconds in between batch uploads, below 500 will mean immediate upload
+        uploadInterval: '0', // Maximum milliseconds in between batch uploads, below 500 will mean immediate upload
```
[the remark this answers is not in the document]
rmi22186: We can discuss.  I would prefer to keep this as a number too, but unfortunately the backend returns it as a string, so I wanted to be consistent so we don't have to do any sort of "or" clauses in the code.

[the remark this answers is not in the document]
rmi22186: image

test/src/tests-main.js
rmi22186: It's returned this way from the server, so we just need to come to a consensus for when it's turned into a number.

src/mp-instance.js
rmi22186: Also, since line 1528 previously was just setting the apiKey on the store, it seemed like it should always have just been passed into the Store.

src/store.ts
rmi22186: I thought it was explained in the comment from 490-500.  I'll combine the comments if that makes it more understandable.

src/store.ts
rmi22186: I suppose they are not really `endpoints` and the variable I named it is a bit misleading, because the endpoint is something like jssdks.mparticle.com/v1/JS/[APIKEY]/events
and these are just the beginning portions of the endpoints, and the full endpoint gets created at runtime via.

it's more like an...endpoint...prefix?

src/store.ts
```diff
@@ -413,32 +416,111 @@ export default function Store(
...
+    if (flags.directURLRouting) {
+        return processDirectBaseUrls(config, apiKey);
+    } else {
+        return processCustomBaseUrls(config);
     }
 }
+
+function processCustomBaseUrls(config: SDKInitConfig): Dictionary<string> {
+    let defaultBaseUrls: Dictionary<string> = Constants.DefaultBaseUrls;
+    let newBaseUrls: Dictionary<string> = {};
```
alexs-mparticle, line 473:
```suggestion
    const defaultBaseUrls: Dictionary<string> = Constants.DefaultBaseUrls;
    const newBaseUrls: Dictionary<string> = {};
```

src/store.ts
```diff
@@ -413,32 +416,111 @@ export default function Store(
...
+
+    return newBaseUrls;
+}
+
+function processDirectBaseUrls(
+    config: SDKInitConfig,
+    apiKey: string
+): Dictionary {
+    let defaultBaseUrls = Constants.DefaultBaseUrls;
+    let directBaseUrls: Dictionary<string> = {};
```
alexs-mparticle, line 489:
```suggestion
    const defaultBaseUrls = Constants.DefaultBaseUrls;
    const directBaseUrls: Dictionary<string> = {};
```

test/src/tests-core-sdk.js
```diff
@@ -1134,4 +1133,172 @@ describe('core SDK', function() {
 
         done();
     });
+
+    describe('pod feature flag', function() {
+        const endpoints = Constants.DefaultBaseUrls;
+        // set up URLs object for each silo
+        let URLs = {
```
alexs-mparticle, line 1140:
```suggestion
        const URLs = {
```

test/src/tests-core-sdk.js
```diff
@@ -1134,4 +1133,172 @@ describe('core SDK', function() {
...
+        beforeEach(function() {
+            window.mParticle.config.flags = {
+                directURLRouting: 'True'
+            };
+        });
+
+        it('should use US1 endpoints for apiKeys that do not start with a prefix', function(done) {
+            let silo = 'us1';
+            let apiKey = 'noSiloPrefixApiKey';
+            let eventsEndpoint = `https://${URLs[silo].v3SecureServiceUrl}${apiKey}/events`;
```
alexs-mparticle, line 1172:
```suggestion
            const silo = 'us1';
            const apiKey = 'noSiloPrefixApiKey';
            const eventsEndpoint = `https://${URLs[silo].v3SecureServiceUrl}${apiKey}/events`;
```

test/src/tests-core-sdk.js
```diff
@@ -1134,4 +1133,172 @@ describe('core SDK', function() {
...
+            mParticle.getInstance()._Store.SDKConfig.v2SecureServiceUrl.should.equal(URLs[silo].v2SecureServiceUrl);
+            mParticle.getInstance()._Store.SDKConfig.v3SecureServiceUrl.should.equal(URLs[silo].v3SecureServiceUrl);
+
+            done();
+        });
+
+        it('should use US1 endpoints for apiKeys with prefix `us1`', function(done) {
+            let silo = 'us1';
+            let apiKey = 'us1-apiKey';
+            let eventsEndpoint = `https://${URLs.us1.v3SecureServiceUrl}${apiKey}/events`;
```
alexs-mparticle, line 1190:
```suggestion
            const silo = 'us1';
            const apiKey = 'us1-apiKey';
            const eventsEndpoint = `https://${URLs.us1.v3SecureServiceUrl}${apiKey}/events`;
```

test/src/tests-core-sdk.js
```diff
@@ -1134,4 +1133,172 @@ describe('core SDK', function() {
...
+            mParticle.getInstance()._Store.SDKConfig.v2SecureServiceUrl.should.equal(URLs[silo].v2SecureServiceUrl);
+            mParticle.getInstance()._Store.SDKConfig.v3SecureServiceUrl.should.equal(URLs[silo].v3SecureServiceUrl);
+
+            done();
+        });
+
+        it('should use US2 endpoints for apiKeys with prefix `us2`', function(done) {
+            let silo = 'us2';
+            let apiKey = 'us2-apiKey';
+            let eventsEndpoint = `https://${URLs[silo].v3SecureServiceUrl}${apiKey}/events`;
```
alexs-mparticle, line 1208:
```suggestion
            const silo = 'us2';
            const apiKey = 'us2-apiKey';
            const eventsEndpoint = `https://${URLs[silo].v3SecureServiceUrl}${apiKey}/events`;
```

test/src/tests-core-sdk.js
```diff
@@ -1134,4 +1133,172 @@ describe('core SDK', function() {
...
+            mParticle.getInstance()._Store.SDKConfig.v2SecureServiceUrl.should.equal(URLs[silo].v2SecureServiceUrl);
+            mParticle.getInstance()._Store.SDKConfig.v3SecureServiceUrl.should.equal(URLs[silo].v3SecureServiceUrl);
+
+            done();
+        });
+
+        it('should use EU1 endpoints for apiKeys with prefix `eu1`', function(done) {
+            let silo = 'eu1';
+            let apiKey = 'eu1-apiKey';
+            let eventsEndpoint = `https://${URLs[silo].v3SecureServiceUrl}${apiKey}/events`;
```
alexs-mparticle, line 1226:
```suggestion
            const silo = 'eu1';
            const apiKey = 'eu1-apiKey';
            const eventsEndpoint = `https://${URLs[silo].v3SecureServiceUrl}${apiKey}/events`;
```

test/src/tests-core-sdk.js
```diff
@@ -1134,4 +1133,172 @@ describe('core SDK', function() {
...
+            mParticle.getInstance()._Store.SDKConfig.v2SecureServiceUrl.should.equal(URLs[silo].v2SecureServiceUrl);
+            mParticle.getInstance()._Store.SDKConfig.v3SecureServiceUrl.should.equal(URLs[silo].v3SecureServiceUrl);
+
+            done();
+        });
+
+        it('should use AU1 endpoints for apiKeys with prefix `au1`', function(done) {
+            let silo = 'au1';
+            let apiKey = 'au1-apiKey';
+            let eventsEndpoint = `https://${URLs[silo].v3SecureServiceUrl}${apiKey}/events`;
```
alexs-mparticle, line 1244:
```suggestion
            const silo = 'au1';
            const apiKey = 'au1-apiKey';
            const eventsEndpoint = `https://${URLs[silo].v3SecureServiceUrl}${apiKey}/events`;
```

test/src/tests-core-sdk.js
```diff
@@ -1134,4 +1133,172 @@ describe('core SDK', function() {
...
+            mParticle.getInstance()._Store.SDKConfig.v2SecureServiceUrl.should.equal(URLs[silo].v2SecureServiceUrl);
+            mParticle.getInstance()._Store.SDKConfig.v3SecureServiceUrl.should.equal(URLs[silo].v3SecureServiceUrl);
+
+            done();
+        });
+
+        it('should use AU1 endpoints for apiKeys with prefix `st1`', function(done) {
+            let silo = 'st1';
+            let apiKey = 'st1-apiKey';
+            let eventsEndpoint = `https://${URLs[silo].v3SecureServiceUrl}${apiKey}/events`;
```
alexs-mparticle, line 1262:
```suggestion
            const silo = 'st1';
            const apiKey = 'st1-apiKey';
            const eventsEndpoint = `https://${URLs[silo].v3SecureServiceUrl}${apiKey}/events`;
```

test/src/tests-core-sdk.js
```diff
@@ -1134,4 +1133,172 @@ describe('core SDK', function() {
...
+            window.mParticle.config.v3SecureServiceUrl = 'testtesttest-custom-v3secureserviceurl/v3/JS/';
+            window.mParticle.config.configUrl ='foo-custom-configUrl/v2/JS/';
+            window.mParticle.config.identityUrl = 'custom-identityUrl/';
+            window.mParticle.config.aliasUrl = 'custom-aliasUrl/';
+
+            const {configUrl, v3SecureServiceUrl, identityUrl, aliasUrl} = window.mParticle.config
+
+            let silo = 'us1';
+            let apiKey = 'noSiloPrefixApiKey';
+            let eventsEndpoint = `https://${v3SecureServiceUrl}${apiKey}/events`;
```
alexs-mparticle, line 1287:
```suggestion
            const silo = 'us1';
            const apiKey = 'noSiloPrefixApiKey';
            const eventsEndpoint = `https://${v3SecureServiceUrl}${apiKey}/events`;
```

test/src/tests-store.ts
```diff
@@ -43,7 +46,7 @@ describe('Store', () => {
         expect(store.isFirstRun, 'isFirstRun').to.eq(null);
         expect(store.clientId, 'clientId').to.eq(null);
         expect(store.deviceId, 'deviceId').to.eq(null);
-        expect(store.devToken, 'devToken').to.eq(null);
+        expect(store.devToken, 'devToken').to.eq(apiKey);
```
alexs-mparticle, line 49: If `apiKey` is optional, this should be null and the following test on line 87 tests for the presence of `apiKey`.

alexs-mparticle (changes requested):

src/constants.ts
```diff
@@ -126,9 +126,9 @@ const Constants = {
         integrationDelayTimeout: 5000, // Milliseconds for forcing the integration delay to un-suspend event queueing due to integration partner errors
         maxCookieSize: 3000, // Number of bytes for cookie size to not exceed
         aliasMaxWindow: 90, // Max age of Alias request startTime, in days
-        uploadInterval: 0, // Maximum milliseconds in between batch uploads, below 500 will mean immediate upload
+        uploadInterval: '0', // Maximum milliseconds in between batch uploads, below 500 will mean immediate upload.  This is a string because that is how the server returns it
```
alexs-mparticle, line 129: As per our conversation on Zoom, this should be a number. All types used within the SDK should conform to JS, even if the server returns only strings.

src/store.ts
```diff
@@ -111,6 +111,13 @@ interface WrapperSDKInfo {
     isInfoSet: boolean;
 }
 
+export interface IFeatureFlags {
+    reportBatching?: string;
+    eventBatchingIntervalMillis?: string;
+    offlineStorage?: string;
+    directURLRouting?: boolean;
+}
```
alexs-mparticle, line 119: As per our conversation on Zoom, this should be a number. All types used within the SDK should conform to JS, even if the server returns only strings.

alexs-mparticle: Create a TODO for `offlineStorage`

src/store.ts
```diff
@@ -160,7 +167,8 @@ export interface IStore {
 export default function Store(
     this: IStore,
     config: SDKInitConfig,
-    mpInstance: MParticleWebSDK
+    mpInstance: MParticleWebSDK,
+    apiKey: string
```
alexs-mparticle, line 171: Confirm: Is this optional or required?

alexs-mparticle (changes requested):

src/constants.ts
```diff
@@ -126,9 +126,9 @@ const Constants = {
         integrationDelayTimeout: 5000, // Milliseconds for forcing the integration delay to un-suspend event queueing due to integration partner errors
         maxCookieSize: 3000, // Number of bytes for cookie size to not exceed
         aliasMaxWindow: 90, // Max age of Alias request startTime, in days
-        uploadInterval: 0, // Maximum milliseconds in between batch uploads, below 500 will mean immediate upload
+        uploadInterval: 0, // Maximum milliseconds in between batch uploads, below 500 will mean immediate upload.  This is a string because that is how the server returns it
```
alexs-mparticle, line 129:
```suggestion
        uploadInterval: 0, // Maximum milliseconds in between batch uploads, below 500 will mean immediate upload.  The server returns this as a string, but we are using it as a number internally
```

src/store.ts
```diff
@@ -111,6 +111,14 @@ interface WrapperSDKInfo {
     isInfoSet: boolean;
 }
 
+// TODO: https://go.mparticle.com/SQDSDKS-5954
```
alexs-mparticle, line 114:
```suggestion
// TODO: https://go.mparticle.com/work/SQDSDKS-5954
```

alexs-mparticle (changes requested):
I think this looks great. Just some minor cleanup.

src/store.ts
einsteinx2: @rmi22186 I used the term "subdomain" in the iOS code assuming we're talking about just the `jssdks` part here

test/src/tests-store.ts
```diff
@@ -185,10 +184,158 @@ describe('Store', () => {
...
+                    'apikey'
+                );
+
+                expect(result).to.deep.equal(baseUrls)
+            });
+
+            it('should return non-default baseUrls for custom baseUrls that are passed', () => {
+                const config = {
+                    v3SecureServiceUrl: 'testtesttest-custom-v3secureserviceurl/v3/JS/',
+                    configUrl: 'foo-custom-configUrl/v2/JS/',
```
einsteinx2, line 261: Here and below, it would be better to use more realistic URLs, e.g. "foo.customer.com"

test/src/tests-core-sdk.js
```diff
@@ -1134,4 +1133,172 @@ describe('core SDK', function() {
...
+            mParticle.getInstance()._Store.SDKConfig.configUrl.should.equal(Constants.DefaultBaseUrls.configUrl);
+            mParticle.getInstance()._Store.SDKConfig.identityUrl.should.equal(URLs[silo].identityUrl);
+            mParticle.getInstance()._Store.SDKConfig.v1SecureServiceUrl.should.equal(URLs[silo].v1SecureServiceUrl);
+            mParticle.getInstance()._Store.SDKConfig.v2SecureServiceUrl.should.equal(URLs[silo].v2SecureServiceUrl);
+            mParticle.getInstance()._Store.SDKConfig.v3SecureServiceUrl.should.equal(URLs[silo].v3SecureServiceUrl);
+
+            done();
+        });
+
+        it('should use AU1 endpoints for apiKeys with prefix `st1`', function(done) {
```
einsteinx2, line 1259:
```suggestion
        it('should use ST1 endpoints for apiKeys with prefix `st1`', function(done) {
```

test/src/tests-core-sdk.js
```diff
@@ -1134,4 +1133,172 @@ describe('core SDK', function() {
 
         done();
     });
+
+    describe('pod feature flag', function() {
+        const endpoints = Constants.DefaultBaseUrls;
+        // set up URLs object for each silo
+        const URLs = {
```
einsteinx2, line 1140: It would be good to also test a non-existent pod, say `sa1` or something

src/store.ts
```diff
@@ -413,32 +417,112 @@ export default function Store(
...
+    // is 1, then it is an older APIkey that should route to us1.
+    // When splitKey.length is greater than 1, then splitKey[0] will be
+    // us1, us2, eu1, au1, or st1, etc as new silos are added
+    const DEFAULT_SILO = 'us1';
+    const splitKey: Array<string> = apiKey.split('-');
+    const routingPrefix: string =
+        splitKey.length <= 1 ? DEFAULT_SILO : splitKey[0];
+
+    for (let baseUrlKey in defaultBaseUrls) {
+        // Any custom endpoints passed to config will take priority over direct
```
einsteinx2, line 506: Suggestion, change config to mpConfig

src/store.ts
```diff
@@ -413,32 +417,112 @@ export default function Store(
...
+        // mapping to the silo
+        if (baseUrlKey === 'configUrl') {
+            directBaseUrls[baseUrlKey] =
+                config[baseUrlKey] || defaultBaseUrls[baseUrlKey];
+            continue;
+        }
+
+        if (config.hasOwnProperty(baseUrlKey)) {
```
einsteinx2, line 514: maybe add comment above this line to say that a CNAME overrides direct url routing

einsteinx2 (changes requested):

src/store.ts
```diff
@@ -111,6 +111,14 @@ interface WrapperSDKInfo {
     isInfoSet: boolean;
 }
 
+// TODO: https://go.mparticle.com/work/SQDSDKS-5954
```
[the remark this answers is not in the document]
rmi22186: It's something we've done on Web SDK as a reminder for sustained engineering work. I do see a reason to not, as it seems like there is "unfinished" work, but our TODOs have mostely been about future optimizations.  For now I can just remove the TODO but keep the link as the reminder.

einsteinx2 (approved):
looks good!

PR #798, opened 2023-11-09
ci: Add Browserstack GIthub workflow

rmi22186:
## Instructions

 ## Summary
This PR sets up Browserstack tests in a Github Action. It follows the example workflow provided by the Browserstack docs here - https://www.browserstack.com/docs/automate/selenium/github-actions#sample-github-workflow-showing-a-browserstack-test

We originally discussed having it only be on release. However, I think it makes sense for it to be on push/pr also because we wouldn't want something to pass our non-CBT tests, merge them in, then upon release see that something is broken.  Given that it is its own workflow, and our workflow minutes are unlimited for public repos, I see no issue with this.  Note that this successfully runs [here](https://github.com/mParticle/mparticle-web-sdk/actions/runs/6818006161/job/18542716276) (although there are some failed tests, Chrome 48 tests here fail a few in the GHA, which matches the locally run Browserstack npm task, and Chrome 49 simply has a failed flaky test):

 ## Testing Plan
 - [] Was this tested locally? If not, explain why.
Not locally, but the action successfully runs here - https://github.com/mParticle/mparticle-web-sdk/actions/runs/6818006161/job/18542716276

 - Closes https://go.mparticle.com/work/SQDSDKS-5792

### DIFF
--- a/.github/workflows/cross-browser-testing.yml
+++ b/.github/workflows/cross-browser-testing.yml
@@ -1,0 +1,40 @@
+name: 'BrowserStack Test'
+on: [push, pull_request, workflow_dispatch]
+
+jobs:
+  ubuntu-job:
+    name: 'BrowserStack Test on Ubuntu'
+    runs-on: ubuntu-latest  # Can be self-hosted runner also
+    steps:
+
+      - name: 'BrowserStack Env Setup'  # Invokes the setup-env action
+        uses: browserstack/github-actions/setup-env@master
+        with:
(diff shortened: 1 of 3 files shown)


alexs-mparticle (approved):

alexs-mparticle (commented):
Code looks good, but you should make sure your tests are passing before you merge it in.

alexs-mparticle (approved):

PR #801, opened 2023-11-16
feat: Support direct url routing from config (#797)

rmi22186:
Duplicating this PR - https://github.com/mParticle/mparticle-web-sdk/pull/797 which was already approved and merged into development, but undone due to blackout requirements.

## Instructions

 ## Summary
This PR provides logic to incorporate the new `directURLRouting` feature flag.  The feature routes all external requests to the equivalent silo ie `jssdkcdn.us1.mparticle.com` as opposed to the default which is going through a generic fastly endpoint ie `jssdkcdn.mparticle.com`.  The logic is as follows:
* When `true`, set any custom endpoints passed to confirm, then set remaining endpoints to be the silo-specific url
* When `false`, set any custom endpoints passed to confirm, then set remaining endpoints to be the generic fastly URL
 
 ## Testing Plan
 * Added unit tests.
 * Also tested in a local app to see if silos were respected when feature flag is true, and if custom URLs took priority over thos still

 - Closes https://go.mparticle.com/work/SQDSDKS-5899

### DIFF
--- a/src/apiClient.ts
+++ b/src/apiClient.ts
@@ -3,7 +3,7 @@ import Types from './types';
 import { BatchUploader } from './batchUploader';
 import { MParticleUser, MParticleWebSDK, SDKEvent } from './sdkRuntimeModels';
 import KitBlocker from './kitBlocking';
-import { Dictionary, isEmpty } from './utils';
+import { Dictionary, getRampNumber, isEmpty, parseNumber } from './utils';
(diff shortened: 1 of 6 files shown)


samdozor (changes requested):
looks good but blocking merge until we are ready to use phased rollout for it.

rmi22186:
@samdozor 🙏

PR #806, opened 2023-11-20
ci: Add pre-release action to prepare release branch

alexs-mparticle:
## Instructions

 ## Summary
  - Builds bundle and prepares for CDN Release without deploying to master
  - Also pushes to NPM

 ## Testing Plan
 - N/A This is a CI/CD Update

 - Closes https://go.mparticle.com/work/SQDSDKS-5966

### DIFF
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -1,4 +1,4 @@
-name: Prepare CDN Pre-Release and NPM Publish
+name: Release - Step 1: Prepare CDN Pre-Release, Publish NPM
 
 on:
     workflow_dispatch:
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -1,0 +1,312 @@
+name: Prepare CDN Pre-Release and NPM Publish
+
+on:
+    workflow_dispatch:
+        inputs:
+            dryRun:
+                description: 'Do a dry run to preview instead of a real release [true/false]'
+                required: true
+                type: boolean
+                default: true
+
+jobs:
+    # SDK release is done from master branch.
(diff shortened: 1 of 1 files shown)


.github/workflows/pre-release.yml
```diff
@@ -0,0 +1,312 @@
+name: Prepare CDN Pre-Release and NPM Publish
```
rmi22186, line 1: // Small grammar nit.  Sounds like you are preparing the NPM publish, as opposed to actually doing it
```suggestion
name: Release - Step 1: Prepare CDN Pre-Release, Publish NPM
```

.github/workflows/pre-release.yml
```diff
@@ -0,0 +1,312 @@
...
+                  npx semantic-release --dry-run
+
+            # NPM Publish happens here
+            - name: Release
+              if: ${{ github.event.inputs.dryRun == 'false'}}
+              run: |
+                  npx semantic-release
+
+            - name: Get Package Version Number
+              uses: dutscher/read-package-json-endpoint-actions@v1.33.7
```
rmi22186, line 294: nit - 

When searching for this action, I saw a market place action here which had a lot more stars (123 vs 4), https://github.com/marketplace/actions/get-current-package-version.  The behavior is fairly simple, but figured star rating is a good indicator of which to choose when everything else works the same.

rmi22186 (changes requested):
Looks good.  Small nits.

Also - I wonder if for all our Github Acitons, we should include a comment at the top that summarizes what is happening in the workflow.  I think the Lucid chart is in a good spot, so I'd make a copy and remove the green highlighting.

rmi22186 (approved):

PR #805, opened 2023-11-20
ci: Add Pre-release action to push to Release Order Branch

alexs-mparticle:
## Instructions
 1. PR target branch should be against `development`
 2. PR title name should follow this format: https://github.com/mParticle/mparticle-workflows/blob/main/.github/workflows/pr-title-check.yml
 3. PR branch prefix should follow this format: https://github.com/mParticle/mparticle-workflows/blob/main/.github/workflows/pr-branch-check-name.yml

 ## Summary
 - Pushes a release branch to a release order branch

 ## Testing Plan
 - [ ] Was this tested locally? If not, explain why.
 - N/A - This is a CI/CD Update

 ## Reference Issue (For mParticle employees only.  Ignore if you are an outside contributor)
 - Closes https://go.mparticle.com/work/SQDSDKS-5964

PR #804, opened 2023-11-20
ci: Add Pre-release action to synchronize branches

alexs-mparticle:
## Instructions
 1. PR target branch should be against `development`
 2. PR title name should follow this format: https://github.com/mParticle/mparticle-workflows/blob/main/.github/workflows/pr-title-check.yml
 3. PR branch prefix should follow this format: https://github.com/mParticle/mparticle-workflows/blob/main/.github/workflows/pr-branch-check-name.yml

 ## Summary
 - This action synchronizes a release version of the web sdk between all necessary branches

 ## Testing Plan
 - [ ] Was this tested locally? If not, explain why.
 - N/A This is a CI/CD Update

 ## Reference Issue (For mParticle employees only.  Ignore if you are an outside contributor)
 - Closes https://go.mparticle.com/work/SQDSDKS-5965

PR #807, opened 2023-11-21
ci: Revise release process for a phased release of the Web SDK

alexs-mparticle:
## Instructions

 ## Summary
 - Separates bundle building and Semantic Release from CDN Deployment
 - Adds processes to push bundle to phased release branches
 - Adds synchronizing to master, development and other branches
 - Combines:
   - #804 
   - #805 
   - #806 

 ## Testing Plan
 - N/A - This is a CI/CD change. However, we will be testing the individual release tasks when we have the next formal SDK release. Currently, no customers are added to Release Order branches, but internal testing apps are.
 - We will use these internal testing apps to try out the release process before formally releasing any SDK changes to the master branch.
 - In the event of any issues, we can still manually roll out a release without customer impact.

 - Closes https://go.mparticle.com/work/SQDSDKS-5706


rmi22186 (approved):

.github/workflows/release-step-1.yml
```diff
@@ -281,11 +284,16 @@ jobs:
...
                   npx semantic-release --dry-run
 
+            # NPM Publish happens here
             - name: Release
               if: ${{ github.event.inputs.dryRun == 'false'}}
               run: |
                   npx semantic-release
 
+            - name: Get Package Version Number
+              uses: martinbeentjes/npm-get-version-action@v1.3.1
```
samdozor, line 294:
```suggestion
              uses: martinbeentjes/npm-get-version-action@3cf273023a0dda27efcd3164bdfb51908dd46a5b #v1.3.1
```

for 3rd party (eg non-github-owned) actions you should use a hardcoded git-sha. otherwise, they could change the tag out from underneath you. i grabbed the 1.3.1 sha from here: https://github.com/martinbeentjes/npm-get-version-action/commits/main/

.github/workflows/release-step-2.yml
```diff
@@ -0,0 +1,84 @@
...
+            - name: Checkout master branch
+              uses: actions/checkout@v3
+              with:
+                  fetch-depth: 0
+                  repository: ${{ github.repository }}
+                  token: ${{ secrets.MP_SEMANTIC_RELEASE_BOT }}
+                  ref: master
+
+            - name: Import GPG Key
+              uses: crazy-max/ghaction-import-gpg@v4
```
samdozor, line 70: same issue re: tag vs git sha - also any reason we're using such an old version?

.github/workflows/release-step-2.yml
```diff
@@ -0,0 +1,84 @@
...
+              with:
+                  gpg_private_key: ${{ secrets.GPG_PRIVATE_KEY }}
+                  passphrase: ${{ secrets.GPG_PASSPHRASE }}
+                  git_user_signingkey: true
+                  git_commit_gpgsign: true
+
+            - name: Merge release/${{ github.event.inputs.versionNumber}} into master branch
```
samdozor, line 77: i'm a little confused on the sequence here. if we have a release tag that we checked out and want to push to a release-order-branch, why change the contents of it by merging it into master?

.github/workflows/release-step-3.yml
```diff
@@ -0,0 +1,69 @@
...
+
+        - name: Merge release branch into master branch
+          run: |
+            git pull origin release/${{ github.event.inputs.versionNumber}}
+        
+        # We will eventually replace master with main
+        - name: Push release commits to main and master
+          run: |
+                git push origin HEAD:master
+                git push origin HEAD:main
```
samdozor, line 55: this is going to force the issue and push a main branch and we'll have both. can we wait to do this intentionally? try not to combine multiple things into one branch/pr like this.

.github/workflows/release-step-3.yml
```diff
@@ -0,0 +1,69 @@
...
+
+        - name: Push release commits to development
+          run: |
+            git push origin HEAD:development
+
+        - name: Push release commits to chore/dependabot
+          run: |
+            git push origin HEAD:chore/dependabot
```
samdozor, line 63: i suspect this will end up flakey. dependabot often runs ahead of our branches, no? so this will often fail or randomly fail. Not good for that to be part of a critical release process. maybe better to skip this entirely, or delete the branch at this point and let it get recreated.

.github/workflows/release-step-3.yml
```diff
@@ -0,0 +1,69 @@
+name: "Release - Step 3: Release to CDN and Synchronize Branches"
+
+on:
+    workflow_dispatch:
+        inputs:
+            versionNumber:
+                description: 'Which SDK version to syncronize? [X.XX.XX]'
```
samdozor, line 7:
```suggestion
                description: 'Which SDK version to synchronize? [X.XX.XX]'
```

.github/workflows/release-step-3.yml
```diff
@@ -0,0 +1,69 @@
...
+        
+        # We will eventually replace master with main
+        - name: Push release commits to main and master
+          run: |
+                git push origin HEAD:master
+                git push origin HEAD:main
+
+        - name: Push release commits to development
+          run: |
+            git push origin HEAD:development
```
samdozor, line 59: when would this be necessary? do we push commits to master before we push to development?

samdozor (changes requested):

.github/workflows/release-step-2.yml
alexs-mparticle: This step lets us ensure that both the release branch and master can be merged. As I look at it, I do see your point and we could potentially remove some redundancy by just having this step checkout the release branch and simply push it directly to the release-order-branch. I think it's safe to assume that if there are issues merging, they would have been caught during step 1.

Good callout.

.github/workflows/release-step-3.yml
alexs-mparticle: Not sure how this step got switched, but I'll move it to push to development before master so that it lines up with the current proecess.

.github/workflows/release-step-3.yml
alexs-mparticle: Good callout. I'll remove this.

.github/workflows/release-step-2.yml
samdozor: that's fine. while you're changing this line - can you please switch it to use the git sha of v4?

.github/workflows/release-step-3.yml
alexs-mparticle: As per our slack convo, this is something we already do to keep things in sync until we can formally cut over to use `main` instead of `master`. I think we should force the switch over within the next month.

.github/workflows/release-step-1.yml
```diff
@@ -250,21 +251,23 @@ jobs:
             NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
 
         steps:
-            - name: Checkout public master branch
+            - name: Checkout master branch
               uses: actions/checkout@v3
               with:
                   fetch-depth: 0
+                  repository: ${{ github.repository }}
+                  token: ${{ secrets.MP_SEMANTIC_RELEASE_BOT }}
```
rmi22186, line 259: These aren't needed currently, is there a reason they are needed moving forward?

.github/workflows/release-step-1.yml
```diff
@@ -250,21 +251,23 @@ jobs:
...
-              uses: crazy-max/ghaction-import-gpg@v4
+              uses: crazy-max/ghaction-import-gpg@e00cb83a68c1158b29afc5217dd0582cada6d172 #v4.4.0
               with:
                   gpg_private_key: ${{ secrets.GPG_PRIVATE_KEY }}
                   passphrase: ${{ secrets.GPG_PASSPHRASE }}
                   git_user_signingkey: true
                   git_commit_gpgsign: true
 
-            - name: Merge release branch into master branch
+            - name: Merge pre-release branch into master branch
```
rmi22186, line 270: Sometimes when I read this, I get confused about the step because I think about the remote master branch.  I think adding local/remote would be valuable when referring to master/development:

```suggestion
            - name: Merge pre-release branch into local master branch
```

.github/workflows/release-step-1.yml
```diff
@@ -296,34 +304,9 @@ jobs:
             - name: Push automated release commits to release branch
```
rmi22186, line 304:
```suggestion
            - name: Push automated release commits to pre-release branch
```

rmi22186 (changes requested):
Small nits.

One question - I know that the current release process does not delete the release branch upon completion.  Are there any negatives of adding that to step 3?

rmi22186 (approved):
LGTM!

PR #815, opened 2023-12-18
docs: Highlight refactor points for Persistence

alexs-mparticle:
## Instructions

 ## Summary
 - Add work links around Persistence Module for future refactor initiatives

 ## Testing Plan
 - This is only updating comments and should not contain any code changes.

 - Closes https://go.mparticle.com/work/SQDSDKS-6022

### DIFF
--- a/src/persistence.js
+++ b/src/persistence.js
@@ -11,10 +11,7 @@ var Base64 = Polyfill.Base64,
 export default function _Persistence(mpInstance) {
     var self = this;
 
-    // TODO: Refactor all the checks for Storage API and Cookie Settings
-    //       into a mechanism that can be the single source of truth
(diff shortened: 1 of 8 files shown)


src/persistence.js
```diff
@@ -1019,7 +1019,6 @@ export default function _Persistence(mpInstance) {
             ),
             date = new Date(),
             key = mpInstance._Store.storageName,
-            // QUESTION: Should this code go below the `useCookieStorage` check?
```
rmi22186: Should there be a ticket for this one?

rmi22186 (approved):

src/persistence.js
rmi22186: Works for me

PR #816, opened 2023-12-19
fix: Add filtered user identities to identity complete methods

alexs-mparticle:
## Instructions

 ## Summary
 - Adds `filteredUserIdentities` parameter to IdentityComplete methods which some kits are expecting

 ## Testing Plan
 - Using a sample app with a web kit that uses the [web kit wrapper](https://github.com/mparticle-integrations/mparticle-javascript-web-kit-wrapper), verify that onIdentifyComplete, onModifyComplete, onLoginComplete and onLogoutComplete pass a second argument within the kit code.
 - [Google Analytics 4](https://github.com/mparticle-integrations/mparticle-javascript-integration-google-analytics-4) is a kit that expects this second parameter

 - Closes https://go.mparticle.com/work/SQDSDKS-5221
 - Should be merged after #817

### DIFF
--- a/src/forwarders.interfaces.ts
+++ b/src/forwarders.interfaces.ts
@@ -9,7 +9,7 @@ import { IKitConfigs } from './configAPIClient';
 import { UserAttributes } from './persistence.interfaces';
 import { IdentityApiData } from '@mparticle/web-sdk';
 
-// TODO: Create Distinct MPForwarder and Kit interfaces
+// TODO: https://go.mparticle.com/work/SQDSDKS-6035
 export type Kit = Dictionary;
 export type MPForwarder = Dictionary;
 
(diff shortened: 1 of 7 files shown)


src/forwarders.interfaces.ts
```diff
@@ -1,9 +1,18 @@
...
+    SDKEvent,
+    SDKEventCustomFlags,
+    SDKUserIdentity,
+} from './sdkRuntimeModels';
 import { Dictionary } from './utils';
 import { IKitConfigs } from './configAPIClient';
 import { UserAttributes } from './persistence.interfaces';
 import { IdentityApiData } from '@mparticle/web-sdk';
 
+// TODO: Create Distinct MPForwarder and Kit interfaces
```
rmi22186, line 12: Let's make a ticket and put that here instead of the description.

rmi22186 (changes requested):
Minor comment, otherwise lgtm

src/forwarders.interfaces.ts
alexs-mparticle, line 12:
```suggestion
// TODO: https://go.mparticle.com/work/SQDSDKS-6035
```

rmi22186 (approved):

PR #817, opened 2023-12-19
fix: Replace identify method strings with constants

alexs-mparticle:
## Instructions

 ## Summary
 - Replace Identify method strings with constants
 - Also removes an unused parameter from `setForwarderOnUserIdentified`

 ## Testing Plan
 - Using a sample app, verify that Identify, Modify, Login and Logout have no changes in behavior.

 - Closes https://go.mparticle.com/work/SQDSDKS-6029

### DIFF
--- a/src/forwarders.js
+++ b/src/forwarders.js
@@ -4,7 +4,7 @@ import { isEmpty } from './utils';
 import KitFilterHelper from './kitFilterHelper';
 import Constants from './constants';
 
-const { Modify } = Constants.IdentityMethods;
+const { Modify, Identify, Login, Logout } = Constants.IdentityMethods;
 
 export default function Forwarders(mpInstance, kitBlocker) {
     var self = this;
--- a/src/forwarders.js
+++ b/src/forwarders.js
@@ -2,6 +2,9 @@ import Types from './types';
 import filteredMparticleUser from './filteredMparticleUser';
 import { isEmpty } from './utils';
 import KitFilterHelper from './kitFilterHelper';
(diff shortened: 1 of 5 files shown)


src/forwarders.js
```diff
@@ -440,6 +443,7 @@ export default function Forwarders(mpInstance, kitBlocker) {
         });
     };
 
+    // TODO: This is a v1 method, we should deprecate
```
rmi22186, line 446: Please instead make this a ticket and add the ticket instead of this description.

src/identity.js
```diff
@@ -1620,6 +1622,8 @@ export default function Identity(mpInstance) {
                             mpInstance._APIClient.prepareForwardingStats
                         );
                     }
+
+                    // TODO: This is a v1 method. We should deprecate
```
rmi22186, line 1626: Please instead make this a ticket and add the ticket instead of this description

rmi22186 (changes requested):
Small comments otherwise looks good.

src/identity.js
alexs-mparticle, line 1626:
```suggestion
                    // TODO: https://go.mparticle.com/work/SQDSDKS-6036
```

src/forwarders.js
alexs-mparticle, line 446:
```suggestion
    // TODO: https://go.mparticle.com/work/SQDSDKS-6036
```

rmi22186 (approved):

PR #819, opened 2023-12-21
feat: Cache identity responses

rmi22186:
## Instructions

 ## Summary
We want to cache our identity responses to limit superfluous identity calls.  This applies to login and identify for now.  In this PR, I leveraged the vault in order to create a new `idCache`.  

The keys for the idCache are a concatenation of the identities being passed to it along with the `das`.  An example value is as follows:

```
{
    mpid: 'foo',
    responseText: // cached response from the XHR request
    expirationTimestamp: // number - the date being 1 day added to the time the identity call was made
}
```

When an identify or login call is attempted, the idCache is checked.  If it is valid, we simply return the cached information, and do not make an identity call.  If the key does not exist, or it has expired, then we fire another identity call, and cache that one.

 ## Testing Plan
Integration tests. Unit tests.

 - Closes https://go.mparticle.com/work/SQDSDKS-5976

### DIFF
--- a/src/identity-utils.ts
+++ b/src/identity-utils.ts
@@ -1,0 +1,105 @@
+import { Dictionary } from './utils';
+import { LocalStorageVault } from './vault';
+import Types from './types';
+import { IdentityApiData, UserIdentities } from '@mparticle/web-sdk';
+import { isObject } from './utils';
+
+export interface IKnownIdentities extends UserIdentities {
+    device_application_stamp?: string;
+}
+
+export const cacheIdentityRequest = (
+    method: string,
+    identities: IKnownIdentities,
+    mpid: string,
+    expireTimestamp: number,
+    idCache: LocalStorageVault<Dictionary>
+): void => {
+    let cache = idCache.retrieve() || {};
+
+    let cacheKey = concatenateIdentities(method, identities);
+
+    cache[cacheKey] = { mpid, expireTimestamp };
+    idCache.store(cache);
+};
+
+// We need to ensure that identities are concatenated in a deterministic way, so
+// we sort the identities based on their enum.
+// we create an array, set the user identity at the index of the user identity type
+export const concatenateIdentities = (
+    method: string,
+    userIdentities: IKnownIdentities
+): string => {
+    // set DAS first since it is not an official identity type
+    let cacheKey: string = `${method}:device_application_stamp=${userIdentities.device_application_stamp};`;
+    const idLength: number = Object.keys(userIdentities).length;
+    let concatenatedIdentities: string;
+
+    if (idLength) {
+        let userIDArray: Array<string> = new Array(idLength);
+        // create an array where each index is equal to the user identity type
+        for (let key in userIdentities) {
+            if (key === 'device_application_stamp') {
+                continue;
+            } else {
+                userIDArray[Types.IdentityType.getIdentityType(key)] =
+                    userIdentities[key];
+            }
+        }
+
+        concatenatedIdentities = userIDArray.reduce(
+            (prevValue: string, currentValue: string, index: number) => {
+                let idName: string = Types.IdentityType.getIdentityName(index);
+                return `${prevValue}${idName}=${currentValue};`;
+            },
+            cacheKey
+        );
+    }
+
+    return concatenatedIdentities;
+};
+
+export const shouldCallIdentity = (
+    method: string,
+    proposedUserIdentities: IKnownIdentities,
+    idCache: LocalStorageVault<Dictionary>
+): Boolean => {
+    const cache = idCache.retrieve();
+
+    if (!cache) {
+        return true;
+    }
+
+    const cacheKey: string = concatenateIdentities(
+        method,
+        proposedUserIdentities
+    );
+
+    if (cache.hasOwnProperty(cacheKey)) {
+        const expireTimestamp = cache[cacheKey].expireTimestamp;
+        if (expireTimestamp > new Date().getTime()) return false;
+    }
+
+    return true;
+};
+
+export const createKnownIdentities = (
+    identityApiData: IdentityApiData,
+    deviceId: string
+): IKnownIdentities => {
+    var identitiesResult: IKnownIdentities = {};
+
+    if (
+        identityApiData &&
+        identityApiData.userIdentities &&
+        isObject(identityApiData.userIdentities)
+    ) {
+        for (var identity in identityApiData.userIdentities) {
+            identitiesResult[identity] =
+                identityApiData.userIdentities[identity];
+        }
+    }
+    identitiesResult.device_application_stamp = deviceId;
+
+    return identitiesResult;
+};
--- a/src/identity-utils.ts
+++ b/src/identity-utils.ts
@@ -27,7 +27,6 @@ export const cacheOrClearIdCache = (
     idCache: LocalStorageVault<Dictionary>,
     xhr: XMLHttpRequest
 ): void => {
-    debugger;
     // default the expire timestamp to one day in milliseconds unless a header comes back
     let expireTimestamp = new Date().getTime() + ONE_DAY_IN_SECONDS * MILLIS_IN_ONE_SEC;
 
--- a/src/identity-utils.ts
+++ b/src/identity-utils.ts
@@ -1,16 +1,14 @@
 import Constants from './constants';
 import { Dictionary, parseNumber } from './utils';
-import { LocalStorageVault } from './vault';
+import { BaseVault } from './vault';
 import Types from './types';
 import { IdentityApiData, UserIdentities } from '@mparticle/web-sdk';
 import { IdentityAPIMethod } from './validators';
 import { isObject } from './utils';
 const { Identify, Modify, Login, Logout } = Constants.IdentityMethods;
+import { ONE_DAY_IN_SECONDS, MILLIS_IN_ONE_SEC } from './constants';
 import { MParticleWebSDK } from './sdkRuntimeModels';
 
-const CACHE_HEADER = 'x-mp-max-age';
-const ONE_DAY_IN_SECONDS = 60 * 60 * 24;
-const MILLIS_IN_ONE_SEC = 1000;
 export interface IKnownIdentities extends UserIdentities {
     device_application_stamp?: string;
 }
@@ -24,9 +22,11 @@ export interface ICachedIdentityCall extends UserIdentities {
 export const cacheOrClearIdCache = (
     method: string,
     knownIdentities: IKnownIdentities,
-    idCache: LocalStorageVault<Dictionary>,
+    idCache: BaseVault<Dictionary>,
     xhr: XMLHttpRequest
 ): void => {
+    const CACHE_HEADER = 'x-mp-max-age';
+
     // default the expire timestamp to one day in milliseconds unless a header comes back
     let expireTimestamp = new Date().getTime() + ONE_DAY_IN_SECONDS * MILLIS_IN_ONE_SEC;
 
@@ -57,11 +57,11 @@ export const cacheIdentityRequest = (
     method: IdentityAPIMethod,
     identities: IKnownIdentities,
     expireTimestamp: number,
-    idCache: LocalStorageVault<Dictionary>,
+    idCache: BaseVault<Dictionary>,
     xhr: XMLHttpRequest
 ): void => {
-    let cache: Dictionary<ICachedIdentityCall> = idCache.retrieve() || {};
-    let cacheKey = concatenateIdentities(method, identities);
+    const cache: Dictionary<ICachedIdentityCall> = idCache.retrieve() || {};
+    const cacheKey = concatenateIdentities(method, identities);
 
     cache[cacheKey] = { responseText: xhr.responseText, status: xhr.status, expireTimestamp};
     idCache.store(cache);
@@ -74,8 +74,9 @@ export const concatenateIdentities = (
     method: IdentityAPIMethod,
     userIdentities: IKnownIdentities
 ): string => {
+    const DEVICE_APPLICATION_STAMP = 'device_application_stamp';
     // set DAS first since it is not an official identity type
-    let cacheKey: string = `${method}:device_application_stamp=${userIdentities.device_application_stamp};`;
+    let cacheKey: string = `${method}:${DEVICE_APPLICATION_STAMP}=${userIdentities.device_application_stamp};`;
     const idLength: number = Object.keys(userIdentities).length;
     let concatenatedIdentities: string = '';
 
@@ -83,7 +84,7 @@ export const concatenateIdentities = (
         let userIDArray: Array<string> = new Array();
         // create an array where each index is equal to the user identity type
         for (let key in userIdentities) {
-            if (key === 'device_application_stamp') {
+            if (key === DEVICE_APPLICATION_STAMP) {
                 continue;
             } else {
                 userIDArray[Types.IdentityType.getIdentityType(key)] =
@@ -93,7 +94,7 @@ export const concatenateIdentities = (
 
         concatenatedIdentities = userIDArray.reduce(
             (prevValue: string, currentValue: string, index: number) => {
-                let idName: string = Types.IdentityType.getIdentityName(index);
+                const idName: string = Types.IdentityType.getIdentityName(index);
                 return `${prevValue}${idName}=${currentValue};`;
             },
             cacheKey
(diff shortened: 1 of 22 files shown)


src/identity-utils.ts
```diff
@@ -0,0 +1,160 @@
...
+        proposedUserIdentities
+    );
+
+    const cache = idCache.retrieve();
+    const cachedIdentity = cache ? cache[cacheKey] : null;
+
+    return cachedIdentity;
+};
+
+export const createKnownIdentities = (
```
rmi22186, line 124: create a new file to put non-cache utility methods in

test/src/tests-identity-utils.ts
```diff
@@ -0,0 +1,428 @@
...
+                facebook: '02',
+                facebookcustomaudienceid: '09',
+                google: '04',
+                twitter: '03',
+                microsoft: '05',
+                yahoo: '06',
+            };
+
+            const key: string = concatenateIdentities('identify', userIdentities);
+            const expectedResult: string = 'identify:device_application_stamp=first;other=00;customerid=01;facebook=02;twitter=03;google=04;microsoft=05;yahoo=06;email=07;facebookcustomaudienceid=09;other2=10;other3=11;other4=12;other5=13;other6=14;other7=15;other8=16;other9=17;other10=18;mobile_number=19;phone_number_2=20;phone_number_3=21;';
```
rmi22186, line 171: base 64 this at the method level

src/identity-utils.ts
```diff
@@ -0,0 +1,160 @@
...
+// We need to ensure that identities are concatenated in a deterministic way, so
+// we sort the identities based on their enum.
+// we create an array, set the user identity at the index of the user identity type
+export const concatenateIdentities = (
+    method: string,
+    userIdentities: IKnownIdentities
+): string => {
+    // set DAS first since it is not an official identity type
+    let cacheKey: string = `${method}:device_application_stamp=${userIdentities.device_application_stamp};`;
+    const idLength: number = Object.keys(userIdentities).length;
```
rmi22186, line 41: this length doesn't actually matter

src/identity-utils.ts
```diff
@@ -0,0 +1,160 @@
...
+    expireTimestamp: number;
+}
+
+export const cacheIdentityRequest = (
+    method: string,
+    identities: IKnownIdentities,
+    expireTimestamp: number,
+    idCache: LocalStorageVault<Dictionary>,
+    xhr: XMLHttpRequest
+): void => {
```
alexs-mparticle, line 24: Based on our conversation, I think keeping this as a pure function makes things cleaner. We should remove the dependency on the vault and simply have it return a dictionary that can be inserted into the vault.

For example, the new signature would become:

```typescript
export interface IdentityRequestCacheItem {
    key: string,
    identityRequestCacheItem: ICacheIdentityCall
}

const createIdentityRequestCacheItem = (
    method: string,
    identities: IKnownIdentities,
    expireTimestamp: number,
    httpStatus: number,
    httpResponseText: string
): IdentityRequestCacheItem => {}
```

src/identity-utils.ts
```diff
@@ -0,0 +1,160 @@
...
+
+// We need to ensure that identities are concatenated in a deterministic way, so
+// we sort the identities based on their enum.
+// we create an array, set the user identity at the index of the user identity type
+export const concatenateIdentities = (
+    method: string,
+    userIdentities: IKnownIdentities
+): string => {
+    // set DAS first since it is not an official identity type
+    let cacheKey: string = `${method}:device_application_stamp=${userIdentities.device_application_stamp};`;
```
alexs-mparticle, line 40: This probably can be a convenience method to make things cleaner.

src/identity-utils.ts
```diff
@@ -0,0 +1,160 @@
...
+
+    return concatenatedIdentities;
+};
+
+export const hasValidCachedIdentity = (
+    method: string,
+    proposedUserIdentities: IKnownIdentities,
+    idCache?: LocalStorageVault<Dictionary>
+): Boolean => {
+    // There is an edhge case where multiple identity calls are taking place
```
alexs-mparticle, line 73:
```suggestion
    // There is an edge case where multiple identity calls are taking place 
```

src/identity-utils.ts
```diff
@@ -0,0 +1,160 @@
...
+    method: string,
+    proposedUserIdentities: IKnownIdentities,
+    idCache?: LocalStorageVault<Dictionary>
+): Boolean => {
+    // There is an edhge case where multiple identity calls are taking place 
+    // before identify fires, so there may not be a cache.  See what happens when 
+    // the ? in idCache is removed to the following test
+    // "queued events contain login mpid instead of identify mpid when calling 
+    // login immediately after mParticle initializes"
+    const cache = idCache?.retrieve();
```
alexs-mparticle, line 78: Instead of taking in the entire Vault Object, maybe we can just have it take the output of the `Vault.retrieve()`, and return a boolean if the item exists?

This also might help remove some of the need for the guards below.

src/identity-utils.ts
rmi22186: I think this can actually be done in a future ticket when identity.js gets refactored into identity.ts.  Right now this is the only non-cache related utility method.

src/identity-utils.ts
rmi22186: Based on our 1/19 discussion, we will err on not complicating the `identity.js` module and tidying up functions, and keeping thing in TS for easier future refactoring when `identity.js --> identity.ts`

src/identity-utils.ts
```diff
@@ -0,0 +1,197 @@
...
+import { Dictionary, parseNumber } from './utils';
+import { LocalStorageVault } from './vault';
+import Types from './types';
+import { IdentityApiData, UserIdentities } from '@mparticle/web-sdk';
+import { IdentityAPIMethod } from './validators';
+import { isObject } from './utils';
+const { Identify, Modify, Login, Logout } = Constants.IdentityMethods;
+import { MParticleWebSDK } from './sdkRuntimeModels';
+
+const CACHE_HEADER = 'x-mp-max-age';
```
rmi22186, line 11: move these into any method if it is only used by that method

src/identity-utils.ts
```diff
@@ -0,0 +1,197 @@
...
+    // set DAS first since it is not an official identity type
+    let cacheKey: string = `${method}:device_application_stamp=${userIdentities.device_application_stamp};`;
+    const idLength: number = Object.keys(userIdentities).length;
+    let concatenatedIdentities: string = '';
+
+    if (idLength) {
+        let userIDArray: Array<string> = new Array();
+        // create an array where each index is equal to the user identity type
+        for (let key in userIdentities) {
+            if (key === 'device_application_stamp') {
```
rmi22186, line 86: make this a constant so that 78/86 can avoid typos

test/src/tests-identity-utils.ts
```diff
@@ -0,0 +1,452 @@
...
+    describe('#concatenateIdentities', () => {
+        it('should return a concatenated user identity string based on the order of IdentityType enum', () => {
+            const userIdentities: IKnownIdentities = {
+                device_application_stamp: 'first',
+                customerid: '01',
+                email: '07',
+                other: '00',
+                other2: '10',
+                other3: '11',
+                other4: '12',
```
rmi22186, line 211: change the values from stringified numbers to more human readable strings because the numbers could be confusing given the method uses array indices.  The two could be conflated.

src/identity.js
```diff
@@ -425,6 +450,39 @@ export default function Identity(mpInstance) {
...
+                    let shouldReturnCachedIdentity = hasValidCachedIdentity(
+                        Login,
+                        identityApiRequest.known_identities,
+                        self.idCache
+                    );
+
+                    // If Identity is cached, then immediately parse the identity response
+                    if (shouldReturnCachedIdentity) {
+                        const cachedIdentity = getCachedIdentity(
+                            Identify,
```
rmi22186, line 467: ?
```suggestion
                            Login,
```

src/identity.js
```diff
@@ -425,6 +450,39 @@ export default function Identity(mpInstance) {
...
+                            identityApiRequest.known_identities,
+                            self.idCache
+                        );
+
+                        self.parseIdentityResponse(
+                            cachedIdentity,
+                            mpid,
+                            callback,
+                            identityApiData,
+                            Identify,
```
rmi22186, line 477: ?
```suggestion
                            Login,
```

src/mockBatchCreator.ts
```diff
@@ -12,7 +12,7 @@ const mockFunction = function() {
 };
 export default class _BatchValidator {
     private getMPInstance() {
-        return {
+        return ({
```
rmi22186, line 15: undo all changes in this file

src/mp-instance.js
```diff
@@ -1302,12 +1304,21 @@ function completeSDKInitialization(apiKey, config, mpInstance) {
...
     }
 
+    // add a new function to apply items to the store that require config to be returned
     mpInstance._Store.storageName = mpInstance._Helpers.createMainStorageName(
         config.workspaceToken
     );
     mpInstance._Store.prodStorageName = mpInstance._Helpers.createProductStorageName(
         config.workspaceToken
     );
+    mpInstance._Identity.idCache = new LocalStorageVault(
```
rmi22186, line 1314: add comment to explain why this is happening here and not in identity.js.
```suggestion
// idCache deepnds on store.storageName, which doesn't exist when identity.js gets instantiated.  identity.js gets instantiated right when mp is initialized, but store doesnt' get instantiated until later.
    mpInstance._Identity.idCache = new LocalStorageVault(
```

src/mp-instance.js
```diff
@@ -1302,12 +1304,21 @@ function completeSDKInitialization(apiKey, config, mpInstance) {
...
+        `${mpInstance._Store.storageName}-id-cache`,
+        {
+            logger: mpInstance.Logger,
+        }
+    );
+
+    removeExpiredIdentityCacheDates(mpInstance._Identity.idCache);
```
rmi22186, line 1321: space after

src/sdkRuntimeModels.ts
```diff
@@ -11,6 +11,7 @@ import { IPersistence } from './persistence.interfaces';
 import { IMPSideloadedKit } from './sideloadedKit';
 import { ISessionManager } from './sessionManager';
 import { Kit, MPForwarder } from './forwarders.interfaces';
+// import { LocalStorageVault } from './vault';
```
rmi22186, line 14: remove

test/src/tests-persistence.ts
```diff
@@ -789,7 +817,8 @@ describe('migrations and persistence-related', () => {
         done();
     });
 
-    it('integration test - should remove a previous MPID as a key from cookies if new user attribute added and exceeds the size of the max cookie size', done => {
+    // this test needs to update the cookie size in order to 
+    xit('integration test - should remove a previous MPID as a key from cookies if new user attribute added and exceeds the size of the max cookie size', done => {
```
rmi22186, line 821: undo this xit

test/src/tests-persistence.ts
```diff
@@ -652,6 +659,8 @@ describe('migrations and persistence-related', () => {
             JSON.stringify({ mpid: 'mpid1', is_logged_in: false }),
         ]);
 
+        // tick clock forward to go beyond the 1 day identity caching
+        clock.tick(MILLISECONDS_IN_ONE_DAY_PLUS_ONE_SECOND);
```
rmi22186, line 663: the clock.ticks can probably be removed now that we are hiding this behind a feature flag

test/src/tests-persistence.ts
```diff
@@ -1794,41 +1876,41 @@ describe('migrations and persistence-related', () => {
...
-
-        mParticle.Identity.identify();
-
-        expect(
-            mParticle.getInstance()._Persistence.getFirstSeenTime('current')
-        ).to.not.equal(null);
-
-        done();
-    });
+    // it.only('fst should be set when the user does not change, after an identify request', done => {
```
rmi22186, line 1879: undo to see if this test passes

src/identity-utils.ts
```diff
@@ -0,0 +1,197 @@
...
+import Types from './types';
+import { IdentityApiData, UserIdentities } from '@mparticle/web-sdk';
+import { IdentityAPIMethod } from './validators';
+import { isObject } from './utils';
+const { Identify, Modify, Login, Logout } = Constants.IdentityMethods;
+import { MParticleWebSDK } from './sdkRuntimeModels';
+
+const CACHE_HEADER = 'x-mp-max-age';
+const ONE_DAY_IN_SECONDS = 60 * 60 * 24;
+const MILLIS_IN_ONE_SEC = 1000;
```
alexs-mparticle, line 13:
```suggestion
const MILLIS_IN_ONE_SEC = 1000;

```

test/src/config/constants.ts
```diff
@@ -12,6 +12,8 @@ export const urls = {
 };
 
 export const MILLISECONDS_IN_ONE_MINUTE = 60000;
+export const MILLISECONDS_IN_ONE_DAY = MILLISECONDS_IN_ONE_MINUTE * 60 * 24;
```
alexs-mparticle, line 15: I see you're already defining this in `src/identity-utils.ts` unless I'm mistaken.

alexs-mparticle (changes requested):

src/identity-utils.ts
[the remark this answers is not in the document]
rmi22186: good suggestion.  i think we can use it elsewhere in the code as well

src/identity-utils.ts
rmi22186: I feel like if it's just a function that will return the same thing, that might be over engineering.

src/identity-utils.ts
rmi22186: Can you clarify?  It's been a month since this message.  Do you still feel strongly about this?

src/mockBatchCreator.ts
rmi22186: seems to throw a TS error when i do .  will keep it for now.

test/src/config/constants.ts
rmi22186: i took a stab at trying to make things a little more consistent.  but i'm going to make a ticket about trying to more globally replace all of our instances of doing manual math like 60 * 24 * 1000, etc etc and i'll add the ticket as a todo

test/src/tests-persistence.ts
rmi22186: needs to stay.  updated comment as a future todo

test/src/tests-persistence.ts
```diff
@@ -792,7 +815,7 @@ describe('migrations and persistence-related', () => {
     it('integration test - should remove a previous MPID as a key from cookies if new user attribute added and exceeds the size of the max cookie size', done => {
         mParticle._resetForTests(MPConfig);
         mParticle.config.useCookieStorage = true;
-        mParticle.config.maxCookieSize = 650;
+        mParticle.config.maxCookieSize = 700;
```
rmi22186, line 818: I had to change this because previously we were not passing in user identities to the identity calls in this test, so identities weren't being saved to cookies. Now that identities are being saved to cookies the cookies are larger and this test was breaking.  I needed to find a larger cookie size that would allow thisi test to behave as it was previously behaving.

src/identity-utils.ts
alexs-mparticle: My initial feedback was that instead of having the third argument be `BaseVault<Dictionary>` it should be an the output of `idCache.retrieve()`. This makes the cache functionality less dependent on the storage method (vault versus some in memory) versus the actual data. I think it only saves us one line of code, so maybe it's not a big deal to leave this as is. I think this is self-contained enough where we can clean it up later if need be.

src/identity-utils.ts
```diff
@@ -0,0 +1,198 @@
...
+        );
+    }
+
+    return concatenatedIdentities;
+};
+
+export const hasValidCachedIdentity = (
+    method: IdentityAPIMethod,
+    proposedUserIdentities: IKnownIdentities,
+    idCache?: BaseVault<Dictionary>
```
alexs-mparticle, line 110: I don't think this should be just a generic `Dictionary` but rather a clearly defined `IdentityCache` Type.

src/validators.ts
```diff
@@ -9,7 +9,7 @@ import {
 import Constants from './constants';
 import { IdentityApiData } from '@mparticle/web-sdk';
 
-type IdentityAPIMethod = 'login' | 'logout' | 'identify' | 'modify';
+export type IdentityAPIMethod = 'login' | 'logout' | 'identify' | 'modify';
```
alexs-mparticle, line 12: I think it's better to reuse the existing constants rather than redefine new strings. Less of a chance of getting bitten by a typo, IMO. 

Also, not sure if you need to export since I don't think this type is being used outside of this file, but I'm not sure.

```suggestion
type IdentityAPIMethod = valueof<typeof Constants.IdentityMethods>;
```

alexs-mparticle (approved):

PR #823, opened 2024-01-03
test: Add browserstack testing 

rmi22186:
## Instructions

 ## Summary
These 6 commits have previously been approved. This branch was rebased with `development`, and we are now merging this to development.

 ## Testing Plan
Locally ran tests and also ran browserstack tests.

 - Closes https://go.mparticle.com/work/SQDSDKS-5780


.github/workflows/cross-browser-testing.yml
```diff
@@ -0,0 +1,40 @@
+name: 'BrowserStack Test'
+on: [push, pull_request, workflow_dispatch]
+
+jobs:
+  ubuntu-job:
```
alexs-mparticle, line 5: I'm curious about why we're calling this `ubuntu-job`. Are we planning on having other jobs for different OSes?

rmi22186: Oh interesting. I pulled the entire job from https://github.com/browserstack/github-actions?tab=readme-ov-file#sample-workflow-with-usage-of-both-actions

alexs-mparticle (approved):

PR #832, opened 2024-01-19
Initial attempt to decouple persistence

alexs-mparticle:
## Instructions

 ## Summary

 ## Testing Plan


src/mp-instance.js
```diff
@@ -59,7 +59,13 @@ export default function mParticleInstance(instanceName) {
     this._instanceName = instanceName;
     this._NativeSdkHelpers = new NativeSdkHelpers(this);
     this._SessionManager = new SessionManager(this);
+
+    // QUESTION: Why can't we just init the store here?
```
rmi22186, line 63: line 169 shows
```
instance._Store = new Store(config, instance);
```
The store needs the config, which is an async call in a self hosted set up.  It would be possible to initialize store if store did not have items from the config.  But then we'd have an incomplete store in the beginning vs a complete store later on in the SDK which could make debugging more difficult.  Alternatively, we could put config items vs non config items in different "stores", but that woul be a larger refactor.

src/mp-instance.js
```diff
@@ -59,7 +59,13 @@ export default function mParticleInstance(instanceName) {
...
+    // QUESTION: Should we still initialize if persistence is off?
```
rmi22186, line 64: Do you mean should be initialize the persistence module if persistence is off?  

Short answer is yes for now.

Long answer - I think persistence is too tightly coupled right now, and may be referenced elsewhere...but in a future state this makes sense.  having it around would take up a bit more memory in the browser which while I think we should optimize for, erring on everything working as expected is more important right now. 

For the MVP I was thinking more about finding the points in code that we save/read from persistence, as opposed to turning it off completely, which would be a larger task.

src/mp-instance.js
```diff
@@ -84,6 +90,8 @@ export default function mParticleInstance(instanceName) {
     this._Identity = new Identity(this);
     this.Identity = this._Identity.IdentityAPI;
     this.generateHash = this._Helpers.generateHash;
+
+    // QUESTION: Why do we not init store here instead of persistence?
```
rmi22186, line 94: Not sure I follow the question.  Persistence is not being initialized here.  the getDeviceId method lives on persistence (maybe it shouldn't), and we are exposing that method to the user.  Here's a great example of where the persistence module needs to remain initialized despite turning saving to persistence off.

src/mp-instance.js
```diff
@@ -157,15 +167,21 @@ export default function mParticleInstance(instanceName) {
             delete instance._Store;
         }
         instance._Store = new Store(config, instance);
-        instance._Store.isLocalStorageAvailable = instance._Persistence.determineLocalStorageAvailability(
-            window.localStorage
-        );
+
```
rmi22186, line 170: Just to be clear, this method is just for the tests.  i don't think there is harm in always clearing persistence, but we can discuss since I haven't actually run the tests.  

I'm just trying to avoid any potential debugging rabbit holes by being overly aggressive on reseting, rather than potentially forgetting to set the flag during a test.  This is something we can add a comment for also.

src/mp-instance.js
```diff
@@ -229,7 +247,12 @@ export default function mParticleInstance(instanceName) {
             self.setDeviceId(guid);
         }, self);
         if (queued) return;
-        this._Persistence.setDeviceId(guid);
+
+        // debugger;
+        // TODO: Replace this with Store.setDeviceId
```
rmi22186, line 252: This could be a blanket TODO. Anywhere where something shouldbe saved to persistence but persistence is off, we should instead save to store.

rmi22186: I think that could be done as a second ticket though. The first ticket is to get the SDK working and not crash when persistence is turned off.  The second ticket could then be save to/read from store instead.

src/mp-instance.js
```diff
@@ -1333,7 +1362,12 @@ function completeSDKInitialization(apiKey, config, mpInstance) {
     // https://go.mparticle.com/work/SQDSDKS-6044
     if (!mpInstance._Store.webviewBridgeEnabled) {
         // Load any settings/identities/attributes from cookie or localStorage
-        mpInstance._Persistence.initializeStorage();
+        // FIXME: if persistence is turned off, device id never gets set
+        // FIXME: maybe we set device id in _Store and that fixes everything?
```
rmi22186, line 1366: Your suggestion about setting to _Store seems rigth.  Set it there and read it from there when persistence is disabled.

src/persistence.js
```diff
@@ -11,6 +11,11 @@ var Base64 = Polyfill.Base64,
 export default function _Persistence(mpInstance) {
     var self = this;
 
+    // QUESTION: Should isEnabled be a param on the constructor, or should it read from config?
+    this.isEnabled = function() {
```
rmi22186, line 15: I think future state where we decouple persistence enough, then if persistence is not initialized, having it as a param on the contructor makes less sense.  for now, let's read from config and check on store.SDKConfig the way you are currently doing it.

src/persistence.js
```diff
@@ -1058,6 +1063,9 @@ export default function _Persistence(mpInstance) {
     };
 
     this.getPersistence = function() {
+        if (!this.isEnabled()) {
```
rmi22186, line 1066: See comment about about not using the param on the constructor.  I am open to feedback about this decision.  I'd rather just have it consistent everywhere though so that it's easier to grok, even if it's more verbose.

src/sessionManager.ts
```diff
@@ -44,6 +44,7 @@ export default function SessionManager(
                 self.endSession();
                 self.startNewSession();
             } else {
+                // QUESTION: Should we check for current user if we don't have persistence?
```
rmi22186, line 47: I would say no.  But the no persistence state is not a realistic use case and more an intermediate state to get us to plug and play persistence types, if that changes what you were thinking about doing here.

src/sessionManager.ts
```diff
@@ -144,33 +148,60 @@ export default function SessionManager(
...
+
+        // debugger;
+
+        // Check Store for last time event was sent. If there is a different value in persistence,
+        // use that and update the Store
+        // TODO: Check if dateLastEventSent and return early if it does not exist
+        let timeLastEventSent: number = mpInstance._Store.dateLastEventSent?.getTime();
+        if (
+            persistence?.gs &&
+            persistence?.gs.les &&
```
rmi22186, line 191:
```suggestion
            persistence?.gs?.les &&
```

src/sessionManager.ts
```diff
@@ -144,33 +148,60 @@ export default function SessionManager(
...
+        const persistence: IPersistenceMinified = mpInstance._Persistence.getPersistence();
 
-        // TODO: https://go.mparticle.com/work/SQDSDKS-5684
-        if (!cookies) {
-            return;
-        }
+        // debugger;
 
-        if (cookies.gs && !cookies.gs.sid) {
+        // QUESTION: What constitutes ending a session if we don't have persistence?
```
rmi22186, line 160: Addressed in earlier comments, but if we don' have persistence, then the expectation is that everything is saved to the store, so the logic would be if persistence is disabled, check the store instead

src/sessionManager.ts
```diff
@@ -179,8 +210,7 @@ export default function SessionManager(
                     messageType: Types.MessageType.SessionEnd,
                 });
 
-                mpInstance._Store.sessionStartDate = null;
```
rmi22186, line 182: I think line 182 was removed erroneously.

src/sessionManager.ts
```diff
@@ -213,20 +245,18 @@ export default function SessionManager(
...
-                }
+            if (
+                persistence?.gs &&
+                persistence?.gs.sid &&
+                mpInstance._Store.sessionId !== persistence.gs.sid
+            ) {
+                mpInstance._Store.sessionId = persistence.gs.sid;
+            }
+
+            // TODO: Make this a store method
```
rmi22186, line 256: Can you clarify this TODO?  Are you arguing that `startNewSession` should be on the store and not on the session module?

src/store.ts
```diff
@@ -279,6 +288,12 @@ export default function Store(
             this.SDKConfig.useCookieStorage = false;
         }
 
+        if (config.hasOwnProperty('usePersistence')) {
+            this.SDKConfig.usePersistence = config.usePersistence;
+        } else {
+            this.SDKConfig.usePersistence = false;
```
rmi22186, line 294: This should be true by default.  Is this `false` because you are in the middle of development?

src/sessionManager.ts
rmi22186: I think the answer the the above is yes.  However I think there's a lot happening in this method that isn't just store related.  Maybe there is a way to keep the session logic in session, and at the end, create a new method about updating the store's session attributes.

rmi22186 (commented):
This looks good so far. Here are my comments for Monday morning.  I didn't look at the tests since they may change given my feedback.

src/mp-instance.js
alexs-mparticle: I think I'm on the same track as you. I simply wanted to flag this question here so we can use it as a touch point for a conversation of what the relationship between the store and persistence _should_ be so that we can think of an plan for decoupling.

src/mp-instance.js
alexs-mparticle: The store currently has several "defaults" that are independent of the `config` and the store also contains its own `SDKConfig` object, so we technically can split this up. Perhaps we can instantiate the Store earlier and then split out the `SDKConfig` logic as its own class/object that initialized after the async config operation. Then this `SDKConfig` gets passed back to the Store after the fact, thus completing the SDK init process.

I agree that this is out of scope and I just wanted to flag this so we can think about the future refactor just so we have a plan lined up.

src/mp-instance.js
alexs-mparticle: My thought is that the Device ID specifically should be part of the Store (or at the very least independent of persistence) since it we are simply generating a random GUID if persistence is unavailable. I think this lines up with the idea of moving the Store "in front of" Persistence, so that we treat Store as the record of truth.

src/sessionManager.ts
alexs-mparticle: I intentionally moved it to a new method `Store.nullifySession()` since it made sense to me to keep this grouped with the other nullify session logic. Is there a reason to not nullify the session start date when we nullify a session?

src/sessionManager.ts
alexs-mparticle: I probably should have put this comment before the `startNewSession` method. My thought is that we're doing a lot of session related activities within the context of a store. Part of me thinks that we should split out the Session related persistence/store attributes into a seaprate Session object, so that the Session Manager can simply manage Sessions. As we've discussed, the store holds a lot of data about the state of the SDK, but a lot of unrelated modules directly modify the state which makes it hard to predict what the current state of the SDK is, so I thought a quick path to tidying things up would be to create a `startNewSession` method in the store to clean some things up.

src/store.ts
alexs-mparticle: Good catch. I made this default to `false` before we made our decision to default it to `true`. I'll update.

src/persistence.js
alexs-mparticle: Based on our conversation, we will leave this `isEnabled` method here in case we have other checks to determine if Persistence is disabled for other reasons besides the config. For example, if local storage not exist.

src/identity.js
```diff
@@ -18,8 +19,17 @@ export default function Identity(mpInstance) {
             if (persistence) {
                 persistence.cu = currentMPID;
                 persistence.gs.csm = currentSessionMPIDs;
-                mpInstance._Persistence.savePersistence(persistence);
+            } else {
+                persistence = mpInstance._Store.getPersistence();
```
alexs-mparticle, line 23:
```suggestion
                persistence = mpInstance._Store.getPersistence();
                persistence.cu = currentMPID;
                persistence.gs.csm = currentSessionMPIDs;
```

src/identity.js
```diff
@@ -1464,30 +1524,42 @@ export default function Identity(mpInstance) {
 
                 if (prevUser) {
                     mpInstance._Persistence.setLastSeenTime(previousMPID);
+                    mpInstance._Store.setLastSeenTime(previousMPID);
                 }
 
+                // QUESTION: Does this only matter for cookies or do we need it for
```
alexs-mparticle, line 1530: It's called cookies, but actually applies to all persistence methods.

src/store.ts
```diff
@@ -161,6 +177,36 @@ export interface IStore {
...
+    getPersistence?(): IPersistenceMinified;
+    getGlobalStorageAttributes?: () => IGlobalStoreV2MinifiedKeys;
+    savePersistence?(persistence: IPersistenceMinified): void;
+    getConsentState?(mpid: MPID): ConsentState; // TODO: Make sure this is the correct type
+    setConsentState?(mpid, consentState: ConsentState): void;
+
+    userAttributes?: Dictionary<UserAttributes>;
+    getUserAttributes?(mpid: MPID): UserAttributes;
+
+    // QUESTION: Is there a difference between this and currentUserIdentities?
```
alexs-mparticle, line 196:
```suggestion
    // QUESTION: Is there a difference between this and currentUserIdentities?
```

src/store.ts
```diff
@@ -417,6 +498,234 @@ export default function Store(
...
+    this.getUserIdentities = (mpid: MPID) => {
+        return { ...this.persistenceData[mpid]?.ui };
+    };
+
+    // QUESTION: Should we just use get/set instead?
+    this.getDeviceId = (): string => this.deviceId;
+    this.setDeviceId = (deviceId: string): void => {
+        // TODO: This should update persistence
+        this.deviceId = deviceId;
+        mpInstance._Persistence.update();
```
alexs-mparticle, line 571: Do we still need to call update internally?

src/store.ts
```diff
@@ -417,6 +498,234 @@ export default function Store(
...
+            return null;
+        }
+
+        return mpInstance._Consent.ConsentSerialization.fromMinifiedJsonObject(
+            consentState
+        );
+    };
+
+    this.setConsentState = (mpid: MPID, consentState: ConsentState) => {
+        // QUESTION: What is the context here?
```
alexs-mparticle, line 586: Comment can be removed as it is not necessary.

PR #833, opened 2024-01-23
feat: Hash keys for idCache

rmi22186:
## Instructions

 ## Summary
Hash the concatenated identities.

 ## Testing Plan
Unit tests

 - Closes https://go.mparticle.com/work/SQDSDKS-6037

### DIFF
--- a/src/identity-utils.ts
+++ b/src/identity-utils.ts
@@ -62,8 +62,9 @@ export const cacheIdentityRequest = (
 ): void => {
     const cache: Dictionary<ICachedIdentityCall> = idCache.retrieve() || ({} as Dictionary<ICachedIdentityCall>);
     const cacheKey = concatenateIdentities(method, identities);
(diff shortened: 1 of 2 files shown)


src/identity-utils.ts
```diff
@@ -62,7 +63,7 @@ export const cacheIdentityRequest = (
     const cache: Dictionary<ICachedIdentityCall> = idCache.retrieve() || ({} as Dictionary<ICachedIdentityCall>);
     const cacheKey = concatenateIdentities(method, identities);
 
-    cache[cacheKey] = { responseText: xhr.responseText, status: xhr.status, expireTimestamp};
+    cache[generateHash(cacheKey)] = { responseText: xhr.responseText, status: xhr.status, expireTimestamp};
```
alexs-mparticle, line 66: I think this should be split into two lines for readability.

alexs-mparticle (approved):

PR #834, opened 2024-01-23
feat: Implement identity caching

rmi22186:
## Instructions

 ## Summary
This PR is a feature branch which composes of the following previously approved PRs:
* https://github.com/mParticle/mparticle-web-sdk/pull/819
* https://github.com/mParticle/mparticle-web-sdk/pull/833
It implements Identity Caching to prevent superfluous identity calls.  Applies only to identify/login right now.  Modify and logout will clear the cache.  Each cache lasts for 24 hours before being removed.

 ## Testing Plan

### DIFF
--- a/src/identity-utils.ts
+++ b/src/identity-utils.ts
@@ -1,12 +1,10 @@
-import Constants from './constants';
+import Constants, { ONE_DAY_IN_SECONDS, MILLIS_IN_ONE_SEC } from './constants';
 import { Dictionary, parseNumber } from './utils';
 import { BaseVault } from './vault';
 import Types from './types';
 import { IdentityApiData, UserIdentities } from '@mparticle/web-sdk';
 import { IdentityAPIMethod } from './sdkRuntimeModels';
-import { isObject } from './utils';
-import { ONE_DAY_IN_SECONDS, MILLIS_IN_ONE_SEC } from './constants';
-import { generateHash } from './utils';
+import { isObject, generateHash } from './utils';
 
 const { Identify, Modify, Login, Logout } = Constants.IdentityMethods;
 
@@ -114,7 +112,7 @@ export const hasValidCachedIdentity = (
     method: IdentityAPIMethod,
     proposedUserIdentities: IKnownIdentities,
     idCache?: BaseVault<Dictionary<ICachedIdentityCall>>
-): Boolean => {
+): boolean => {
     // There is an edge case where multiple identity calls are taking place 
     // before identify fires, so there may not be a cache.  See what happens when 
     // the ? in idCache is removed to the following test
@@ -178,7 +176,7 @@ export const createKnownIdentities = (
         identityApiData?.userIdentities &&
         isObject(identityApiData.userIdentities)
     ) {
-        for (var identity in identityApiData.userIdentities) {
+        for (let identity in identityApiData.userIdentities) {
             identitiesResult[identity] =
                 identityApiData.userIdentities[identity];
         }
--- a/src/identity-utils.ts
+++ b/src/identity-utils.ts
@@ -24,16 +24,20 @@ export const cacheOrClearIdCache = (
     method: string,
     knownIdentities: IKnownIdentities,
     idCache: BaseVault<Dictionary<ICachedIdentityCall>>,
-    xhr: XMLHttpRequest
+    xhr: XMLHttpRequest,
+    parsingCachedResponse: boolean,
 ): void => {
+    // when parsing a response that has already been cached, simply return instead of attempting another cache
+    if (parsingCachedResponse) { return; }
+
     const CACHE_HEADER = 'x-mp-max-age';
 
     // default the expire timestamp to one day in milliseconds unless a header comes back
-    let expireTimestamp = new Date().getTime() + ONE_DAY_IN_SECONDS * MILLIS_IN_ONE_SEC;
-
+    let now = new Date().getTime();
+    let expireTimestamp = now + ONE_DAY_IN_SECONDS * MILLIS_IN_ONE_SEC;
     if (xhr.getAllResponseHeaders().includes(CACHE_HEADER)) {
         expireTimestamp =
-            parseNumber(xhr.getResponseHeader(CACHE_HEADER)) * MILLIS_IN_ONE_SEC;
+            now + (parseNumber(xhr.getResponseHeader(CACHE_HEADER)) * MILLIS_IN_ONE_SEC);
     }
 
     switch (method) {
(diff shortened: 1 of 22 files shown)


src/identity-utils.ts
```diff
@@ -0,0 +1,201 @@
+import Constants from './constants';
+import { Dictionary, parseNumber } from './utils';
+import { BaseVault } from './vault';
+import Types from './types';
+import { IdentityApiData, UserIdentities } from '@mparticle/web-sdk';
+import { IdentityAPIMethod } from './sdkRuntimeModels';
+import { isObject } from './utils';
+const { Identify, Modify, Login, Logout } = Constants.IdentityMethods;
+import { ONE_DAY_IN_SECONDS, MILLIS_IN_ONE_SEC } from './constants';
+import { generateHash } from './utils';
```
alexs-mparticle, line 10:
```suggestion
import { ONE_DAY_IN_SECONDS, MILLIS_IN_ONE_SEC } from './constants';
import { generateHash } from './utils';

const { Identify, Modify, Login, Logout } = Constants.IdentityMethods;
```

src/identity-utils.ts
```diff
@@ -0,0 +1,201 @@
...
+    expireTimestamp: number;
+}
+
+export const cacheOrClearIdCache = (
+    method: string,
+    knownIdentities: IKnownIdentities,
+    idCache: BaseVault<Dictionary<ICachedIdentityCall>>,
+    xhr: XMLHttpRequest
+): void => {
+    const CACHE_HEADER = 'x-mp-max-age';
```
alexs-mparticle, line 28: I think this should go to `./constants.js`. I can see us reusing this elsewhere.

alexs-mparticle (changes requested):

src/identity-utils.ts
rmi22186: I'm going to push back on this one. I actually don't see a use right now for the future.  If there is though then I can move it over during that code commit when it is used elsewhere, but since it's just used in this function, I'd prefer to keep it here.

src/identity-utils.ts
```diff
@@ -24,16 +24,20 @@ export const cacheOrClearIdCache = (
...
+    // when parsing a response that has already been cached, simply return instead of attempting another cache
+    if (parsingCachedResponse) { return; }
+
     const CACHE_HEADER = 'x-mp-max-age';
 
     // default the expire timestamp to one day in milliseconds unless a header comes back
-    let expireTimestamp = new Date().getTime() + ONE_DAY_IN_SECONDS * MILLIS_IN_ONE_SEC;
-
+    let now = new Date().getTime();
+    let expireTimestamp = now + ONE_DAY_IN_SECONDS * MILLIS_IN_ONE_SEC;
```
rmi22186, line 37: Originally I was just setting the timestamp to what was in the header, which is just 1 day in milliseconds, so it was firing the identity calls all the time since the logic was such that it would fire if the timestamp expires, which it always does.

src/identity-utils.ts
```diff
@@ -24,16 +24,20 @@ export const cacheOrClearIdCache = (
...
     if (xhr.getAllResponseHeaders().includes(CACHE_HEADER)) {
         expireTimestamp =
-            parseNumber(xhr.getResponseHeader(CACHE_HEADER)) * MILLIS_IN_ONE_SEC;
+            now + (parseNumber(xhr.getResponseHeader(CACHE_HEADER)) * MILLIS_IN_ONE_SEC);
```
rmi22186, line 40: Originally I was just setting the timestamp to what was in the header, which is just 1 day in milliseconds, so it was firing the identity calls all the time since the logic was such that it would fire if the timestamp expires, which it always does.

src/identity-utils.ts
```diff
@@ -0,0 +1,259 @@
...
+    identityApiData: IdentityApiData,
+    deviceId: string
+): IKnownIdentities => {
+    const identitiesResult: IKnownIdentities = {};
+
+    if (
+        identityApiData?.userIdentities &&
+        isObject(identityApiData.userIdentities)
+    ) {
+        for (let identity in identityApiData.userIdentities) {
```
alexs-mparticle, line 188: I think this can be reduced to a single expression.

```suggestion
    if (isObject(identityApiData?.userIdentities)) {
        for (let identity in identityApiData.userIdentities) {
```

src/identity-utils.ts
```diff
@@ -0,0 +1,259 @@
...
+    }
+    identitiesResult.device_application_stamp = deviceId;
+
+    return identitiesResult;
+};
+
+export const removeExpiredIdentityCacheDates = (idCache: BaseVault<Dictionary<ICachedIdentityCall>>) => {
+    const cache: Dictionary<ICachedIdentityCall> = idCache.retrieve() || {} as Dictionary<ICachedIdentityCall>;
+
+    idCache.purge();
```
alexs-mparticle, line 201: Nit: There's no need to purge the cache since `.store()` overwrites what's already in there.

src/identity-utils.ts
```diff
@@ -0,0 +1,259 @@
...
+    mpid: string,
+    callback: IdentityCallback,
+    identityApiData: IdentityApiData,
+    identityMethod: IdentityAPIMethod
+): boolean =>  {
+    if (
+        mpInstance._Helpers.getFeatureFlag(
+            Constants.FeatureFlags.CacheIdentity
+        )
+    ) {
```
alexs-mparticle, line 229: Based on our conversation earlier, you should do this check outside of this function. This way, there's no need to pass in the entire `mpInstance` object, and we reduce the number of arguments.

src/mparticle-instance-manager.js
```diff
@@ -355,7 +355,7 @@ function mParticle() {
     };
 
     this.Identity = {
-        HTTPCodes: self.getInstance().Identity.HTTPCodes,
+        HTTPCodes: Constants.HTTPCodes,
```
alexs-mparticle, line 358: Thank you.

src/identityApiClient.js
```diff
@@ -80,7 +81,9 @@ export default function IdentityAPIClient(mpInstance) {
                         previousMPID,
                         callback,
                         originalIdentityApiData,
-                        method
+                        method,
+                        knownIdentities,
+                        false
```
alexs-mparticle, line 86: Is this boolean a new argument? I don't see it in the signature of `parseIdentityResponse` in Identity.js

alexs-mparticle (changes requested):

src/identityApiClient.js
rmi22186: It's there.  
image

alexs-mparticle (approved):

PR #839, opened 2024-01-30
feat: Implement a flag to disable persistence

alexs-mparticle:
## Instructions

 ## Summary
 - Provide a simple means to disable persistence so that neither cookie or local storage are directly accessed.

 ## Testing Plan
 - When the `usePersistence` flag is set to `false`, neither cookie or localStorage should be accessed via the Persistence layer
 - Note: This is independent of Offline Storage, as that can be controlled via a feature flag for the time being.

 - Closes https://go.mparticle.com/work/SQDSDKS-6072

### DIFF
--- a/src/cookieSyncManager.js
+++ b/src/cookieSyncManager.js
@@ -7,6 +7,10 @@ export default function cookieSyncManager(mpInstance) {
 
     // Public
     this.attemptCookieSync = function(previousMPID, mpid, mpidIsNotInCookies) {
+        if (!mpInstance._Persistence.isEnabled()) {
+            return false;
+        }
+
         // TODO: These should move inside the for loop
         var pixelConfig,
             lastSyncDateForModule,
--- a/src/persistence.js
+++ b/src/persistence.js
@@ -11,6 +11,10 @@ var Base64 = Polyfill.Base64,
 export default function _Persistence(mpInstance) {
     var self = this;
 
+    this.isEnabled = function() {
+        return mpInstance._Store.SDKConfig.usePersistence;
+    };
+
     // https://go.mparticle.com/work/SQDSDKS-5022
     this.useLocalStorage = function() {
         return (
@@ -42,7 +46,9 @@ export default function _Persistence(mpInstance) {
 
             // https://go.mparticle.com/work/SQDSDKS-6046
             if (mpInstance._Store.isLocalStorageAvailable) {
-                storage = window.localStorage;
+                if (this.isEnabled()) {
+                    storage = window.localStorage;
+                }
                 if (mpInstance._Store.SDKConfig.useCookieStorage) {
                     // For migrating from localStorage to cookies -- If an instance switches from localStorage to cookies, then
                     // no mParticle cookie exists yet and there is localStorage. Get the localStorage, set them to cookies, then delete the localStorage item.
@@ -57,7 +63,9 @@ export default function _Persistence(mpInstance) {
                         } else {
                             allData = localStorageData;
                         }
-                        storage.removeItem(mpInstance._Store.storageName);
+                        if (this.isEnabled()) {
+                            storage.removeItem(mpInstance._Store.storageName);
+                        }
                     } else if (cookies) {
                         allData = cookies;
                     }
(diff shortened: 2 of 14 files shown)


src/persistence.js
```diff
@@ -11,6 +11,15 @@ var Base64 = Polyfill.Base64,
 export default function _Persistence(mpInstance) {
     var self = this;
 
+    this.isEnabled = function () {
+        return (
+            (mpInstance._Store &&
+                mpInstance._Store.SDKConfig &&
+                mpInstance._Store.SDKConfig.usePersistence) ||
+            false
```
alexs-mparticle, line 19: Should return true by default.

src/persistence.js
```diff
@@ -19,7 +28,10 @@ export default function _Persistence(mpInstance) {
         );
     };
 
-    this.initializeStorage = function() {
+    this.initializeStorage = function () {
+        if (!this.isEnabled()) {
```
alexs-mparticle, line 32: This can be moved further down in the method so that we still have some of the initialization occur.

src/persistence.js
```diff
@@ -466,7 +491,10 @@ export default function _Persistence(mpInstance) {
         document.cookie = cookieName + '=' + '' + expires + '; path=/' + domain;
     };
 
-    this.getCookie = function() {
+    this.getCookie = function () {
+        if (!this.isEnabled()) {
+            return;
```
alexs-mparticle, line 496: return null instead

src/persistence.js
```diff
@@ -844,7 +875,10 @@ export default function _Persistence(mpInstance) {
         }
     };
 
-    this.getCookieDomain = function() {
+    this.getCookieDomain = function () {
+        if (!this.isEnabled()) {
+            return;
```
alexs-mparticle, line 880: return null

rmi22186:
This looks good so far.  I think more tests need to be added.  There should be integration tests for:
1. Initialize mParticle, send an event (with custom attributes, custom flags), the event should be sent as expected in a batch (I'd also check a few fields to ensure they are filled in appropriately, like sessionid, the custom attributes/flags).  it's possible when you check for the batch, it doesn't exist yet because of the timeout on a batch, so calling mParticle.upload() would ensure the batch is uploaded.
2. set session attribute, call session end and confirm that event has the session attributes on it
3. set user attributes, log an event, ensure the batch has the user attributes on it
4. identify or login with user identities, ensure the batch has the user identities on it
5. you mentioned consent breaks things, so maybe that comes as part of the next PR to prevent any blockers

src/cookieSyncManager.js
```diff
@@ -7,6 +7,10 @@ export default function cookieSyncManager(mpInstance) {
 
     // Public
     this.attemptCookieSync = function(previousMPID, mpid, mpidIsNotInCookies) {
+        if (!mpInstance._Persistence.isEnabled()) {
```
rmi22186, line 10: looks like you can mark https://go.mparticle.com/work/SQDSDKS-6100 as done in this sprint also :)

test/src/tests-event-logging.js
```diff
@@ -1044,4 +1048,122 @@ describe('event logging', function() {
...
+        );
+        expect(
+            testEventBatch.source_request_id,
+            'batch source_request_id'
+        ).to.not.equal(null);
+        expect(
+            testEventBatch.timestamp_unixtime_ms,
+            'batch timestamp'
+        ).to.not.equal(null);
+        expect(testEventBatch.user_attributes, 'batch user attributes').to.eqls(
```
rmi22186, line 1138: i just did a search for what `eqls` means.  apparently it's deep equality. Is there a difference than between `to.eqls` and `to.deep.equal`?   I prefer the latter since it's more intuitive

test/src/tests-identities-attributes.js
```diff
@@ -1214,7 +1214,71 @@ describe('identities and attributes', function() {
...
+    });
+
+    afterEach(function() {
+        mockServer.restore();
+        fetchMock.restore();
     });
 
-});
\ No newline at end of file
+    it('shoulds et session attribute', done => {
```
rmi22186, line 1243:
```suggestion
    it('should set session attribute', done => {
```

rmi22186 (changes requested):
very minor nits..otherwise looks good!

rmi22186 (approved):

PR #841, opened 2024-02-09
feat: Implement fetching user audiences

rmi22186:
## Instructions

 ## Summary
This PR provides for a new API on the current user: mParticle.Identity.getCurrentUser().getUserAudiences(callback).

The url it accesses is: https://jssdks.mparticle.com/v1/{apikey}/audiences?mpid=3984273492347328 and will return to the user the following schema to a callback:

```
{
  currentAudiences: Audience[],
  pastAudiences: Audience[]
}
```

 ## Testing Plan
 - [] Was this tested locally? If not, explain why.
We are waiting for the backend to be updated before 

 - Closes https://go.mparticle.com/work/SQDSDKS-5994

### DIFF
--- a/src/audienceManager.ts
+++ b/src/audienceManager.ts
@@ -4,7 +4,7 @@ import {
     XHRUploader,
     AsyncUploader,
     fetchPayload
-} from './batchUploader';
+} from './uploaders';
 import Audience from './audience';
 
 type AudienceResponseMessage = 'ar';
@@ -12,9 +12,9 @@ export interface IAudienceServerResponse {
     dt: AudienceResponseMessage;
     id: string;
     ct: number;
-    m: IAudienceMembership[];
+    m: IMinifiedAudienceMembership[];
 }
-export interface IAudienceMembership {
+export interface IMinifiedAudienceMembership {
     id: number; // AudienceId
     n: string; // External Name
     c: AudienceListMembershipUpdateType[]; // AudienceMembershipChanges
@@ -116,7 +116,7 @@ export default class AudienceManager {
 export const parseUserAudiences = (audienceServerResponse: IAudienceServerResponse): IMPParsedAudienceMemberships => {
     const currentAudiences: Audience[] = [];
     const pastAudiences: Audience[] = [];
-    audienceServerResponse?.m?.forEach((membership: IAudienceMembership) => {
+    audienceServerResponse?.m?.forEach((membership: IMinifiedAudienceMembership) => {
         if (membership.c[0].a === AudienceMembershipChangeAction.Add) {
             currentAudiences.push(new Audience(
                 membership.id,
--- a/src/audienceManager.ts
+++ b/src/audienceManager.ts
@@ -1,0 +1,143 @@
+import { SDKLoggerApi } from './sdkRuntimeModels';
+import {
+    FetchUploader,
+    XHRUploader,
+    AsyncUploader,
+    fetchPayload
+} from './batchUploader';
+import Audience from './audience';
+
+type AudienceResponseMessage = 'ar';
+export interface IAudienceServerResponse {
+    dt: AudienceResponseMessage;
+    id: string;
+    ct: number;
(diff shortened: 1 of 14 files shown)


src/audienceManager.ts
```diff
@@ -0,0 +1,138 @@
...
+import {
+    FetchUploader,
+    XHRUploader,
+    AsyncUploader,
+    fetchPayload
+} from './batchUploader';
+import Audience from './audience';
+
+type AudienceResponseMessage = 'ar';
+export interface IAudienceServerResponse {
```
alexs-mparticle, line 11: We should append the word `Minified` to these interfaces just so we can differentiate from the "Full" interfaces if we ever implement them.

src/batchUploader.ts
```diff
@@ -424,25 +424,26 @@ export class BatchUploader {
     }
 }
 
-abstract class AsyncUploader {
+export abstract class AsyncUploader {
```
alexs-mparticle, line 427: If you're reusing the async uploader, let's just move them into their own file.

test/src/tests-audience-manager.ts
```diff
@@ -0,0 +1,147 @@
...
+        const audienceManager = new AudienceManager(
+            Constants.DefaultBaseUrls.userAudienceUrl,
+            apiKey,
+            newLogger,
+            testMPID
+        );
+
+        expect(audienceManager.logger).to.be.ok;
+        expect(audienceManager.url).to.equal(userAudienceUrl);
+        expect(audienceManager.url).to.equal(userAudienceUrl);
```
alexs-mparticle, line 70: Is this a duplicate test?

alexs-mparticle (changes requested):

test/src/tests-audience-manager.ts
rmi22186: Good catch

src/batchUploader.ts
alexs-mparticle: Let's do it!

test/src/tests-audience-manager.ts
```diff
@@ -0,0 +1,147 @@
...
+    });
+
+    afterEach(() => {
+        sinon.restore();
+        mockServer.reset();
+        fetchMock.restore();
+    });
+
+    it('should have proper properties on AudienceManager initialization', () => {
+        const newLogger:SDKLoggerApi = new Logger(window.mParticle.config);
```
alexs-mparticle, line 60:
```suggestion
        const newLogger: SDKLoggerApi = new Logger(window.mParticle.config);
```

alexs-mparticle (changes requested):

alexs-mparticle (approved):

src/identity.js
```diff
@@ -1265,6 +1267,23 @@ export default function Identity(mpInstance) {
             getFirstSeenTime: function() {
                 return mpInstance._Persistence.getFirstSeenTime(mpid);
             },
+            /**
+             * Get user segments for user
+             * @method getUserSegments
+             * @return {Object} an array with user segments????
```
BrandonStalnaker, line 1273: comments probably need cleaned up here?

BrandonStalnaker (commented):
Looks good just 1 minor thing

rmi22186:
@BrandonStalnaker i updated.  please take another look.

src/audienceManager.ts
```diff
@@ -0,0 +1,138 @@
...
+}
+export interface IMinifiedAudienceMembership {
+    id: number; // AudienceId
+    n: string; // External Name
+    c: AudienceListMembershipUpdateType[]; // AudienceMembershipChanges
+    s: string[]; // ModuleEndpointExternalReferenceKeys
+}
+
+export enum AudienceMembershipChangeAction {
+    Unknown = 'unkown',
```
BrandonStalnaker, line 25: misspelling?

rmi22186: fixed

PR #844, opened 2024-03-06
fix: Update audience logic to new backend response

rmi22186:
## Instructions

 ## Summary
Code has been updated to respect the following simplified backend response:

``` 
{
    ct: 1710441407915,
    dt: 'cam',  // current audience memberships
    id: 'foo-id-2',
    current_audience_memberships: [
        {
            audience_id: 9876,
            expiration_timestamp_ms: 1234
        },
        {
            audience_id: 5432,
            expiration_timestamp_ms: null
        },
    ]
};
```

The response to the user is an object that removes `ct`, `dt`, and `id` from the above, and only returns `current_audience_memberships`.

 ## Testing Plan
 - [X] Was this tested locally? If not, explain why. - Tested using PR environment

 - Closes https://go.mparticle.com/work/SQDSDKS-6149

### DIFF
--- a/src/audience.ts
+++ b/src/audience.ts
@@ -1,9 +1,12 @@
 export default class Audience {
-    public id: number;
-    public name: string;
+    public audience_id: number;
+    public expiration_timestamp_ms?: number;
 
-    constructor(id: number, name: string) {
-        this.id = id;
-        this.name = name;
+    constructor(
+        audience_id: number,
+        expiration_timestamp_ms?: number
+    ) {
+        this.audience_id = audience_id;
+        this.expiration_timestamp_ms = expiration_timestamp_ms || null;
     }
 }
--- a/src/audience.ts
+++ b/src/audience.ts
@@ -1,6 +1,5 @@
 export default class Audience {
     public audience_id: number;
-    public expiration_timestamp_ms?: number;
 
     constructor(
         audience_id: number,
--- a/src/audience.ts
+++ b/src/audience.ts
@@ -3,9 +3,7 @@ export default class Audience {
 
     constructor(
         audience_id: number,
-        expiration_timestamp_ms?: number
     ) {
         this.audience_id = audience_id;
-        this.expiration_timestamp_ms = expiration_timestamp_ms || null;
     }
 }
--- a/src/audienceManager.ts
+++ b/src/audienceManager.ts
@@ -29,7 +29,7 @@ export default class AudienceManager {
         logger: SDKLoggerApi
     ) {
         this.logger = logger;
-        this.url = `https://${userAudienceUrl}${apiKey}/audience?mpid=`;
+        this.url = `https://${userAudienceUrl}${apiKey}/audience`;
         this.userAudienceAPI = window.fetch
             ? new FetchUploader(this.url)
             : new XHRUploader(this.url);
@@ -44,7 +44,7 @@ export default class AudienceManager {
                 Accept: '*/*',
             },
         };
-        const audienceURLWithMPID = `${this.url}${mpid}`;
+        const audienceURLWithMPID = `${this.url}?mpid=${mpid}`;
         let userAudiencePromise: Response;
 
         try {
--- a/src/audienceManager.ts
+++ b/src/audienceManager.ts
@@ -11,37 +11,6 @@ export interface IAudienceMemberships {
     audience_memberships: Audience[];
 }
 
-// export interface IAudienceMemberships extends IAudienceServerResponse {}
-
-// export interface IAudienceServerResponse {
-//     dt: AudienceResponseMessage;
-//     id: string;
-//     ct: number;
-//     m: IMinifiedAudienceMembership[];
-// }
-// export interface IMinifiedAudienceMembership {
-//     id: number; // AudienceId
-//     n: string; // External Name
-//     c: AudienceListMembershipUpdateType[]; // AudienceMembershipChanges
-//     s: string[]; // ModuleEndpointExternalReferenceKeys
-// }
-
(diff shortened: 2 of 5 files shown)


src/audienceManager.ts
```diff
@@ -45,17 +26,16 @@ export default class AudienceManager {
     constructor(
         userAudienceUrl: string,
         apiKey: string,
-        logger: SDKLoggerApi,
-        mpid: string
+        logger: SDKLoggerApi
     ) {
         this.logger = logger;
-        this.url = `https://${userAudienceUrl}${apiKey}/audience?mpid=${mpid}`;
+        this.url = `https://${userAudienceUrl}${apiKey}/audience?mpid=`;
```
alexs-mparticle, line 32: I think it makes more sense to just have the base url here and then append `?mpid=` below. It makes it more obvious as to what's going on.

```suggestion
        this.url = `https://${userAudienceUrl}${apiKey}/audience`;
```

src/audience.ts
```diff
@@ -1,9 +1,12 @@
 export default class Audience {
-    public id: number;
-    public name: string;
+    public audience_id: number;
+    public expiration_timestamp_ms?: number;
```
einsteinx2, line 3: Should we just remove the `expiration_timestamp_ms` field entirely since we aren't using it? Customers may think null means never expires or something when really we're just never returning anything

rmi22186, line 3:
```suggestion
```

test/src/tests-audience-manager.ts
```diff
@@ -21,7 +17,7 @@ declare global {
     }
 }
 
-const userAudienceUrl = `https://${Constants.DefaultBaseUrls.userAudienceUrl}${apiKey}/audience?mpid=${testMPID}`
+const userAudienceUrl = `https://${Constants.DefaultBaseUrls.userAudienceUrl}${apiKey}/audience?mpid=`
```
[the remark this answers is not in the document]
rmi22186: I think this is a broader issue and probably should be its own ticket as part of the api client refactor, which i made here https://mparticle-eng.atlassian.net/browse/SQDSDKS-6234.  2 reasons.  

1. Our URLs have a few things in common, but when we use something like Helpers.createServiceUrl, there can still be a bunch of appending things still happening.

https://github.com/mParticle/mparticle-web-sdk/blob/4740596cb9f19e968e74c4d3c103d50c846d0888/src/identityApiClient.js#L115-L131

2. In a test environment, I think there is value in knowing exactly what format a URL is, as opposed to sending it to a function.

Do you have a specific suggestion, or is my interpretation of your comment off?

src/audience.ts
```diff
@@ -1,9 +1,11 @@
 export default class Audience {
-    public id: number;
-    public name: string;
+    public audience_id: number;
 
-    constructor(id: number, name: string) {
-        this.id = id;
-        this.name = name;
+    constructor(
```
einsteinx2, line 4: Looks like the constructor needs to be updated to remove the references to expiration_timestamp_ms

einsteinx2 (changes requested):

test/src/tests-audience-manager.ts
```diff
@@ -71,73 +66,117 @@ describe('AudienceManager', () => {
...
+            fetchMock.get(testMPIDAudienceEndpoint, {
+                status: 200,
+                body: JSON.stringify(audienceMembershipServerResponse)
+            });
+
+            const audienceMembershipServerResponse2: IAudienceMembershipsServerResponse = {
+                ct: 1710441407915,
+                dt: 'cam',
+                id: 'foo-id-2',
+                current_audience_memberships: [
```
einsteinx2, line 135: Tests should be updated to remove references to expiration_timestamp_ms as well

einsteinx2 (changes requested):

BrandonStalnaker (approved):
Looks good logic wise

einsteinx2 (approved):
Looks good!

src/audienceManager.ts
```diff
@@ -7,34 +7,15 @@ import {
 } from './uploaders';
 import Audience from './audience';
 
-type AudienceResponseMessage = 'ar';
-export interface IAudienceServerResponse {
-    dt: AudienceResponseMessage;
+export interface IAudienceMembershipsServerResponse {
+    dt: 'cam';  // current audience memberships
```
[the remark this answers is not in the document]
rmi22186: yep.

src/audienceManager.ts
```diff
@@ -64,26 +44,30 @@ export default class AudienceManager {
...
+                const parsedUserAudienceMemberships: IAudienceMemberships = {
+                    currentAudienceMemberships: userAudienceMembershipsServerResponse?.current_audience_memberships
+                }
+
                 try {
-                    callback(parsedUserAudiences);
+                    callback(parsedUserAudienceMemberships);
                 } catch(e) {
-                    this.logger.error('Error invoking callback on user audience response.')
+                    this.logger.error('Error invoking callback on IAudienceMemberships.');
```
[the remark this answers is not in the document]
rmi22186: i agree.  not sure how that happened.

rmi22186, line 70:
```suggestion
                    this.logger.error('Error invoking callback on user audience response.');
```

test/src/tests-audience-manager.ts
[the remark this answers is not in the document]
alexs-mparticle: 1. You bring up some interesting points and I think as part of the ticket you created, we probably should write unit tests around url generation logic specifically so that it things are better organized.

2. I agree with this completely

alexs-mparticle (approved):

PR #847, opened 2024-03-20
Initial attempt to refactor persistence

alexs-mparticle:
## Instructions

 ## Summary

 ## Testing Plan

### DIFF
--- a/src/filteredMparticleUser.js
+++ b/src/filteredMparticleUser.js
@@ -67,7 +67,6 @@ export default function filteredMparticleUser(
             return userAttributesLists;
         },
         getAllUserAttributes: function() {
-            // TODO: May not need to make a copy since Store should return a copy
             var userAttributesCopy = {};
             var userAttributes = mpInstance._Store.getUserAttributes(mpid);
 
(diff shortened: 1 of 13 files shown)


src/filteredMparticleUser.js
```diff
@@ -67,10 +67,9 @@ export default function filteredMparticleUser(
             return userAttributesLists;
         },
         getAllUserAttributes: function() {
+            // TODO: May not need to make a copy since Store should return a copy
```
alexs-mparticle, line 70: Note te self: we need this empty object to create a new subset of attributes for filtering.

src/identity.js
```diff
@@ -1007,9 +1001,13 @@ export default function Identity(mpInstance) {
 
                     delete userAttributes[key];
 
-                    if (cookies && cookies[mpid]) {
-                        cookies[mpid].ua = userAttributes;
-                        mpInstance._Persistence.savePersistence(cookies, mpid);
+                    // QUESTION: can we just use mpInstance._Store.setUserAttributes here?
+                    if (persistence && persistence[mpid]) {
+                        persistence[mpid].ua = userAttributes;
+                        // QUESTION: why is mpid an argument here when it's not in the signature?
```
alexs-mparticle, line 1007: Note to self: Defined in `getCurrentUser` at the top of this nested block

PR #849, opened 2024-03-20
fix: Create a simple merge object utility

alexs-mparticle:
## Instructions

 ## Summary

 ## Testing Plan
 - Run automated tests

 - Closes https://go.mparticle.com/work/SQDSDKS-6249

### DIFF
--- a/test/src/tests-utils.ts
+++ b/test/src/tests-utils.ts
@@ -370,6 +370,10 @@ describe('Utils', () => {
 
         it('returns true if Custom Flag Property is not valid', () => {
             expect(isValidCustomFlagProperty(null), 'null').to.equal(false);
+            expect(isValidCustomFlagProperty(function(){}), 'function').to.equal(false);
+            expect(isValidCustomFlagProperty(undefined), 'undefined').to.equal(false);
+        });
+    });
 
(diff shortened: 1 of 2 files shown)


test/src/tests-utils.ts
```diff
@@ -369,8 +370,50 @@ describe('Utils', () => {
...
+            const obj1 = { foo: 'bar' };
+            const obj2 = { fizz: 'buzz' };
+
+            expect(mergeObjects<any>(obj1, obj2)).to.deep.equal({
+                foo: 'bar',
+                fizz: 'buzz',
+            });
+        });
+
+        it('should override values with the same keys', () => {
```
rmi22186, line 389: Does this make more sense?  I was having a hard time trying to think of something that made sense.

```suggestion
        it('should allow later arguments to override values of previous arguments with the same keys', () => {
```

rmi22186 (changes requested):
slight nit in the test phrasing

test/src/tests-utils.ts
alexs-mparticle: What about `should override values with the same keys in the order they appear`?

rmi22186 (approved):

PR #855, opened 2024-04-02
fix: Migrate deviceId methods to Store

alexs-mparticle:
## Instructions

 ## Summary
 - Move device ID methods from Persistence into Store

 ## Testing Plan
 - Using a sample app, verify that updating device id affects changes in the Store and Persistence.

 - Closes https://go.mparticle.com/work/SQDSDKS-6290

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -455,8 +455,7 @@ export default function Store(
     this.setDeviceId = (deviceId: string) => {
         this.deviceId = deviceId;
         this.persistenceData.gs.das = deviceId;
-
-        mpInstance._Persistence.savePersistence(this.persistenceData);
+        mpInstance._Persistence.update();
     };
 
     this.nullifySession = (): void => {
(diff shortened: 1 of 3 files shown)


src/store.ts
```diff
@@ -418,6 +445,14 @@ export default function Store(
             }
         }
     }
+
+    this.getDeviceId = () => this.deviceId;
+    this.setDeviceId = (deviceId: string) => {
+        this.deviceId = deviceId;
+        this.persistenceData.gs.das = deviceId;
+
+        mpInstance._Persistence.savePersistence(this.persistenceData);
```
rmi22186, line 454: I originally thought that since nothing else is saving to the store persistence, that if the user ever set the deviceid, that only the deviceId would be saved the the store, and that everything else would be null.  Then I saw that this was merged into a feature branch, so I think we just need to be 100% sure before merging feature branch into development that every item in Store.persistenceData is filled out and matches persistence.

rmi22186 (approved):

rmi22186 (approved):

rmi22186 (approved):

PR #856, opened 2024-04-02
fix: Migrate First and Last Seen Time Setters and Getters to Store

alexs-mparticle:
## Instructions

 ## Summary
 - Migrate First and Last Seen Time Setters and Getters to Store

 ## Testing Plan
 - Run automated tests
 - Functions are not integrated into main SDK logic and cannot be locally tested yet.

 - Closes https://go.mparticle.com/work/SQDSDKS-6308

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -169,6 +169,11 @@ export interface IStore {
     getDeviceId?(): string;
     setDeviceId?(deviceId: string): void;
 
+    getFirstSeenTime?(mpid: MPID): number;
+    setFirstSeenTime?(mpid: MPID, time?: number): void;
+    getLastSeenTime?(mpid: MPID): number;
+    setLastSeenTime?(mpid: MPID, time?: number): void;
+
     nullifySession?: () => void;
 }
 
@@ -463,6 +468,83 @@ export default function Store(
         this.dateLastEventSent = null;
         this.sessionAttributes = {};
         mpInstance._Persistence.update();
+    };
+
+    this.getFirstSeenTime = (mpid: MPID) => {
+        if (!mpid) {
+            return null;
+        }
+
+        if (
+            this.persistenceData &&
+            this.persistenceData[mpid] &&
+            this.persistenceData[mpid].fst
+        ) {
+            return this.persistenceData[mpid].fst;
+        } else {
+            return null;
+        }
+    };
+
+    this.setFirstSeenTime = (mpid: MPID, _time?: number) => {
+        if (!mpid) {
+            return;
+        }
+
+        const time = _time || new Date().getTime();
+
+        if (this.persistenceData) {
+            if (!this.persistenceData[mpid]) {
+                this.persistenceData[mpid] = {};
+            }
+            if (!this.persistenceData[mpid].fst) {
+                this.persistenceData[mpid].fst = time;
+                mpInstance._Persistence.savePersistence(this.persistenceData);
+            }
+        }
+    };
+
+    this.getLastSeenTime = (mpid: MPID) => {
+        if (!mpid) {
+            return null;
+        }
+
+        // TODO: We should store the current user in the Store's mpid rather than
+        //       relying on the Identity object
+        const currentUser = mpInstance.Identity.getCurrentUser();
+
+        if (mpid === currentUser?.getMPID()) {
+            // if the mpid is the current user, its last seen time is the current time
+            return new Date().getTime();
+        } else {
+            if (
+                this.persistenceData &&
+                this.persistenceData[mpid] &&
+                this.persistenceData[mpid].lst
+            ) {
+                return this.persistenceData[mpid].lst;
+            } else {
+                return null;
+            }
+        }
+    };
+
+    this.setLastSeenTime = (mpid: MPID, _time?: number) => {
+        if (!mpid) {
+            return;
+        }
+
+        const time = _time || new Date().getTime();
+
+        if (this.persistenceData) {
+            if (!this.persistenceData[mpid]) {
+                this.persistenceData[mpid] = {};
+            }
+            if (!this.persistenceData[mpid].lst) {
+                this.persistenceData[mpid].lst = time;
+                mpInstance._Persistence.savePersistence(this.persistenceData);
+            }
+        }
     };
 }
 
--- a/src/store.ts
+++ b/src/store.ts
@@ -515,15 +515,14 @@ export default function Store(
         if (mpid === currentUser?.getMPID()) {
             // if the mpid is the current user, its last seen time is the current time
             return new Date().getTime();
+        } else if (
+            this.persistenceData &&
+            this.persistenceData[mpid] &&
+            this.persistenceData[mpid].lst
+        ) {
+            return this.persistenceData[mpid].lst;
         } else {
-            if (
-                this.persistenceData &&
-                this.persistenceData[mpid] &&
-                this.persistenceData[mpid].lst
-            ) {
-                return this.persistenceData[mpid].lst;
-            } else {
-                return null;
+            return null;
             }
         }
     };
--- a/src/store.ts
+++ b/src/store.ts
@@ -509,8 +509,7 @@ export default function Store(
             return null;
         }
 
-        // TODO: We should store the current user in the Store's mpid rather than
-        //       relying on the Identity object
+        // https://go.mparticle.com/work/SQDSDKS-6315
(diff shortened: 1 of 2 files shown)


src/store.ts
```diff
@@ -453,6 +458,83 @@ export default function Store(
...
+    this.getFirstSeenTime = (mpid: MPID) => {
+        if (!mpid) {
+            return null;
+        }
+
+        if (
+            this.persistenceData &&
+            this.persistenceData[mpid] &&
+            this.persistenceData[mpid].fst
+        ) {
```
rmi22186, line 471:
```suggestion
        if (this.persistenceData?[mpid]?.fst) {
```

rmi22186: or maybe could it be as simple as 
```
return this.persistenceData?[mpid]?.fst || null
```
?

src/store.ts
```diff
@@ -453,6 +458,83 @@ export default function Store(
...
+    };
+
+    this.setFirstSeenTime = (mpid: MPID, _time?: number) => {
+        if (!mpid) {
+            return;
+        }
+
+        const time = _time || new Date().getTime();
+
+        if (this.persistenceData) {
```
rmi22186, line 485: Do we ever have to check if `this.persistenceData` doesn't exist if there will always be persisetenceData [as part of the default store creation](https://github.com/mParticle/mparticle-web-sdk/pull/855/files#diff-2717c7fb31c2070dd96f07394997809bc9be7f3fff03ab638db696cc480f39b6R225)?

rmi22186: i'm not opposed to playing it safe.  i thought TS might help us here though.

src/store.ts
```diff
@@ -453,6 +458,83 @@ export default function Store(
...
+            }
+        }
+    };
+
+    this.getLastSeenTime = (mpid: MPID) => {
+        if (!mpid) {
+            return null;
+        }
+
+        // TODO: We should store the current user in the Store's mpid rather than
```
rmi22186, line 501: I like this Idea - it is probably done the other way as a "just in case", but if we can ensure it is also always on the store, then this could simply things.

src/store.ts
```diff
@@ -453,6 +458,83 @@ export default function Store(
...
+        } else {
+            if (
+                this.persistenceData &&
+                this.persistenceData[mpid] &&
+                this.persistenceData[mpid].lst
+            ) {
+                return this.persistenceData[mpid].lst;
+            } else {
+                return null;
+            }
```
rmi22186, line 517:
```suggestion
            return this.persistenceData?[mpid]?.lst? || null;
```

test/src/tests-store.ts
```diff
@@ -288,6 +293,202 @@ describe('Store', () => {
         });
     });
 
+    describe('#getFirstSeenTime', () => {
+        it('should return the firstSeenTime from the store', () => {
+            debugger;
```
rmi22186, line 298:
```suggestion
```

test/src/tests-store.ts
```diff
@@ -288,6 +293,202 @@ describe('Store', () => {
...
+            );
+
+            store.setFirstSeenTime(testMPID, 54321);
+            const fromPersistence = window.mParticle
+                .getInstance()
+                ._Persistence.getPersistence();
+
+            expect(fromPersistence[testMPID]).to.be.ok;
+            expect(fromPersistence[testMPID].fst).to.be.ok;
+            expect(fromPersistence[testMPID].fst).to.equal(54321);
```
rmi22186, line 380: If the test is to see that it is overriding, maybe we test that the sid is null?  As opposed to the store extended what is currently in persistence?  It's a weird test right now but technically that's "overriding" what is in persistence.

rmi22186: or I realize you can set an sid of `foo-sid` directly on the store before calling setFirstSeenTime and then check that the sid is now `foo-sid`

test/src/tests-store.ts
```diff
@@ -288,6 +293,202 @@ describe('Store', () => {
...
+        });
+    });
+
+    describe('#getLastSeenTime', () => {
+        it('should return the lastSeenTime from the store', () => {
+            debugger;
```
rmi22186, line 386:
```suggestion
```

test/src/tests-store.ts
```diff
@@ -288,6 +293,202 @@ describe('Store', () => {
...
+            );
+
+            store.setLastSeenTime(testMPID, 54321);
+            const fromPersistence = window.mParticle
+                .getInstance()
+                ._Persistence.getPersistence();
+
+            expect(fromPersistence[testMPID]).to.be.ok;
+            expect(fromPersistence[testMPID].lst).to.be.ok;
+            expect(fromPersistence[testMPID].lst).to.equal(54321);
```
rmi22186, line 488: same comment as above about testing that what is on the store is overriding persistence.

rmi22186 (changes requested):
Looking good!  a few comments...

src/store.ts
alexs-mparticle: I think I tried this in a previous revision and it was giving me inconsistent results. My thought was for this PR, I would copy the existing code as is, and then optimize it later.

src/store.ts
alexs-mparticle: As mentioned above, I'm copying over the existing functions as is (or close to as is) without optimization to keep the changes as small as possible. I think there's a few places where we can optimize or DRY things up after we port overt he functionality, but that's out of scope IMO.

src/store.ts
alexs-mparticle: Yeah, I think I should create a ticket for this so we can decouple Identity from the Store as a next pass.

src/store.ts
alexs-mparticle: Same comment as above. Your suggestion transpiles differently than the existing if/else check and I would like to keep the changes smaller in this pass.

src/store.ts
rmi22186: SGTM

src/store.ts
rmi22186: SGTM

test/src/tests-store.ts
```diff
@@ -248,6 +253,242 @@ describe('Store', () => {
...
+
+            expect(fromPersistence[testMPID]).to.be.ok;
+            expect(fromPersistence[testMPID].fst).to.be.ok;
+            expect(fromPersistence[testMPID].fst).to.equal(12345);
+        });
+
+        it('should override persistence with store values', () => {
+            const persistenceValue = JSON.stringify({
+                gs: {
+                    sid: 'sid',
```
alexs-mparticle, line 357: Note to self: Can sid and les be removed?

test/src/tests-store.ts
```diff
@@ -248,6 +253,242 @@ describe('Store', () => {
...
+                    les: new Date().getTime(),
+                },
+                testMPID: {
+                    fst: 12345,
+                },
+                cu: testMPID,
```
alexs-mparticle, line 363: Note to self: Can this be removed?

test/src/tests-store.ts
```diff
@@ -248,6 +253,242 @@ describe('Store', () => {
...
+                ._Persistence.getPersistence();
+
+            expect(fromPersistence[testMPID]).to.be.ok;
+            expect(fromPersistence[testMPID].lst).to.be.ok;
+            expect(fromPersistence[testMPID].lst).to.equal(12345);
+        });
+
+        it('should override persistence with store values', () => {
+            const persistenceValue = JSON.stringify({
+                gs: {
```
alexs-mparticle, line 464: Note to self: Remove gs and cu

test/src/tests-store.ts
```diff
@@ -378,6 +383,190 @@ describe('Store', () => {
...
+        it('should return null if no lastSeenTime is found', () => {
+            const store: IStore = new Store(
+                sampleConfig,
+                window.mParticle.getInstance()
+            );
+
+            expect(store.getLastSeenTime(testMPID)).to.equal(null);
+        });
+
+        it('should return null if mpid is null', () => {
```
rmi22186, line 491: to be consistent with tests, can this test be added to `getFIrstSeenTime` too?

rmi22186 (changes requested):
tiny nit otherwise lgtm

test/src/tests-store.ts
alexs-mparticle: Good call out

rmi22186 (approved):

PR #864, opened 2024-04-08
refactor: Refactor ParseIdentityResponse Workflow

alexs-mparticle:
## Instructions

 ## Summary
 - Identify potential refactor points within the Identity Module

 ## Testing Plan
 - Smoke test an anonymous user session in a sample app, then attempt to login and log out

 - Closes https://go.mparticle.com/work/SQDSDKS-6310

### DIFF
--- a/src/identity.js
+++ b/src/identity.js
@@ -6,32 +6,27 @@ import {
     tryCacheIdentity,
 } from './identity-utils';
 import AudienceManager from './audienceManager';
-
-const { Messages, HTTPCodes, FeatureFlags } = Constants;
+const { Messages, HTTPCodes, FeatureFlags, IdentityMethods } = Constants;
 const { ErrorMessages } = Messages;
-
-const { Identify, Modify, Login, Logout } = Constants.IdentityMethods;
+const { CacheIdentity } = FeatureFlags;
+const { Identify, Modify, Login, Logout } = IdentityMethods;
+import {
+    generateDeprecationMessage,
+    isEmpty,
+    isFunction,
+    isObject,
+} from './utils';
+import { didUserChange } from './user-utils';
+import { getNewIdentitiesByName } from './type-utils';
 
 export default function Identity(mpInstance) {
+    const { getFeatureFlag, extend } = mpInstance._Helpers;
+
     var self = this;
     this.idCache = null;
     this.audienceManager = null;
 
-    this.checkIdentitySwap = function(
-        previousMPID,
-        currentMPID,
-        currentSessionMPIDs
-    ) {
-        if (previousMPID && currentMPID && previousMPID !== currentMPID) {
-            var persistence = mpInstance._Persistence.getPersistence();
-            if (persistence) {
-                persistence.cu = currentMPID;
-                persistence.gs.csm = currentSessionMPIDs;
-                mpInstance._Persistence.savePersistence(persistence);
-            }
-        }
-    };
-
+    // https://go.mparticle.com/work/SQDSDKS-6353
     this.IdentityRequest = {
         preProcessIdentityRequest: function(identityApiData, callback, method) {
             mpInstance.Logger.verbose(
@@ -135,9 +130,9 @@ export default function Identity(mpInstance) {
             var key;
             if (
                 newIdentities &&
-                mpInstance._Helpers.isObject(newIdentities) &&
+                isObject(newIdentities) &&
                 previousIdentities &&
-                mpInstance._Helpers.isObject(previousIdentities)
+                isObject(previousIdentities)
             ) {
                 for (key in newIdentities) {
                     identityChanges.push({
@@ -154,11 +149,7 @@ export default function Identity(mpInstance) {
         // takes 2 UI objects keyed by name, combines them, returns them keyed by type
         combineUserIdentities: function(previousUIByName, newUIByName) {
             var combinedUIByType = {};
-
-            var combinedUIByName = mpInstance._Helpers.extend(
-                previousUIByName,
-                newUIByName
-            );
+            var combinedUIByName = extend({}, previousUIByName, newUIByName);
 
             for (var key in combinedUIByName) {
                 var type = Types.IdentityType.getIdentityType(key);
@@ -232,6 +223,7 @@ export default function Identity(mpInstance) {
          * @param {Function} [callback] A callback function that is called when the identify request completes
          */
         identify: function(identityApiData, callback) {
+            // https://go.mparticle.com/work/SQDSDKS-6337
             var mpid,
                 currentUser = mpInstance.Identity.getCurrentUser(),
                 preProcessResult = mpInstance._Identity.IdentityRequest.preProcessIdentityRequest(
@@ -325,6 +317,7 @@ export default function Identity(mpInstance) {
          * @param {Function} [callback] A callback function that is called when the logout request completes
          */
         logout: function(identityApiData, callback) {
+            // https://go.mparticle.com/work/SQDSDKS-6337
             var mpid,
                 currentUser = mpInstance.Identity.getCurrentUser(),
                 preProcessResult = mpInstance._Identity.IdentityRequest.preProcessIdentityRequest(
@@ -414,6 +407,7 @@ export default function Identity(mpInstance) {
          * @param {Function} [callback] A callback function that is called when the login request completes
          */
         login: function(identityApiData, callback) {
+            // https://go.mparticle.com/work/SQDSDKS-6337
             var mpid,
                 currentUser = mpInstance.Identity.getCurrentUser(),
                 preProcessResult = mpInstance._Identity.IdentityRequest.preProcessIdentityRequest(
@@ -509,6 +503,7 @@ export default function Identity(mpInstance) {
          * @param {Function} [callback] A callback function that is called when the modify request completes
          */
         modify: function(identityApiData, callback) {
+            // https://go.mparticle.com/work/SQDSDKS-6337
             var mpid,
                 currentUser = mpInstance.Identity.getCurrentUser(),
                 preProcessResult = mpInstance._Identity.IdentityRequest.preProcessIdentityRequest(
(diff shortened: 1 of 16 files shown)


src/identity.js
```diff
@@ -1899,4 +1803,62 @@ export default function Identity(mpInstance) {
...
+        // https://go.mparticle.com/work/SQDSDKS-6036
+        mpInstance._Forwarders.setForwarderUserIdentities(
+            user.getUserIdentities().userIdentities
+        );
+        mpInstance._Forwarders.setForwarderOnIdentityComplete(user, method);
+        mpInstance._Forwarders.setForwarderOnUserIdentified(user);
+    };
+}
+
+function didUserChange(prevUser, newUser) {
```
rmi22186, line 1826: let's export and test this function

src/identity.js
```diff
@@ -1899,4 +1803,62 @@ export default function Identity(mpInstance) {
...
+    return (
+        !prevUser ||
+        newUser.getMPID() !== prevUser.getMPID() ||
+        prevUser.isLoggedIn() !== newUser.isLoggedIn()
+    );
+}
+
+// https://go.mparticle.com/work/SQDSDKS-6359
+function tryOnUserAlias(previousUser, newUser, identityApiData, logger) {
```
rmi22186, line 1835: I don't think this needs to be exported and tested since we are going to deprecate

src/identity.js
```diff
@@ -1786,27 +1690,27 @@ export default function Identity(mpInstance) {
 
     this.sendUserIdentityChangeEvent = function(
         newUserIdentities,
-        method,
         mpid,
+
+        // Mislabeled, could be previous or incoming
```
rmi22186, line 1695: Let's remove this comment and update the variable name to something more accurate

src/identity.js
```diff
@@ -1786,27 +1690,27 @@ export default function Identity(mpInstance) {
...
-        if (!mpid) {
-            if (method !== Modify) {
-                return;
-            }
-        }
-
-        currentUserInMemory = this.IdentityAPI.getUser(mpid);
+        // What's the difference between all of our getuser methods?
+        // We may not need to reaquire the user since we know the mpid
+        // WE already have this user from the identityApiResult
```
rmi22186, line 1700: let's make a jira ticket and remove this comment since this is an official PR and not a draft now.

src/identity.js
rmi22186: I'd also be open to moving a bunch of these newly created functions to a TS module, then importing them here to put us in a better position for next gen.

src/identity.js
```diff
@@ -1824,7 +1728,7 @@ export default function Identity(mpInstance) {
         identityType,
         newIdentity,
         oldIdentity,
-        newCreatedThisBatch,
+        newCreatedThisBatch, // New Identity Type introduced within the batch
```
rmi22186, line 1731: Rather than having the comment, maybe we can rename this variable to: `wasIdentityCreatedThisBatch`.  I don't really like that either, but it's more informative and supplants the need for a comment.

src/utils.ts
```diff
@@ -227,6 +227,9 @@ const mergeObjects = <T extends object>(...objects: T[]): T => {
     return Object.assign({}, ...objects);
 };
 
+const moveToElementEnd = <T>(array: T[], index: number): T[] =>
```
rmi22186, line 230:
```suggestion
const moveElementToEnd = <T>(array: T[], index: number): T[] =>
```

test/src/tests-store.ts
```diff
@@ -254,6 +254,60 @@ describe('Store', () => {
...
+            store.addMpidToSessionHistory('oldmpid');
+
+            expect(store.currentSessionMPIDs).to.deep.equal([
+                'newermpid',
+                'recentmpid',
+                'oldmpid',
+            ]);
+        });
+
+        it('should move an mpid to the end of the currentSessionMPIDs list if it is a previous mpid', () => {
```
rmi22186, line 293:
```suggestion
        it('should move an mpid to the end of the currentSessionMPIDs list if it is also the previous mpid', () => {
```

src/store.ts
```diff
@@ -461,6 +467,27 @@ export default function Store(
...
+        const indexOfMPID = this.currentSessionMPIDs.indexOf(mpid);
+
+        if (
+            // QUESTION: Is sessionId necessary here?
+            this.sessionId &&
+            mpid &&
+            previousMPID !== mpid &&
+            indexOfMPID < 0
+        ) {
+            this.currentSessionMPIDs.push(mpid);
```
rmi22186, line 480: to make this function more readable in that it will only ever hit one of these, maybe adding a return here would help

```suggestion
            this.currentSessionMPIDs.push(mpid);
            return;
```

test/src/tests-utils.ts
```diff
@@ -420,4 +421,12 @@ describe('Utils', () => {
             expect(merged.foo).to.equal('not-bar');
         });
     });
+
+    describe("#moveToElementEnd", () => {
```
rmi22186, line 425: since mpids are strings, can you instead do this test with strings?

rmi22186 (changes requested):
this is a great refactor! a few comments

src/identity.js
```diff
@@ -1567,80 +1550,44 @@ export default function Identity(mpInstance) {
...
-
-                    var incomingMpidUIByName = incomingUser
-                        ? incomingUser.getUserIdentities().userIdentities
-                        : {};
-
-                    var incomingMpidUIByNameCopy = mpInstance._Helpers.extend(
-                        {},
-                        incomingMpidUIByName
-                    );
                     mpInstance.Logger.verbose(
```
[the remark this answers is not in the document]
alexs-mparticle: I agree that 2 is a better option

src/identity.js
alexs-mparticle: I think my shorthand for booleans is to prefix with `is` so I'm going to go with `isIdentityTypeNewToBatch`

src/identity.js
```diff
@@ -1899,4 +1803,62 @@ export default function Identity(mpInstance) {
...
+        } catch (e) {
+            logger.error(
+                'There was an error with your onUserAlias function - ' + e
+            );
+        }
+    }
+}
+
+// QUESTION?: This process relies on combineUserIdentities but feels like it's
+// the reverse of that process. Can we combine them?
```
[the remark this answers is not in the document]
alexs-mparticle: I think when I wrote this comment, I saw that both of the processes use the IdentityType methods and assumed they were the reverse of each other.

src/sdkRuntimeModels.ts
```diff
@@ -306,10 +306,11 @@ export interface SDKConfigApi {
...
 export interface MParticleUser {
-    getMPID(): string;
-    getConsentState(): SDKConsentState;
-    getAllUserAttributes(): any; // FIXME;
-    getUserIdentities(): IdentityApiData; // FIXME: Is this correct?
+    getMPID?(): string;
+    getConsentState?(): SDKConsentState;
+    getAllUserAttributes?(): any; // FIXME;
+    getUserIdentities?(): IdentityApiData; // FIXME: Is this correct?
+    isLoggedIn?(): boolean;
```
rmi22186, line 313: Why are these optional now?  

Also - can we turn the `FIXME`s into work items and link to them here?

src/user-utils.ts
```diff
@@ -0,0 +1,12 @@
+import { MParticleUser } from './sdkRuntimeModels';
+
+export function didUserChange(
+    previousUser: MParticleUser,
+    newUser: MParticleUser
+) {
```
rmi22186, line 6:
```suggestion
export function didUserChange(
    previousUser: MParticleUser,
    newUser: MParticleUser
): boolean {
```

test/jest/user-utils.spec.ts
```diff
@@ -0,0 +1,35 @@
+import { didUserChange } from '../../src/user-utils';
+
+describe('user-utils', () => {
+    describe('#didUserChange', () => {
+        it('returns true if previousUser is null', () => {
+            expect(didUserChange(null, null)).toBeTruthy();
+        });
+
+        it('returns true if previousUser and newUser have different mpids', () => {
```
rmi22186, line 9: please add another test to return false if the user's new mpid and old mpid are the same

rmi22186 (changes requested):
small nits. otherwise this is a great improvement!

src/sdkRuntimeModels.ts
alexs-mparticle: They're optional so that the mockBatchCreator in our tests doesn't complain. We already have a sustained eng task to refactor that but it's lower in urgency.

alexs-mparticle: They're optional so that the mockBatchCreator in our tests doesn't complain. We already have a sustained eng task to refactor that but it's lower in urgency.

rmi22186: sgtm

src/sdkRuntimeModels.ts
```diff
@@ -303,13 +303,12 @@ export interface SDKConfigApi {
...
 
-// FIXME: Resolve with User in @types/mparticle-web-sdk
-//        https://go.mparticle.com/work/SQDSDKS-5033
+// https://go.mparticle.com/work/SQDSDKS-5033
 export interface MParticleUser {
     getMPID?(): string;
     getConsentState?(): SDKConsentState;
-    getAllUserAttributes?(): any; // FIXME;
-    getUserIdentities?(): IdentityApiData; // FIXME: Is this correct?
+    getAllUserAttributes?(): UserAttributes; // FIXME;
```
rmi22186, line 310: remove FIXME

rmi22186 (changes requested):
please resolve this build issue:
https://github.com/mParticle/mparticle-web-sdk/actions/runs/8755335851/job/24029176083?pr=864

rmi22186 (approved):

rmi22186:
approved!  just be sure that when you squash and commit, you have a semantic release-approved commit message

PR #865, opened 2024-04-09
fix: Implement audience API feature flag

rmi22186:
## Instructions

 ## Summary
 - This PR implements a feature flag for audience API to work

 ## Testing Plan

 - Closes https://go.mparticle.com/work/SQDSDKS-6318

### DIFF
--- a/src/identity.js
+++ b/src/identity.js
@@ -7,7 +7,7 @@ import {
 } from './identity-utils';
 import AudienceManager from './audienceManager';
 
-const { Messages, HTTPCodes } = Constants;
+const { Messages, HTTPCodes, FeatureFlags } = Constants;
 const { ErrorMessages } = Messages;
 
 const { Identify, Modify, Login, Logout } = Constants.IdentityMethods;
@@ -1276,7 +1276,7 @@ export default function Identity(mpInstance) {
                 // user audience API is feature flagged
                 if (
                     !mpInstance._Helpers.getFeatureFlag(
-                        Constants.FeatureFlags.AudienceAPI
+                        FeatureFlags.AudienceAPI
                     )
                 ) {
                     mpInstance.Logger.error(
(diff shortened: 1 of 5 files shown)


src/identity.js
```diff
@@ -1273,6 +1273,17 @@ export default function Identity(mpInstance) {
...
              */
             getUserAudiences: function(callback) {
+                // user audience API is feature flagged
+                if (
+                    !mpInstance._Helpers.getFeatureFlag(
+                        Constants.FeatureFlags.AudienceAPI
+                    )
+                ) {
+                    mpInstance.Logger.error(
+                        Constants.Messages.ErrorMessages.AudienceAPINotEnabled
```
alexs-mparticle, line 1283: Nit: In the top of the file, we destructure these nested constants, so this can be broken out like so:

```
// Top of file:
const ErrorMessages = Constants.Messages;

// Then here:
mpInstance.Logger.error(ErrorMessages.AudienceAPINotEnabled);
```

alexs-mparticle (changes requested):

einsteinx2 (approved):
LGTM

src/identity.js
```diff
@@ -1273,6 +1273,17 @@ export default function Identity(mpInstance) {
              * @param {Function} [callback] A callback function that is invoked when the user audience request completes
              */
             getUserAudiences: function(callback) {
+                // user audience API is feature flagged
+                if (
+                    !mpInstance._Helpers.getFeatureFlag(
+                        Constants.FeatureFlags.AudienceAPI
```
alexs-mparticle, line 1279:
```suggestion
                        FeatureFlags.AudienceAPI
```

src/identity.js
```diff
@@ -7,8 +7,8 @@ import {
 } from './identity-utils';
 import AudienceManager from './audienceManager';
 
-var Messages = Constants.Messages,
-    HTTPCodes = Constants.HTTPCodes;
+const { Messages, HTTPCodes } = Constants;
```
alexs-mparticle, line 10:
```suggestion
const { Messages, HTTPCodes, FeatureFlags } = Constants;
```

alexs-mparticle (changes requested):

alexs-mparticle (approved):

PR #866, opened 2024-04-10
feat: Implement fetching user audiences

rmi22186:
## Instructions

 ## Summary
This contains commits that were part of previously approved PRs.  See each individual commits PR# for details.

 ## Testing Plan

 - Closes https://go.mparticle.com/work/SQDSDKS-5994


src/apiClient.ts
```diff
@@ -40,7 +40,7 @@ export default function APIClient(
         if (!this.uploader) {
             const millis: number = parseNumber(mpInstance._Helpers.getFeatureFlag(
                 Constants.FeatureFlags.EventBatchingIntervalMillis
-            ));
+            ) as string);
```
alexs-mparticle, line 43: Can you add this so we can fix up the feature flags later?

https://go.mparticle.com/work/SQDSDKS-6317

src/audienceManager.ts
```diff
@@ -0,0 +1,98 @@
...
+            method: 'GET',
+            headers: {
+                Accept: '*/*',
+            },
+        };
+        const audienceURLWithMPID = `${this.url}?mpid=${mpid}`;
+        let userAudiencePromise: Response;
+
+        try {
+             userAudiencePromise = await this.userAudienceAPI.upload(
```
alexs-mparticle, line 51: I think you can just add `const userAudiencePromise ...` here since you're not using this outside of the try/catch block.

src/audienceManager.ts
```diff
@@ -0,0 +1,98 @@
...
+
+                const userAudienceMembershipsServerResponse: IAudienceMembershipsServerResponse = await userAudiencePromise.json();
+                const parsedUserAudienceMemberships: IAudienceMemberships = {
+                    currentAudienceMemberships: userAudienceMembershipsServerResponse?.audience_memberships
+                }
+
+                try {
+                    callback(parsedUserAudienceMemberships);
+                } catch(e) {
+                    this.logger.error('Error invoking callback on user audience response.');
```
alexs-mparticle, line 70: You're doing a lot of custom error loggers in this function, but you're also nesting a try/catch block when it's unnecessary.

Each of these loggers should instead do a `throw new Error()` and then you can use your catch to pass it to our logger.

It would make things cleaner.

src/audienceManager.ts
```diff
@@ -0,0 +1,98 @@
...
+        const audienceURLWithMPID = `${this.url}?mpid=${mpid}`;
+        let userAudiencePromise: Response;
+
+        try {
+             userAudiencePromise = await this.userAudienceAPI.upload(
+                 fetchPayload,
+                 audienceURLWithMPID
+             );
+
+            if (
```
alexs-mparticle, line 56: Is there a reason we're not using a `switch`?

src/batchUploader.ts
```diff
@@ -10,6 +10,7 @@ import { convertEvents } from './sdkToEventsApiConverter';
 import Types from './types';
 import { getRampNumber, isEmpty } from './utils';
 import { SessionStorageVault, LocalStorageVault } from './vault';
+import { AsyncUploader, FetchUploader, XHRUploader, fetchPayload } from './uploaders';
```
alexs-mparticle, line 13: I think the linter should have split this line up for you.

src/batchUploader.ts
```diff
@@ -108,11 +109,11 @@ export class BatchUploader {
             _Store: { deviceId },
         } = this.mpInstance;
 
-        const offlineStorageFeatureFlagValue = getFeatureFlag(
+        const offlineStorageFeatureFlagValue: string = getFeatureFlag(
```
alexs-mparticle, line 112: Let's add this to sustained engineering.

```suggestion
        // https://go.mparticle.com/work/SQDSDKS-6317
        const offlineStorageFeatureFlagValue: string = getFeatureFlag(
```

src/sdkRuntimeModels.ts
```diff
@@ -246,19 +246,26 @@ export interface SDKIdentityApi {
...
     canLog?(): boolean;
-    createServiceUrl(arg0: string, arg1: string): void;
+    createServiceUrl(url: string, devToken?: string): void;
     createXHR?(cb: () => void): XMLHttpRequest;
     extend?(...args: any[]);
     parseNumber?(value: string | number): number;
     generateUniqueId();
     generateHash?(value: string): string;
-    getFeatureFlag?(feature: string); // TODO: Feature Constants should be converted to enum
+    getFeatureFlag?(feature: string): boolean | string; // TODO: Feature Constants should be converted to enum
```
alexs-mparticle, line 255:
```suggestion
    // https://go.mparticle.com/work/SQDSDKS-6317
    getFeatureFlag?(feature: string): boolean | string; // TODO: Feature Constants should be converted to enum
```

src/sdkRuntimeModels.ts
```diff
@@ -246,19 +246,26 @@ export interface SDKIdentityApi {
...
     isDelayedByIntegration?(
         delayedIntegrations: Dictionary<boolean>,
         timeoutStart: number,
         now: number
     ): boolean;
     isObject?(item: any);
+    invokeCallback?(
+        cb: ()=> void,
```
alexs-mparticle, line 263: I think we have a callback type you can use.

Also, can we use `callback` instead of `cb`?

src/sdkRuntimeModels.ts
```diff
@@ -246,19 +246,26 @@ export interface SDKIdentityApi {
...
+        code: string,
+        body: string,
+        mParticleUser?: MParticleUser,
+        previousMpid?: string
```
alexs-mparticle, line 267:
```suggestion
        previousMpid?: MPID
```

test/src/tests-audience-manager.ts
```diff
@@ -0,0 +1,172 @@
...
+        const audienceManager = new AudienceManager(
+            Constants.DefaultBaseUrls.userAudienceUrl,
+            apiKey,
+            newLogger,
+        );
+
+        expect(audienceManager.logger).to.be.ok;
+        expect(audienceManager.url).to.equal(userAudienceUrl);
+        expect(audienceManager.userAudienceAPI).to.be.ok;
+    });
```
alexs-mparticle, line 61: This is a personal style choice, but I found that wrapping init logic in its own describe block makes it easier to collapse things in the IDE, and gives us room if we ever need to add new init tests in the future.

```suggestion
    describe('initialization', () => {
        it('should have proper properties on AudienceManager', () => {
            const newLogger: SDKLoggerApi = new Logger(window.mParticle.config);
            const audienceManager = new AudienceManager(
                Constants.DefaultBaseUrls.userAudienceUrl,
                apiKey,
                newLogger,
            );

            expect(audienceManager.logger).to.be.ok;
            expect(audienceManager.url).to.equal(userAudienceUrl);
            expect(audienceManager.userAudienceAPI).to.be.ok;
        });
    });
```

test/src/tests-identity.js
```diff
@@ -3359,6 +3359,80 @@ describe('identity', function() {
         });
     });
 
+    describe('user audiences', function() {
```
alexs-mparticle, line 3362: Is there a reason we are putting these in the identity tests rather than the audience manager tests? They seem to only be testing if feature flag is on or off, so I don't see why this would be in identity?

If anything, maybe we make a feature flags tests file?

alexs-mparticle (changes requested):

src/audienceManager.ts
rmi22186: I actually did this at first and it wasn't working as expected, according to [this](https://stackoverflow.com/questions/41480811/switch-statement-with-numeric-values), number values don't work in a switch statement, and we'd have to do something like 
```
switch (true) {
   case userAudiencePromise.status >= 200 &&
                userAudiencePromise.status < 300
}
```
I wasn't a fan of that.  However, since http logic is done here and in batchuploader, maybe we can decide what the preferred option is as a follow up and do it to both in a single ticket?

src/batchUploader.ts
rmi22186: We'll need to get our linters aligned one day.  maybe a sustained engineering session.  mine seems to work with js but not ts, and freak out sometimes.  i'll update for now though.

alexs-mparticle: I already have one on the books: https://go.mparticle.com/work/SQDSDKS-5147

alexs-mparticle (changes requested):

alexs-mparticle (approved):

PR #868, opened 2024-04-15
refactor: Migrate User Attribute  Persistence Methods

alexs-mparticle:
## Instructions

 ## Summary
 - Migrate User Attribute Setter and Getter to Store

 ## Testing Plan

 - Closes https://go.mparticle.com/work/SQDSDKS-6311

### DIFF
--- a/src/persistence.interfaces.ts
+++ b/src/persistence.interfaces.ts
@@ -125,7 +125,6 @@ export interface IPersistence {
     getDomain(doc: string, locationHostname: string): string;
     getCartProducts(mpid: MPID): Product[];
     setCartProducts(allProducts: Product[]): void;
-    getUserIdentities(mpid: MPID): UserIdentities;
     saveUserCookieSyncDatesToPersistence(mpid: MPID, csd: CookieSyncDate): void;
     saveUserConsentStateToCookies(mpid, consentState: ConsentState): void;
     savePersistence(persistance: IPersistenceMinified): void;
--- a/src/persistence.interfaces.ts
+++ b/src/persistence.interfaces.ts
@@ -124,9 +124,9 @@ export interface IPersistence {
     decodePersistence(persistenceString: string): string;
     getCookieDomain(): string;
     getDomain(doc: string, locationHostname: string): string;
-    getAllUserAttributes(mpid: MPID): AllUserAttributes;
     getCartProducts(mpid: MPID): Product[];
     setCartProducts(allProducts: Product[]): void;
+    getUserIdentities(mpid: MPID): UserIdentities;
     saveUserAttributesToPersistence(
         mpid: MPID,
         userAttributes: UserAttributes
--- a/src/persistence.interfaces.ts
+++ b/src/persistence.interfaces.ts
@@ -22,7 +22,6 @@ export interface iForwardingStatsBatches {
     forwardingStatsEventQueue: ForwardingStatsData[];
 }
 
-// TODO: Migrate this to @types/mparticle__web-sdk
 //       https://go.mparticle.com/work/SQDSDKS-5196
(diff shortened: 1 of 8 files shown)


src/store.ts
```diff
@@ -168,6 +173,8 @@ export interface IStore {
 
     getDeviceId?(): string;
     setDeviceId?(deviceId: string): void;
+    getUserAttributes?(mpid: MPID): UserAttributes;
+    setUserAttributes?(mpid: MPID, attributes: UserAttributes): void;
```
rmi22186, line 177: There is both an API for `setUserAttribute`, as well as `setUserAttributes`.  The latter is included here, but not the former.  Should the former also be included here?

rmi22186:
only other thing is the PR is potentially missing the invocations within the SDK to use these methods. I think that should be included in the PR

src/store.ts
alexs-mparticle: `setUserAttribute` and `setUserAttributes` are part of the Identity API, but internally, they all write to persistence the same (ish) way. Fort the Store, we only store user attributes as `ua`, so we probably only need a single setter/getter.

src/store.ts
```diff
@@ -613,6 +620,20 @@ export default function Store(
         );
     };
 
+    this.getUserAttributes = (mpid: MPID): UserAttributes =>
+        this.persistenceData &&
+        this.persistenceData[mpid] &&
+        this.persistenceData[mpid].ua
+            ? this.persistenceData[mpid].ua
+            : {};
```
alexs-mparticle, line 628: Note to self: use `_getPersistence` here

src/store.ts
```diff
@@ -613,6 +618,16 @@ export default function Store(
         );
     };
 
+    this.getUserAttributes = (mpid: MPID): UserAttributes =>
+        this._getFromPersistence(mpid, 'ua') || {};
```
rmi22186, line 622: You should also pass the <type> like line 542-543 above.

rmi22186 (changes requested):
minor nit, otherwise lgtm!

src/store.ts
alexs-mparticle: That's a good call out. I would like to address it as part of a separate cleanup PR since I think we missed a few of these during the individual PRs.

src/persistence.interfaces.ts
```diff
@@ -124,13 +123,9 @@ export interface IPersistence {
...
     getCookieDomain(): string;
     getDomain(doc: string, locationHostname: string): string;
-    getAllUserAttributes(mpid: MPID): AllUserAttributes;
     getCartProducts(mpid: MPID): Product[];
     setCartProducts(allProducts: Product[]): void;
-    saveUserAttributesToPersistence(
-        mpid: MPID,
-        userAttributes: UserAttributes
-    ): void;
+    getUserIdentities(mpid: MPID): UserIdentities;
```
rmi22186, line 128: it looks like this has been added back in after removing it from https://github.com/mParticle/mparticle-web-sdk/pull/869/files.  probably a rebasing issue.

src/persistence.js
```diff
@@ -882,11 +882,11 @@ export default function _Persistence(mpInstance) {
         return '';
     };
 
-    this.getAllUserAttributes = function(mpid) {
+    this.getUserIdentities = function(mpid) {
```
rmi22186, line 885: it looks like this has been added back in after removing it from https://github.com/mParticle/mparticle-web-sdk/pull/869/files.  probably a rebasing issue.

rmi22186 (changes requested):
minor comment from rebasing.  otherwise lgtm

rmi22186 (approved):

PR #870, opened 2024-04-18
refactor: Migrate globalStorageAttribute getter

alexs-mparticle:
## Instructions

 ## Summary
 - Refactors `globalStorageAttribute` setter into a getter and cleans up generation of the `gs` attribute of Persistence.

 ## Testing Plan
 - Smoke test an anonymous user session in a sample app, then attempt to login and log out

 - Closes https://go.mparticle.com/work/SQDSDKS-6349

### DIFF
--- a/test/src/tests-store.ts
+++ b/test/src/tests-store.ts
@@ -1209,6 +1209,116 @@ describe('Store', () => {
         });
     });
 
+    describe('#getGlobalStorageAttributes', () => {
+        it('should return the global storage attributes from the store', () => {
+            const store: IStore = new Store(
+                sampleConfig,
+                window.mParticle.getInstance()
+            );
+
+            const dateLastEventSent = new Date();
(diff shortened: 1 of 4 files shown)


test/src/tests-store.ts
```diff
@@ -734,6 +734,116 @@ describe('Store', () => {
...
+            );
+            expect(actualGS.csm, 'current session mpids').to.deep.equal(
+                expectedGS.csm
+            );
+
+            // Tests to make sure we're not accidentally adding anything extra to the global storage attributes
+            expect(actualGS).to.deep.equal(expectedGS);
+        });
+
+        it('should return null if last event sent is null', () => {
```
rmi22186, line 807:
```suggestion
        it('should return null for les if last event sent is null', () => {
```

rmi22186 (approved):

PR #872, opened 2024-04-22
refactor: Add test cases for setting isLoggedIn value via persistence

alexs-mparticle:
## Instructions

 ## Summary
 - Adds extra tests around setting logged in state for persistence.

 ## Testing Plan
 - Run automated tests
 - Use a sample app to verify logged in/logged out state changes in persistence

 - Closes https://go.mparticle.com/work/SQDSDKS-6352

### DIFF
--- a/src/persistence.interfaces.ts
+++ b/src/persistence.interfaces.ts
@@ -44,7 +44,11 @@ export interface IGlobalStoreV2MinifiedKeys {
 export interface IPersistenceMinified extends Dictionary {
     cu: MPID; // Current User MPID
     gs: IGlobalStoreV2MinifiedKeys;
-    l: boolean; // IsLoggedIn - stored as 0 or 1 but returns a boolean
+
+    // Stored as 0 or 1 in device persistence but returned as a
+    // boolean when decoding from device persistence via
+    // _Persistence.getPersistence and _Persistence.decodePersistence
+    l: boolean; // IsLoggedIn
 
     // Persistence Minified can also store optional dictionaries with
     // an idex of MPID
--- a/src/persistence.interfaces.ts
+++ b/src/persistence.interfaces.ts
@@ -44,7 +44,7 @@ export interface IGlobalStoreV2MinifiedKeys {
 export interface IPersistenceMinified extends Dictionary {
     cu: MPID; // Current User MPID
     gs: IGlobalStoreV2MinifiedKeys;
-    l: boolean; // IsLoggedIn
+    l: boolean; // IsLoggedIn - stored as 0 or 1 but returns a boolean
 
     // Persistence Minified can also store optional dictionaries with
     // an idex of MPID
--- a/test/src/tests-persistence.ts
+++ b/test/src/tests-persistence.ts
@@ -20,6 +20,7 @@ import {
     IPersistenceMinified,
 } from '../../src/persistence.interfaces';
 import { ConsentState } from '@mparticle/web-sdk';
+import { MParticleUser } from '../../src/sdkRuntimeModels';
 
 const {
     findCookie,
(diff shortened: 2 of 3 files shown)


src/persistence.js
```diff
@@ -356,7 +356,7 @@ export default function _Persistence(mpInstance) {
         }
 
         if (!mpInstance._Store.SDKConfig.useCookieStorage) {
-            localStorageData.l = mpInstance._Store.isLoggedIn ? 1 : 0;
+            localStorageData.l = mpInstance._Store.isLoggedIn; // ? 1 : 0;
```
rmi22186, line 359: I don't think this should be changed.  We want to set it as a number in the `IPersistenceMinified`, but as a boolean on the store.

src/persistence.interfaces.ts
```diff
@@ -45,7 +45,7 @@ export interface IGlobalStoreV2MinifiedKeys {
 export interface IPersistenceMinified extends Dictionary {
     cu: MPID; // Current User MPID
     gs: IGlobalStoreV2MinifiedKeys;
-    l: boolean; // IsLoggedIn
+    l: boolean; // IsLoggedIn - stored as 0 or 1 but returns a boolean
```
rmi22186, line 48:
```suggestion
    l: number; // IsLoggedIn - stored as 0 or 1 but returns a boolean
```

src/persistence.js
rmi22186: here is a screenshot of a test app
image

src/persistence.js
```diff
@@ -513,7 +513,7 @@ export default function _Persistence(mpInstance) {
             cookies.cu = mpid;
         }
 
-        cookies.l = mpInstance._Store.isLoggedIn ? 1 : 0;
+        cookies.l = mpInstance._Store.isLoggedIn;
```
rmi22186, line 516: Also revert

test/src/tests-persistence.ts
```diff
@@ -1483,6 +1484,68 @@ describe('persistence', () => {
...
+        mParticle.init(apiKey, mParticle.config);
+        let user: MParticleUser = mParticle
+            .getInstance()
+            .Identity.getCurrentUser()
+        expect(user).to.be.ok;
+        expect(user.isLoggedIn()).to.be.false;
+
+        let localStorageData = mParticle.getInstance()._Persistence.getPersistence();
+
+        // The `l` property of Persistence is a boolean, but internally it is stored as a 0 or 1
```
rmi22186, line 1506: Can you be more specific with `internally`?

test/src/tests-store.ts
```diff
@@ -522,7 +522,7 @@ describe('Store', () => {
                     ssd: 8675309,
                 } as IGlobalStoreV2MinifiedKeys,
                 cu: 'mpid3',
-                l: false,
+                l: true,
```
rmi22186, line 525: Would this be 0 or 1 then?

rmi22186 (changes requested):
I think there is some question as to whether or not it's true/false of 0/1 in certain places. May be worth hopping on a call.

src/persistence.js
rmi22186: this changes what is in persistence actually.  it saves `true` instead of `1/0` and that's because the 1:0 translation should be happening inside of the encodePersistence, similar to what is happening when the key === `ie`. 

image

Please revert this, or add this to the encoding layer.

rmi22186 (changes requested):

test/src/tests-store.ts
rmi22186: any particular reason to change this at all from `false` to `true`?  i think the update in this test would be to look for the `l` on the persistence.  seems to not be testing for it at all, and that's why this test is still passing.

src/persistence.interfaces.ts
```diff
@@ -44,7 +44,7 @@ export interface IGlobalStoreV2MinifiedKeys {
...
```
rmi22186, line 47:
```suggestion
    l: boolean; // IsLoggedIn - stored as 0 or 1 in local storage/cookies but returned as a boolean when decoding from persistence
```

rmi22186 (approved):

PR #873, opened 2024-04-24
refactor: Migrate cookieSyncDates setters and getters to Store

alexs-mparticle:
## Instructions

 ## Summary
 - Migrate Cookie Sync Dates functionality from Persistence into Store

 ## Testing Plan
 - Run automated tests and test in a sample app set to use Cookies

 - Closes https://go.mparticle.com/work/SQDSDKS-6351

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -34,6 +34,7 @@ import {
 import { IMinifiedConsentJSONObject, SDKConsentState } from './consent';
 import { Kit, MPForwarder } from './forwarders.interfaces';
 import {
+    CookieSyncDate,
     IGlobalStoreV2MinifiedKeys,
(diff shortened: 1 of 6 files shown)


src/store.ts
```diff
@@ -34,6 +34,7 @@ import {
 import { IMinifiedConsentJSONObject, SDKConsentState } from './consent';
 import { Kit, MPForwarder } from './forwarders.interfaces';
 import {
+    CookieSyncDate,
```
rmi22186, line 37: Could you make this plural in the interfaces so it's `CookieSyncDates`?  Seems related and quick enough to not be _totally_ out of scope...but up to you.

rmi22186 (approved):

PR #876, opened 2024-04-26
refactor: Remove usages of Persistence First/Last Seen Times

alexs-mparticle:
## Instructions

 ## Summary
 -  Remove usages of Persistence First/Last Seen Times

 ## Testing Plan

 - Closes https://go.mparticle.com/work/SQDSDKS-6400


test/src/tests-identity.js
```diff
@@ -3084,6 +3086,46 @@ describe('identity', function() {
...
+            // We set the cookies because the init process will load cookies into
+            // our in-memory store, which we will check before and after the identify request
+            setCookie(workspaceCookieName, cookies, true);
+            mParticle.config.useCookieStorage = true;
+
+            mParticle.init(apiKey, mParticle.config);
+
+            expect(
+                mParticle.getInstance()._Store.getFirstSeenTime('current')
+            ).to.equal(null);
```
rmi22186, line 3117: Can you try the following?  Comment out `sid` and `les` from the cookies above ([line 3095-3096](https://github.com/mParticle/mparticle-web-sdk/pull/876/files#diff-f9b04257df2922fb4c4451e6ba6453eb1ab9dd5c51f3e148939dca7986bed66aR3115-R3117) in this commit)

Does the test in 3115 fail?  

I think it would. Because normally `init` will call `identify`, and the reason it is not calling it here, and you have to manually call it below this line, is because there is an `sid`, and/or `les`.  If you step through, I think it will skip the `identify` call because of that.

If the above is in fact the case, we should add a comment above `mParticle.init` to explain why it isn't called, to help future developers out.

test/src/tests-store.ts
```diff
@@ -1037,13 +1029,39 @@ describe('Store', () => {
...
+
+        it('should set firstSeenTime once', () => {
+            const store: IStore = new Store(
+                sampleConfig,
+                window.mParticle.getInstance()
+            );
+
+            store.persistenceData['previous-set-mpid'] = {
+                fst: 100
+            }
```
rmi22186, line 1052: is there a reason you are setting fst for `previous-set-mpid` this way directly on `store.persistenceData` instead of how you are setting it for `current-mpid` and `previous-mpid` below?

test/src/tests-store.ts
```diff
@@ -1147,6 +1166,29 @@ describe('Store', () => {
             expect(fromPersistence[testMPID].lst).to.be.ok;
             expect(fromPersistence[testMPID].lst).to.equal(54321);
         });
+
+        it('returns current time for the current user', () => {
```
rmi22186, line 1170:
```suggestion
        it('returns current time for lastSeenTime for the current user', () => {
```

test/src/tests-store.ts
```diff
@@ -1814,4 +1856,29 @@ describe('Store', () => {
...
+    describe('integration tests', () => {
+        describe('#firstSeenTime', () => {
+            it('should be null for users in storage without an lst value', done => {
+                const cookies = JSON.stringify({
+                    gs: {
+                        sid: 'lst Test',
+                        les: new Date().getTime(),
+                    },
+                    previous: {},
+                    cu: 'current',
```
rmi22186, line 1869: This applies to all the tests you are changing.  (And future updates). 

I'd say instead of giving the sample MPIDs here `previous` and `current`, we shold call it `PREVIOUS_MPID`/`previous_mpid` and `CURRENT_MPID`/`current_mpid`. This maes it more clear to someone who is reading this code that the key is an `MPID`.

test/src/tests-store.ts
```diff
@@ -1814,4 +1856,29 @@ describe('Store', () => {
...
+                window.mParticle.config.useCookieStorage = true;
+        
+                window.mParticle.init(apiKey, window.mParticle.config);
+                expect(
+                    window.mParticle.getInstance()._Store.getFirstSeenTime('previous')
+                ).to.equal(null);
+        
+                done();
+            });
+
```
rmi22186, line 1881:
```suggestion
```

rmi22186 (changes requested):

test/src/tests-store.ts
rmi22186: that sounds fine to me

test/src/tests-identity.js
```diff
@@ -3110,7 +3110,13 @@ describe('identity', function() {
             setCookie(workspaceCookieName, cookies, true);
             mParticle.config.useCookieStorage = true;
 
-            mParticle.init(apiKey, mParticle.config);
+
+            // As part of init, there is a call to Identity.Identify. However, in this test case,
```
rmi22186, line 3114: 👍

rmi22186 (approved):

PR #881, opened 2024-05-02
test: Add Test Coverage for Modify and Identity Requests

alexs-mparticle:
## Instructions

 ## Summary
 - During a recent refactor, #864 we introduced a bug with how we process `modify` requests, which revealed a gap in our test coverage. We have update some of our older tests to properly mock `modify` requests.

 ## Testing Plan
 - The draft commit #880 highlights these updated tests within the context of the bug to show how tests failing when the bug is introduced.

 - Closes https://go.mparticle.com/work/SQDSDKS-6403

### DIFF
--- a/test/src/tests-event-logging.js
+++ b/test/src/tests-event-logging.js
@@ -591,7 +591,7 @@ describe('event logging', function() {
     });
 
     it('should log identify event', function(done) {
-        mockServer.respondWith(urls.logout, [
+        mockServer.respondWith(urls.identify, [
             200,
             {},
             JSON.stringify({ mpid: testMPID, is_logged_in: false }),
--- a/test/src/tests-event-logging.js
+++ b/test/src/tests-event-logging.js
@@ -658,7 +658,14 @@ describe('event logging', function() {
         mockServer.respondWith(urls.modify, [
             200,
             {},
-            JSON.stringify({ mpid: testMPID, is_logged_in: false }),
+            JSON.stringify({
+                change_results: [
+                    {
+                        identity_type: 'email',
+                        modified_mpid: testMPID,
+                    },
+                ],
+            }),
         ]);
         mParticle.Identity.modify();
         const data = getIdentityEvent(mockServer.requests, 'modify');
--- a/test/src/tests-event-logging.js
+++ b/test/src/tests-event-logging.js
@@ -590,6 +590,27 @@ describe('event logging', function() {
         done();
     });
 
+    it('should log identify event', function(done) {
+        mockServer.respondWith(urls.logout, [
+            200,
+            {},
+            JSON.stringify({ mpid: testMPID, is_logged_in: false }),
+        ]);
+        mParticle.Identity.identify();
+        const data = getIdentityEvent(mockServer.requests, 'identify');
+        data.should.have.properties(
+            'client_sdk',
+            'environment',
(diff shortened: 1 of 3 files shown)


test/src/tests-event-logging.js
```diff
@@ -590,6 +590,27 @@ describe('event logging', function() {
         done();
     });
 
+    it('should log identify event', function(done) {
+        mockServer.respondWith(urls.logout, [
```
rmi22186, line 594: This should be `identify` and not `logout`

test/src/tests-identity.js
```diff
@@ -45,6 +46,179 @@ describe('identity', function() {
...
+            ]);
+
+            mParticle.Identity.identify({
+                userIdentities: {
+                    email: 'test@email.com',
+                },
+            });
+            const data = getIdentityEvent(mockServer.requests, 'identify');
+
+            console.log('data identify', data);
```
rmi22186, line 72: remove

test/src/tests-identity.js
```diff
@@ -1655,13 +1843,46 @@ describe('identity', function() {
...
+                validUserIdentities.forEach(function(goodIdentities) {
+                    mParticle.Identity[identityMethod](
+                        goodIdentities,
+                        callback
+                    );
+                    expect(
+                        result.httpCode,
+                        `valid ${identityMethod} httpCode`
+                    ).to.equal(200);
+                    console.warn(result.body);
```
rmi22186, line 1859: remove

test/src/tests-identity.js
```diff
@@ -1655,13 +1843,46 @@ describe('identity', function() {
...
+                    expect(
+                        result.body.mpid,
+                        `valid ${identityMethod} mpid `
+                    ).to.be.ok;
+                    expect(
+                        result.body.mpid,
+                        `valid ${identityMethod} mpid`
+                    ).to.equal(testMPID);
+                    result = null;
```
rmi22186, line 1868: add a comment above here saying that this needs to be set to null because the test is inside of a loop and needs to be reset before each test loop

test/src/tests-identity.js
```diff
@@ -1655,13 +1843,46 @@ describe('identity', function() {
...
+                    );
+                    expect(
+                        result.body.change_results[0].modified_mpid,
+                        `valid modify change_results modified_mpid`
+                    ).to.equal(testMPID);
+                    expect(
+                        result.body.change_results[0].identity_type,
+                        `valid modify change_results identity_type`
+                    ).to.be.ok;
+                });
```
rmi22186, line 1884: should you also reset `result=null` like above here?

rmi22186 (changes requested):
You missed this one.  https://github.com/mParticle/mparticle-web-sdk/pull/881/commits/9853e77609df8cf6a60c31d5b068b2981061c692#diff-9312ed43259c225c470943ef298f0e49ae5cbe99dc998f4ebc98d9d7747b7192R657 

Do a global search for `urls.modify` in the tests to ensure you got every occurrence/.

Looks like https://github.com/mParticle/mparticle-web-sdk/blob/9853e77609df8cf6a60c31d5b068b2981061c692/test/src/tests-identity.js#L1487-L1491 is also still the old incorrect signature

rmi22186 (approved):

PR #884, opened 2024-05-09
refactor: Migrate Identity Tests to TypeScript

alexs-mparticle:
## Instructions

 ## Summary
 - Migrates Unit and Integration test for Identity from plain JavaScript to TypeScript, along with new interfaces.

 ## Testing Plan
 - Run all build and automated tests to verify they pass correctly
 - Using a sample app, manually verify that identify, login, logout, and modify requests behave as expected

 - Closes https://go.mparticle.com/work/SQDSDKS-6419

### DIFF
--- a/src/identity-user-interfaces.ts
+++ b/src/identity-user-interfaces.ts
@@ -1,12 +1,36 @@
 import {
     AllUserAttributes,
     IdentityCallback,
+    IdentityResult,
+    IdentityResultBody,
+    MPID,
     Product,
     User,
 } from '@mparticle/web-sdk';
 import { SDKIdentityTypeEnum } from './identity.interfaces';
 import { MessageType } from './types.interfaces';
 import { BaseEvent } from './sdkRuntimeModels';
+
+// Cart is Deprecated and private to mParticle user in @mparticle/web-sdk
+// but we need to expose it here for type safety in some of our tests
+interface Cart {
+    /**
+     * @deprecated Cart persistence in mParticle has been deprecated. Please use mParticle.eCommerce.logProductAction(mParticle.ProductActionType.AddToCart, [products])
+     */
+    add: (product: Product, logEventBoolean?: boolean) => void;
+    /**
+     * @deprecated Cart persistence in mParticle has been deprecated. Please use mParticle.eCommerce.logProductAction(mParticle.ProductActionType.RemoveFromCart, [products])
+     */
+    remove: (product: Product, logEventBoolean?: boolean) => void;
+    /**
+     * @deprecated Cart persistence in mParticle has been deprecated.
+     */
+    clear: () => void;
+    /**
+     * @deprecated Cart Products have been deprecated
+     */
+    getCartProducts(): Product[];
+}
 
 // https://go.mparticle.com/work/SQDSDKS-5033
 // https://go.mparticle.com/work/SQDSDKS-6354
@@ -15,6 +39,10 @@ export interface IMParticleUser extends User {
     setUserTag(tagName: string, value?: any): void;
     setUserAttribute(key: string, value: any): void;
     getUserAudiences?(callback?: IdentityCallback): void;
+    /*
+     * @deprecated
+     */
+    getCart(): Cart;
 }
 
 export interface ISDKUserIdentity {
@@ -51,6 +79,19 @@ export interface ISDKUserAttributeChangeData {
 export interface IUserAttributeChangeEvent extends BaseEvent {
     messageType: MessageType.UserAttributeChange;
     userAttributeChanges: ISDKUserAttributeChangeData;
+}
+
+// https://go.mparticle.com/work/SQDSDKS-6460
+export interface IIdentityCallback extends IdentityCallback {
+    (result: IdentityResult): void;
+}
+
+export interface IIdentityResult extends Omit<IdentityResult, 'body'> {
+    body: IIdentityResultBody;
+}
+
+export interface IIdentityResultBody extends IdentityResultBody {
+    mpid: MPID;
 }
 
 export interface mParticleUserCart {
--- a/src/identity-user-interfaces.ts
+++ b/src/identity-user-interfaces.ts
@@ -1,12 +1,4 @@
-import {
-    AllUserAttributes,
-    IdentityCallback,
-    IdentityResult,
-    IdentityResultBody,
-    MPID,
-    Product,
-    User,
-} from '@mparticle/web-sdk';
+import { AllUserAttributes, MPID, Product, User } from '@mparticle/web-sdk';
 import { SDKIdentityTypeEnum } from './identity.interfaces';
 import { MessageType } from './types.interfaces';
 import { BaseEvent } from './sdkRuntimeModels';
@@ -82,19 +74,26 @@ export interface IUserAttributeChangeEvent extends BaseEvent {
 }
 
 // https://go.mparticle.com/work/SQDSDKS-6460
-export interface IIdentityCallback extends IdentityCallback {
-    (result: IIdentityResult): void;
+export interface IdentityCallback {
+    (result: IdentityResult): void;
 }
 
-export interface IIdentityResult extends Omit<IdentityResult, 'body'> {
-    body: IIdentityResultBody | IIdentityModifyResultBody;
+export interface IdentityResult {
+    httpCode: any;
+    getPreviousUser(): User;
+    getUser(): User;
+    body: IdentityResultBody | IdentityModifyResultBody;
 }
 
-export interface IIdentityResultBody extends IdentityResultBody {
+export interface IdentityResultBody {
+    context: string | null;
+    is_ephemeral: boolean;
+    is_logged_in: boolean;
+    matched_identities: Record<string, unknown>;
     mpid?: MPID;
 }
 
-export interface IIdentityModifyResultBody {
+export interface IdentityModifyResultBody {
     change_results?: {
         identity_type: SDKIdentityTypeEnum;
         modified_mpid: MPID;
--- a/src/identity-user-interfaces.ts
+++ b/src/identity-user-interfaces.ts
@@ -87,11 +87,18 @@ export interface IIdentityCallback extends IdentityCallback {
 }
 
 export interface IIdentityResult extends Omit<IdentityResult, 'body'> {
-    body: IIdentityResultBody;
+    body: IIdentityResultBody | IIdentityModifyResultBody;
 }
 
 export interface IIdentityResultBody extends IdentityResultBody {
-    mpid: MPID;
+    mpid?: MPID;
+}
+
+export interface IIdentityModifyResultBody {
+    change_results?: {
+        identity_type: SDKIdentityTypeEnum;
+        modified_mpid: MPID;
+    };
 }
 
 export interface mParticleUserCart {
--- a/src/identity-user-interfaces.ts
+++ b/src/identity-user-interfaces.ts
@@ -2,6 +2,8 @@ import { AllUserAttributes, MPID, Product, User } from '@mparticle/web-sdk';
 import { SDKIdentityTypeEnum } from './identity.interfaces';
 import { MessageType } from './types.interfaces';
 import { BaseEvent } from './sdkRuntimeModels';
+import Constants from './constants';
+const { HTTPCodes } = Constants;
 
 // Cart is Deprecated and private to mParticle user in @mparticle/web-sdk
 // but we need to expose it here for type safety in some of our tests
@@ -79,7 +81,7 @@ export interface IdentityCallback {
 }
 
 export interface IdentityResult {
-    httpCode: any;
+    httpCode: typeof HTTPCodes;
     getPreviousUser(): User;
     getUser(): User;
     body: IdentityResultBody | IdentityModifyResultBody;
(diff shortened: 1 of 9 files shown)


test/src/tests-identity.ts
```diff
@@ -1466,6 +1482,16 @@ describe('identity', function() {
...
 
         expect(combinedUIsByType).to.have.all.keys([1, 3, 4, 7]);
+        expect(combinedUIsByType[1]).to.equal('customerid1');
+        expect(combinedUIsByType[3]).to.equal('twitter5');
+        expect(combinedUIsByType[4]).to.equal('google4');
+        expect(combinedUIsByType[7]).to.equal('email2@test.com');
+
+        // if an invalid identity type is added to the identity call,
+        // it will be removed from the combinedUIsByType
+        // FIXME: What is the actual test case this is trying to test?
```
alexs-mparticle, line 1492: @rmi22186 We probably should talk about this

src/type-utils.ts
```diff
@@ -0,0 +1,5 @@
+import { IdentityType } from './types.interfaces';
+
+export interface IIdentityType {
+    getIdentityType(identityType: string): IdentityType | null;
```
alexs-mparticle, line 4: Note to self: Is this a duplicate of `types.interfaces.ts`?

test/src/tests-identity.ts
```diff
@@ -1442,25 +1465,19 @@ describe('identity', function() {
                 newUIByName
             );
 
-        combinedUIsByType.should.have.properties([1, 3, 4, 7]);
```
alexs-mparticle, line 1432: Note to self. Make sure this was not accidentally removed

test/src/tests-identity.ts
```diff
@@ -1687,83 +1704,6 @@ describe('identity', function() {
         done();
     });
 
-    it('should properly validate identityApiRequest values', function(done) {
```
alexs-mparticle, line 1677: @rmi22186 These tests are redundant since they're in the Validator tests.

test/src/tests-identity.ts
```diff
@@ -3400,16 +3354,16 @@ describe('identity', function() {
 
             mParticle.init(apiKey, window.mParticle.config);
 
-            let idCache = JSON.parse(
-                localStorage.getItem('mprtcl-v4_abcdef-id-cache')
-            );
+            let idCache: BaseVault<
```
alexs-mparticle, line 3375: Note to self. This should be a `const`

test/src/tests-identity.ts
```diff
@@ -3400,16 +3354,16 @@ describe('identity', function() {
...
 
             // a single identify cache key will be on the idCache
-            Should(Object.keys(idCache).length).equal(1);
+            expect(Object.keys(idCache).length).to.equal(1);
             for (let key in idCache) {
                 // we previously ticked forward 1 second, so the expire timestamp should be 1 second more than the X_MP_MAX_AGE
-                Should(idCache[key].expireTimestamp).equal(
-                    X_MP_MAX_AGE * 1000 + 1
+                expect(idCache[key].expireTimestamp).to.equal(
+                    ((X_MP_MAX_AGE as unknown) as number) * 1000 + 1
```
alexs-mparticle, line 3384: Note to self, add an explanation why this needs to be a string.

test/src/tests-identity.ts
rmi22186: this is a bizarre test.  i agree. 

there is a `return false` at the bottom of [getIdentityType](https://github.com/mParticle/mparticle-web-sdk/blob/master/src/types.js#L162-L206), which would only ever happen is something that wasn't a real  `identityName` got passed to this.  the idea here is that `newUIByName` has an `invalidKey`, which _would_ get the `return false` except that the logic in the [combineUserIdentities](https://github.com/mParticle/mparticle-web-sdk/blob/master/src/identity.js#L164-L169) already removes this possibility.

does that make sense?

rmi22186 (changes requested):
looking good.  a few nits

src/identity-user-interfaces.ts
```diff
@@ -53,6 +81,19 @@ export interface IUserAttributeChangeEvent extends BaseEvent {
     userAttributeChanges: ISDKUserAttributeChangeData;
 }
 
+// https://go.mparticle.com/work/SQDSDKS-6460
+export interface IIdentityCallback extends IdentityCallback {
+    (result: IdentityResult): void;
+}
+
```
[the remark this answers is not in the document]
alexs-mparticle: `IIdentityCallback` is meant to override `IdentityCallback` so that we can eventually patch `IdentityResultBody` which is missing `mpid`. I think I forgot to swap out  `IdentityResult` with `IIdentityResult`, so that should be good now.

src/identity.interfaces.ts
```diff
@@ -78,6 +99,10 @@ export interface IIdentityAPIIdentityChangeData {
 }
 
 export interface IIdentityRequest {
+    combineUserIdentities(
+        currentUserIdentities: UserIdentities,
+        newUserIdentities: UserIdentities
```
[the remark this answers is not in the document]
alexs-mparticle: I can't remember where we discussed this, but I remember us having a conversation where we thought `currentUserIdentities` and `newUserIdentities` was more appropriate and we should move towards that name.

I can go either way on this.

src/sdkRuntimeModels.ts
```diff
@@ -137,6 +144,9 @@ interface IEvents {
 
 export interface MParticleWebSDK {
     addForwarder(mockForwarder: MPForwarder): void;
+    // TODO: Figure out how we can make IdentityType Work
```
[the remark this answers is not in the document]
alexs-mparticle: Sorry. That's something for myself and we can punt on it for now.

test/src/tests-identity.ts
alexs-mparticle: Makes sense in a squirrely sort of way.

I would suggest we can get rid of the `expect(combinedUIsByType).to.not.have.property(false)` assertion.

Thoughts?

src/identity.interfaces.ts
```diff
@@ -24,6 +19,32 @@ export type IdentityPreProcessResult = {
 
 export type IdentityAPIMethod = valueof<typeof Constants.IdentityMethods>;
 
+// TODO: Consolidate this with @mparticle/web-sdk
+export type UserIdentities = {
+    customerid?: string | undefined;
+    email?: string | undefined;
+    other?: string | undefined;
+    other1?: string | undefined; // QUESTION: Why is this missing?
```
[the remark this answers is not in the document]
alexs-mparticle: From what I can tell looking at the backend code, `other1` doesn't seem to exist. However, in some of our old tests, we do make several references to it: [here](https://github.com/mParticle/mparticle-web-sdk/blob/efc1f7ba05a54725f09b338d3552c28aaf4acd05/test/src/tests-identity.js#L658) and [here](https://github.com/mParticle/mparticle-web-sdk/blob/efc1f7ba05a54725f09b338d3552c28aaf4acd05/test/src/tests-identities-attributes.js#L945)

I wanted to bubble up the question since part of this exercise is to highlight places where our assumptions or tests were incorrect.

test/src/tests-identity.ts
```diff
@@ -34,17 +39,47 @@ declare global {
...
+    userIdentities: 'badUserIdentitiesString',
+} as unknown) as IdentityApiData;
+
+const BadUserIdentitiesAsArray = ({
+    userIdentities: ['bad', 'user', 'identities', 'array'],
+} as unknown) as IdentityApiData;
+
+const BadUserIdentitiesAsNull = ({
+    userIdentities: null,
+} as unknown) as IdentityApiData;
```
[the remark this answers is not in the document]
alexs-mparticle: Not sure what you're referring to here. Maybe the link moved?

test/src/tests-identity.ts
rmi22186: Agreed that it's squirrely.  I think the point of it is to confir that there is a check for the falsey conditional though. Does the test currently fail?  If so i'd say there's a regression somewhere?  it's technically an integration test, and so should pass.  Any reason to not keep it and add commented code for context?

src/identity.interfaces.ts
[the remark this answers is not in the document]
rmi22186: The second link is actually an `other` key with value of `other1`, so that one doesn't apply.  First test is wrong.  It's possible that in the `createModifyIdentityRequest`, it doesn't do a validation of the identity types.  So either it should, or there is some sort of validation happening prior to that in the actual code.  My gut tells me to fix the test in the first link by removing the `other1` key, but double check the above, then make the call.

src/identity.interfaces.ts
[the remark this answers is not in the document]
rmi22186: I can't remember that discussion.  I'd say let's stick to the former for now since that's how the code currently is.  If there's a reason to change it in the future (maybe when refactoring identity.js --> identity.ts), then it'll come up again.

test/src/tests-identity.ts
```diff
@@ -34,17 +39,47 @@ declare global {
     }
 }
 
-let mockServer;
-const mParticle = window.mParticle;
+const mParticle = window.mParticle as MParticleWebSDK;
+
+const BadUserIdentitiesAsString = ({
```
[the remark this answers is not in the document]
rmi22186: I agree - all caps for constants for best practices.

I think i just meant `BadUserIdentitiesString` `BadUserIdentitiesArray` which is what it was before.  But I don't really care, it can be kept as is.

test/src/tests-identity.ts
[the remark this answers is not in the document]
rmi22186: Weird.  I just looked again for it.  This is the valid [constant](https://github.com/mParticle/mparticle-web-sdk/blob/master/test/src/tests-identity.js#L1708-L1712)
Along with the [valid test](https://github.com/mParticle/mparticle-web-sdk/blob/master/test/src/tests-identity.js#L1747)

src/identity-user-interfaces.ts
```diff
@@ -83,15 +83,22 @@ export interface IUserAttributeChangeEvent extends BaseEvent {
 
 // https://go.mparticle.com/work/SQDSDKS-6460
 export interface IIdentityCallback extends IdentityCallback {
-    (result: IdentityResult): void;
+    (result: IIdentityResult): void;
 }
 
 export interface IIdentityResult extends Omit<IdentityResult, 'body'> {
-    body: IIdentityResultBody;
+    body: IIdentityResultBody | IIdentityModifyResultBody;
```
rmi22186, line 90: If this is omiting the body but then adding a body with a new signature, can't you not omit it and just extend with a new body?

src/identity-utils.ts
```diff
@@ -7,6 +7,8 @@ import { IdentityAPIMethod } from './identity.interfaces';
 
 const { Identify, Modify, Login, Logout } = Constants.IdentityMethods;
 
+export type IdentityCache = BaseVault<Dictionary<ICachedIdentityCall>>;
```
rmi22186, line 10: nice change!

test/src/tests-identity.ts
```diff
@@ -12,9 +12,7 @@ import {
     workspaceCookieName,
 } from './config/constants';
 import { Callback, IdentityApiData, Product } from '@mparticle/web-sdk';
-import { BaseVault } from '../../src/vault';
-import { Dictionary } from '../../src/utils';
-import { ICachedIdentityCall } from '../../src/identity-utils';
+import { IdentityCache } from '../../src/identity-utils';
```
rmi22186, line 15: another great change 👍

rmi22186 (changes requested):
a few comments.  looking forward to discussing during our call

src/identity-user-interfaces.ts
alexs-mparticle: Good call. I'll try this out.

test/src/tests-identity.ts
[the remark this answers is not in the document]
rmi22186: We clarified during our conversation.  I'm wrong, you're right.

src/sdkRuntimeModels.ts
```diff
@@ -210,7 +210,10 @@ export type LogLevelType = 'none' | 'verbose' | 'warning' | 'error';
 // Currently, this extends MPConfiguration in @types/mparticle__web-sdk
 // and the two will be merged in once the Store module is refactored
 export interface SDKInitConfig
-    extends Omit<MPConfiguration, 'dataPlan' | 'logLevel'> {
+    extends Omit<
+        MPConfiguration,
+        'dataPlan' | 'logLevel' | 'identityCallback'
```
rmi22186, line 215: when i remove this change, it still passes.  any reason for adding this?

alexs-mparticle: Weird. I probably undid the change that needed this to be added in and didn't notice. Good catch.

src/identity-user-interfaces.ts
```diff
@@ -53,6 +73,33 @@ export interface IUserAttributeChangeEvent extends BaseEvent {
     userAttributeChanges: ISDKUserAttributeChangeData;
 }
 
+// https://go.mparticle.com/work/SQDSDKS-6460
+export interface IdentityCallback {
+    (result: IdentityResult): void;
+}
+
+export interface IdentityResult {
+    httpCode: any;
```
jared-dickman, line 82: **nit** doubt this is really _any_...

rmi22186 (approved):

PR #885, opened 2024-05-15
Refactor: Migrate Identity Attribute Tests to TypeScript

alexs-mparticle:
## Instructions

 ## Summary
 - Migrates Unit and Integration test for Identity Attributes from plain JavaScript to TypeScript.

 ## Testing Plan
- Run all build and automated tests to verify they pass correctly
- Using a sample app, manually verify that setting user attributes and other user identity functions behave as expected

 - Closes https://go.mparticle.com/work/SQDSDKS-6463

### DIFF
--- a/test/src/tests-identities-attributes.ts
+++ b/test/src/tests-identities-attributes.ts
@@ -26,7 +26,6 @@ let mockServer;
 declare global {
     interface Window {
         mParticle: MParticleWebSDK;
-        fetchMock: any;
     }
 }
 
--- a/test/src/tests-identities-attributes.ts
+++ b/test/src/tests-identities-attributes.ts
@@ -1,4 +1,3 @@
-import Utils from './config/utils';
 import sinon from 'sinon';
 import { expect } from 'chai';
 import fetchMock from 'fetch-mock/esm/client';
@@ -9,6 +8,7 @@ import {
     MPConfig,
     MILLISECONDS_IN_ONE_DAY_PLUS_ONE_SECOND,
 } from './config/constants';
+import Utils from './config/utils';
 import { MParticleWebSDK } from '../../src/sdkRuntimeModels';
 import { UserAttributesValue } from '@mparticle/web-sdk';
 import { UserAttributes } from '../../src/identity-user-interfaces';
@@ -84,11 +84,11 @@ describe('identities and attributes', function() {
 
         const event = findBatch(fetchMock.calls(), 'test user attributes');
 
-        event.should.have.property('user_attributes');
-        event.user_attributes.should.have.property('gender', 'male');
+        expect(event).to.have.property('user_attributes');
+        expect(event.user_attributes).to.have.property('gender', 'male');
 
         const cookies = getLocalStorage();
-        cookies[testMPID].ua.should.have.property('gender', 'male');
+        expect(cookies[testMPID].ua).to.have.property('gender', 'male');
 
         done();
     });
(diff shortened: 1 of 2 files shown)


test/src/tests-identities-attributes.ts
```diff
@@ -628,8 +721,11 @@ describe('identities and attributes', function() {
         fetchMock.resetHistory();
         mParticle.Identity.getCurrentUser().setUserAttribute('age', '25');
 
-        let body = JSON.parse(fetchMock.lastOptions().body)
-        body.user_attributes.should.have.property('age', '25')
+        let body: UserAttributes = JSON.parse(
+            `${fetchMock.lastOptions().body}`
+        ) as UserAttributes; // Convert object to string
```
rmi22186, line 748: Nit - I think this comment should go above this code block, rather than being in line with the TS `as` declaration since the JSON.parse extends 3 lines

rmi22186: also, is the reason you are using interpolation because TS doesn't recognize that this `fetchMock.lastOptions().body` is already a string?

test/src/tests-identities-attributes.ts
```diff
@@ -0,0 +1,1376 @@
...
+            userIdentities: {
+                customerid: 'customerid1',
+            },
+        };
+        mParticle.Identity.login(loginUser);
+
+        let body = JSON.parse(`${fetchMock.lastOptions().body}`);
+        // should be the new MPID
+
+        expect(body.mpid).to.equal('anotherMPID');
```
rmi22186, line 980: nit.  bring this comment closer to the test

rmi22186 (changes requested):

test/src/tests-identities-attributes.ts
alexs-mparticle: I'm actually going to rework this to be a bit more clear as to why we're parsing the body as a string and so we can maintain the correct type of the subsequent Event.

TS actually recognizes that `fetchMock.lastOptions().body` is an Object, but doesn't like that we're trying to parse it as if it were a string in this line because TS expects and object. I personally find using string interpolation to be more readable than using `as unknown as string`, but I can use that if you think it's more clear.

test/src/tests-identities-attributes.ts
```diff
@@ -0,0 +1,1389 @@
...
+    getLocalStorage,
+    MockForwarder,
+} = Utils;
+
+let mockServer;
+
+declare global {
+    interface Window {
+        mParticle: MParticleWebSDK;
+        fetchMock: any;
```
rmi22186: `fetchMock` is being imported. Is there a reason you are putting it on the window here?

test/src/tests-identities-attributes.ts
```diff
@@ -9,9 +9,10 @@ import {
     MILLISECONDS_IN_ONE_DAY_PLUS_ONE_SECOND,
 } from './config/constants';
 import Utils from './config/utils';
-import { MParticleWebSDK } from '../../src/sdkRuntimeModels';
-import { UserAttributesValue } from '@mparticle/web-sdk';
+import { MParticleWebSDK, SDKEvent } from '../../src/sdkRuntimeModels';
+import { AllUserAttributes, UserAttributesValue } from '@mparticle/web-sdk';
 import { UserAttributes } from '../../src/identity-user-interfaces';
+import { CustomEvent, UserAttributeChangeEvent } from '@mparticle/event-models';
```
rmi22186, line 15: `CustomEvent` is not being used anywhere

rmi22186 (approved):

PR #886, opened 2024-05-22
refactor: Identify refactor points in Identity related workflows

alexs-mparticle:
## Instructions

 ## Summary
 - Identify future refactor points for Identity related workflows

 ## Testing Plan
 - N/A - this only adds comments to certain parts of the code

 - Closes https://go.mparticle.com/work/SQDSDKS-6486

### DIFF
--- a/src/identity.js
+++ b/src/identity.js
@@ -1788,7 +1788,6 @@ export default function Identity(mpInstance) {
         mpid,
         prevUserIdentities
     ) {
-        // https://go.mparticle.com/work/SQDSDKS-6354
         var currentUserInMemory, userIdentityChangeEvent;
 
         if (!mpid) {
@@ -1797,6 +1796,7 @@ export default function Identity(mpInstance) {
             }
         }
 
+        // https://go.mparticle.com/work/SQDSDKS-6354
         currentUserInMemory = this.IdentityAPI.getUser(mpid);
 
         for (var identityType in newUserIdentities) {
--- a/src/identity.js
+++ b/src/identity.js
@@ -17,6 +17,7 @@ export default function Identity(mpInstance) {
     this.idCache = null;
     this.audienceManager = null;
 
+    // https://go.mparticle.com/work/SQDSDKS-6353
     this.IdentityRequest = {
         preProcessIdentityRequest: function(identityApiData, callback, method) {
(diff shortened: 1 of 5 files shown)


src/identity.js
```diff
@@ -796,6 +801,7 @@ export default function Identity(mpInstance) {
         },
     };
 
+    //go.mparticle.com/work/SQDSDKS-6354
```
rmi22186, line 804:
```suggestion
    // https://go.mparticle.com/work/SQDSDKS-6354
```

src/identity.js
```diff
@@ -1791,6 +1803,7 @@ export default function Identity(mpInstance) {
         mpid,
         prevUserIdentities
     ) {
+        // https://go.mparticle.com/work/SQDSDKS-6354
```
rmi22186, line 1806: Going to this work item doesn't match the code block here.

src/sdkRuntimeModels.ts
```diff
@@ -34,7 +34,10 @@ export interface SDKEvent {
     IsFirstRun: boolean;
     EventName: string;
     EventCategory: number;
+
+    // https://go.mparticle.com/work/SQDSDKS-5196
     UserAttributes?: { [key: string]: string | string[] | null };
+
```
rmi22186, line 40: is this whitespace intentional?

src/serverModel.ts
```diff
@@ -46,7 +46,10 @@ export interface IServerV2DTO {
     dp_v?: number;
     n?: string;
     et?: number;
+
+    // https://go.mparticle.com/work/SQDSDKS-5196
     ua?: Dictionary<string | string[]>;
+
```
rmi22186, line 52: is this whitespace intentional?

rmi22186 (changes requested):

src/sdkRuntimeModels.ts
alexs-mparticle: Yes. It's to show that the comment only applies to that specific type definition.

src/serverModel.ts
alexs-mparticle: Yes. It's to show that the comment only applies to that specific type definition.

rmi22186 (approved):

PR #888, opened 2024-05-22
refactor: Refactor ParseIdentityResponse Workflow

alexs-mparticle:
## Instructions

 ## Summary
 - Reintroduces #864 after a revert #879, with more test coverage.

 ## Testing Plan
 - Using a sample app, verify that all identity methods work as expected

 - Closes https://go.mparticle.com/work/SQDSDKS-6461

### DIFF
--- a/src/identity.interfaces.ts
+++ b/src/identity.interfaces.ts
@@ -188,9 +188,9 @@ export interface IIdentity {
     ): void;
     sendUserIdentityChangeEvent(
         newUserIdentities: UserIdentities,
-        method: IdentityAPIMethod,
+        prevUserIdentities: UserIdentities,
         mpid: MPID,
-        prevUserIdentities: UserIdentities
+        method: IdentityAPIMethod
     ): void;
 
     /**
--- a/src/identity.interfaces.ts
+++ b/src/identity.interfaces.ts
@@ -188,9 +188,9 @@ export interface IIdentity {
     ): void;
     sendUserIdentityChangeEvent(
         newUserIdentities: UserIdentities,
-        prevUserIdentities: UserIdentities,
+        method: IdentityAPIMethod,
         mpid: MPID,
-        method: IdentityAPIMethod
+        prevUserIdentities: UserIdentities
     ): void;
 
     /**
--- a/src/identity.js
+++ b/src/identity.js
@@ -1723,8 +1723,8 @@ export default function Identity(mpInstance) {
                 newUserIdentities[identityType]
             ) {
                 // If a new identity type was introduced when the identity changes
-                // we need to notify the server so that it will associate the new
-                // identity type with subsequent events
+                // we need to notify the server so that the user profile is updated in
+                // the mParticle UI.
                 const isNewUserIdentityType = !prevUserIdentities[identityType];
                 const userIdentityChangeEvent = self.createUserIdentityChange(
                     identityType,
--- a/src/identity.js
+++ b/src/identity.js
@@ -6,13 +6,22 @@ import {
     tryCacheIdentity,
 } from './identity-utils';
 import AudienceManager from './audienceManager';
-
-const { Messages, HTTPCodes, FeatureFlags } = Constants;
+const { Messages, HTTPCodes, FeatureFlags, IdentityMethods } = Constants;
 const { ErrorMessages } = Messages;
-
-const { Identify, Modify, Login, Logout } = Constants.IdentityMethods;
+const { CacheIdentity } = FeatureFlags;
+const { Identify, Modify, Login, Logout } = IdentityMethods;
+import {
+    generateDeprecationMessage,
+    isEmpty,
+    isFunction,
+    isObject,
+} from './utils';
+import { didUserChange } from './user-utils';
+import { getNewIdentitiesByName } from './type-utils';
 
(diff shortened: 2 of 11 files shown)


src/identity.js
```diff
@@ -1550,6 +1542,14 @@ export default function Identity(mpInstance) {
                     );
                 }
 
+                const incomingUser = self.IdentityAPI.getUser(
+                    identityApiResult.mpid
+                );
+
+                const incomingMpidUIByName = incomingUser
```
rmi22186, line 1549: i think `incomingMpidUIByName` can be shortened to `incomingUIByName`.  Elsewhere we have `previousUIByName`, so it's just being consistent with that.

src/identity.js
```diff
@@ -1784,31 +1700,28 @@ export default function Identity(mpInstance) {
...
-        prevUserIdentities
+        currentUserIdentities
     ) {
-        var currentUserInMemory, userIdentityChangeEvent;
-
-        if (!mpid) {
-            if (method !== Modify) {
-                return;
-            }
-        }
```
rmi22186, line 1797: I'm a little nervous to remove this code, even if it doesn't break any tests given the code was not commented at all before.  Maybe we can add this code back in, and add a go link for trying to figure out what is going on with this code block.

src/identity.js
```diff
@@ -1663,106 +1622,63 @@ export default function Identity(mpInstance) {
...
+                );
+            } else if (
+                identityApiResult &&
+                !isEmpty(identityApiResult.errors)
+            ) {
+                mpInstance.Logger.error(
+                    'Received HTTP response code of ' +
+                        xhr.status +
+                        ' - ' +
+                        identityApiResult.errors[0].message
```
rmi22186, line 1676: Can we add a go link for figuring out why we'd only want to return the 1st error here?  It could start with figuring out what potential errors would be returned.

rmi22186 (approved):

rmi22186 (changes requested):
accidentally approved, i meant to click request changes.

src/identity.js
alexs-mparticle: I think the reason we removed it is because the original usage was possibly incorrect. For example:

```javascript
self.sendUserIdentityChangeEvent(
    newIdentitiesByName,
    method,
    identityApiResult.mpid,
    method === Modify
        ? previousUIByNameCopy
        : incomingMpidUIByNameCopy
);
```

Given that, we'd also have to restore the original signature as well.

src/identity.interfaces.ts
```diff
@@ -188,9 +188,9 @@ export interface IIdentity {
     ): void;
     sendUserIdentityChangeEvent(
         newUserIdentities: UserIdentities,
-        method: IdentityAPIMethod,
+        prevUserIdentities: UserIdentities,
         mpid: MPID,
-        prevUserIdentities: UserIdentities
+        method: IdentityAPIMethod
```
rmi22186, line 193: let's flip these 2 back to reduce the number of changes

rmi22186 (changes requested):

src/user-utils.ts
```diff
@@ -0,0 +1,12 @@
+import { IMParticleUser } from './identity-user-interfaces';
+
+export function didUserChange(
```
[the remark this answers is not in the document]
alexs-mparticle: I think the context here is how the user changes. For `didUserChange` we're comparing previous and new user mpid and `isLoggedIn` values, but in `hasUserChanged` we're technically comparing whether the `mpid` of the user and Identity API Data is different.

This may be unnecessarily verbose, but perhaps we can rename them to `hasMPID` and `hasMPIDAndUserLoginChanged`?

test/src/tests-store.ts
```diff
@@ -253,6 +253,60 @@ describe('Store', () => {
...
+
+        it('should move an mpid to the end of the currentSessionMPIDs list if it is also the previous mpid', () => {
+            const store: IStore = new Store(
+                sampleConfig,
+                window.mParticle.getInstance()
+            );
+
+            store.sessionId = 'my-session-id';
+            store.currentSessionMPIDs = ['oldmpid', 'newermpid', 'recentmpid'];
+            store.addMpidToSessionHistory('oldmpid', 'oldmpid');
```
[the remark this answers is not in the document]
alexs-mparticle: Ignoring this after our recent conversation

test/jest/user-utils.spec.ts
```diff
@@ -1,50 +1,139 @@
...
 
-            expect(didUserChange(previousUser, newUser)).toBeTruthy();
+            expect(
+                hasMPIDAndUserLoginChanged(previousUser, newUser)
+            ).toBeTruthy();
+        });
+    });
+
+    describe('#hasMPIDChanged', () => {
+        const identityResultBody: IdentityResultBody = {
```
rmi22186, line 66: there are a few tests here where the incoming user is empty, or mpid is null/undefined.  this makes sense from a code perspective, but maybe there needs to be a jira ticket added about determining which is not actually possible from a use case perspective.

alexs-mparticle: Created https://go.mparticle.com/work/SQDSDKS-6504 to address this.

rmi22186 (approved):

PR #890, opened 2024-05-29
refactor: Migrate User Attribute Persistence Methods (#868)

alexs-mparticle:
## Instructions

 ## Summary
 - Migrates Persistence methods for User Attributes to Store

 ## Testing Plan
 - Smoke test a new, anonymous user session in a sample app, then attempt to add new user attributes
 - Subsequent events should have updated user attribute properties
 - Refreshing the app should retain the new attributes

 - Closes https://go.mparticle.com/work/SQDSDKS-6513


src/identity-user-interfaces.ts
```diff
@@ -62,6 +62,10 @@ export interface IUserIdentityChangeEvent extends BaseEvent {
     userIdentityChanges: ISDKUserIdentityChanges;
 }
 
+export interface ISDKUserAttribute {
```
rmi22186, line 65: nit: I think this should be plural. saying UserAttribute makes it sound like single attribute, which could be either the key or the value.  We want to make it more clear that it is the object of all k:v pairs.

```suggestion
export interface ISDKUserAttributes {
```

test/src/tests-store.ts
```diff
@@ -722,6 +722,109 @@ describe('Store', () => {
...
+                sampleConfig,
+                window.mParticle.getInstance()
+            );
+
+            expect(store.getUserAttributes(testMPID)).to.deep.equal({});
+        });
+    });
+
+    describe('#setUserAttributes', () => {
+        it('should set user attributes in the store', () => {
```
rmi22186, line 761:
```suggestion
        it('should set user attributes in the store, overwriting any previous user attributes that exist', () => {
```

rmi22186 (changes requested):
minor nits

rmi22186 (approved):

PR #898, opened 2024-07-30
feat: Refactor usage of XHR for Config API

alexs-mparticle:
## Instructions

 ## Summary
 - Replaces XHR Uploader with AsyncUploader, which allows config api requests to support both XHR and Fetch.

 ## Testing Plan
 - Using both self-hosted and CDN sample apps, attempt to load the Web SDK and verify expected behavior.

 - Closes https://go.mparticle.com/work/SQDSDKS-6566

### DIFF
--- a/src/configAPIClient.ts
+++ b/src/configAPIClient.ts
@@ -13,13 +13,6 @@ import {
     FetchUploader,
     XHRUploader,
 } from './uploaders';
-import Constants from './constants';
-
-export type SDKCompleteInitCallback = (
-    apiKey: string,
-    config: SDKInitConfig,
-    mpInstance: MParticleWebSDK
-) => void;
 
 export interface IKitConfigs extends IKitFilterSettings {
     name: string;
@@ -103,15 +96,13 @@ export interface IConfigAPIClient {
     getSDKConfiguration: () => Promise<IConfigResponse>;
 }
 
-// QUESTION: Should the uploader care about building urls?
 const buildUrl = (
     configUrl: string,
     apiKey: string,
     dataPlanConfig?: DataPlanConfig | null,
     isDevelopmentMode?: boolean | null
 ): string => {
-    let url = configUrl + apiKey + '/config';
-
+    const url = configUrl + apiKey + '/config';
     const env = isDevelopmentMode ? '1' : '0';
     const queryParams = [`env=${env}`];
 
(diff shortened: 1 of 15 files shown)


src/identity.js
```diff
@@ -1665,7 +1665,7 @@ export default function Identity(mpInstance) {
 
             mpInstance.Logger.verbose('Successfully parsed Identity Response');
 
-            mpInstance._APIClient.processQueuedEvents();
+            mpInstance._APIClient?.processQueuedEvents();
```
rmi22186, line 1668:
```suggestion
            // https://go.mparticle.com/work/SQDSDKS-6568
            mpInstance._APIClient?.processQueuedEvents();
```

src/configAPIClient.ts
```diff
@@ -91,78 +90,95 @@ export interface IConfigResponse {
...
-                apiKey +
-                '/config?env=';
-            if (config.isDevelopmentMode) {
-                url = url + '1';
-            } else {
-                url = url + '0';
-            }
+            const response = await uploader.upload(fetchPayload);
+            if (response.status === 200) {
+                console.log(
```
rmi22186, line 150: use mpinstance.logger instead of console.log

src/configAPIClient.ts
```diff
@@ -91,78 +90,95 @@ export interface IConfigResponse {
...
-            const dataPlan = config.dataPlan as DataPlanConfig;
-            if (dataPlan) {
-                if (dataPlan.planId) {
-                    url = url + '&plan_id=' + dataPlan.planId || '';
-                }
-                if (dataPlan.planVersion) {
-                    url = url + '&plan_version=' + dataPlan.planVersion || '';
+                // https://go.mparticle.com/work/SQDSDKS-6568
+                // FetchUploader returns the response as a JSON object that we have to await
+                if (response.json) {
```
rmi22186, line 159: is this the only thing that will determine if it is a fetch uploader?  ether way, for readability, can we make the xhrresponse below part of an `else`?  it seems more clear that it's one or the other

rmi22186: then return the config response once to DRY it up.  i'm not committed to this idea, just a thought.

src/configAPIClient.ts
```diff
@@ -91,78 +90,95 @@ export interface IConfigResponse {
...
+                // XHRUploader returns the response as a string that we need to parse
+                const xhrResponse = response as unknown as XMLHttpRequest;
+                configResponse = JSON.parse(xhrResponse.responseText);
+                return configResponse;
             }
+
+            mpInstance?.Logger?.verbose(
+                'Issue with receiving configuration from server, received HTTP Code of ' +
+                    response.statusText
+            );
```
rmi22186, line 174: i don't think this is ever reached?  I think this can be removed.

src/uploaders.ts
```diff
@@ -33,26 +33,39 @@ export class XHRUploader extends AsyncUploader {
...
-            fetchPayload.body
+            fetchPayload.body,
+            fetchPayload.method as 'post' | 'get'
         );
         return response;
     }
 
-    private async makeRequest(url: string, data: string): Promise<Response> {
+    // XHR Ready States
+    // https://stackoverflow.com/questions/30522565/what-is-meaning-of-xhr-readystate-4#:~:text=State%204%20means%20that%20the,finished%20downloading%20the%20response%20content.
```
rmi22186, line 43: I'd rather use this link...just seems more reputable
```suggestion
    // https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/readyState
```

rmi22186 (changes requested):
very minor nits

src/configAPIClient.ts
alexs-mparticle: No need. There's already an mp logger below. This console.log was just for testing.

src/configAPIClient.ts
alexs-mparticle: I'm good with that.

rmi22186 (approved):

PR #900, opened 2024-08-09
fix: Respect directUrlRouting flag in self hosted environments

rmi22186:
## Instructions

 ## Summary
The Direct Url Routing flag currently is respected when customers load us via CDN, but I discovered a bug where it wasn't on self hosted environments.  This is because of how the config is treated and processed in CDN vs self hosted.

* CDN - All of the config and flags are in the same mparticle.js file that the server returns via CDN.  In this set up, the [flags are processed](https://github.com/mParticle/mparticle-web-sdk/blob/master/src/store.ts#L294), then the [baseUrls are configured](https://github.com/mParticle/mparticle-web-sdk/blob/master/src/store.ts#L307-L316) shortly after when creating the `Store`.  This is all done in a method when we initialize mParticle called `runPreConfigFetchInitialization`.
* Self hosted - the mparticle code is bundled into an app, and we have to asyncronously fetch the config from the server.  While we still process a user provided config in `runPreConfigFetchInitialization`, we also need to [reprocess the newly returned config](https://github.com/mParticle/mparticle-web-sdk/blob/master/src/mp-instance.js#L1292-L1297) as it has settings like `flags` on it.  

In this `processConfig` function, we were not reprocessing the baseUrls, which I am now doing.

 ## Testing Plan

Tested locally using a test app, and added integration tests as well

 - Closes https://go.mparticle.com/work/SQDSDKS

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -83,6 +83,7 @@ export interface SDKConfig {
     useNativeSdk?: boolean;
     useCookieStorage?: boolean;
     v1SecureServiceUrl?: string;
+    // https://go.mparticle.com/work/SQDSDKS-6618
     v2SecureServiceUrl?: string;
     v3SecureServiceUrl?: string;
     webviewBridgeName?: string;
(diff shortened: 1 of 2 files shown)


einsteinx2 (approved):
LGTM, reviewed live on Zoom

PR #901, opened 2024-08-09
feat: Enable passing sourceMessageId to logProductAction

rmi22186:
## Instructions

 ## Summary
`logBaseEvent` allowed for `sourceMessageId` to be passed to it - [docs](https://docs.mparticle.com/integrations/facebook/event/#other-scenarios)

This is a feature so that facebook can de-dupe events based on an id.

However, this only works for custom events.  In the past, we added an options parameter to `logProductAction` (among other APIs), to allow for optional flags to be passed through.  We will leverage this open API to now let a key of `sourceMessageId` to be passed to it, so that commerce events also can have a user defined sourceMessageId.

 ## Testing Plan

 - Closes https://go.mparticle.com/work/SQDSDKS-6606

### DIFF
--- a/test/src/tests-eCommerce.js
+++ b/test/src/tests-eCommerce.js
@@ -1611,5 +1611,45 @@ describe('eCommerce', function() {
             productAction.ShippingAmount.should.equal(0)
             productAction.TaxAmount.should.equal(0)
         });
+
+        it('should allow a user to pass in a source_message_id to a commerce event', function() {
+             const product = mParticle.eCommerce.createProduct(
+                'iPhone',
+                '12345',
+                '400',
+                2,
+                'Plus',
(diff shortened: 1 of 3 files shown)


test/src/tests-eCommerce.js
```diff
@@ -1611,5 +1611,44 @@ describe('eCommerce', function() {
...
+            transactionAttributes = mParticle.eCommerce.createTransactionAttributes(
+                '12345',
+                'test-affiliation',
+                'coupon-code',
+                44334,
+                600,
+                200
+            );
+            
+
```
rmi22186, line 1638:
```suggestion
```

BrandonStalnaker (approved):
LGTM

PR #904, opened 2024-08-22
refactor: Use fetch instead of XHR for forwarding stats uploads

rmi22186:
## Instructions

 ## Summary
This PR prioritizes using `fetch` in when sending forwarding stats to mParticle servers.  XHR is used as a fallback.

 ## Testing Plan
 - [X] Was this tested locally? If not, explain why.  Updated unit test.  Tested on local app using fetch, and then deleting `window.fetch` in order to test XHR fallback.

 - Closes https://go.mparticle.com/work/SQDSDKS-6567


src/apiClient.ts
```diff
@@ -18,18 +23,32 @@ export interface IAPIClient {
...
-        forwardingStatsData: ForwardingStatsData
+        sendSingleForwardingStatsToServer: IForwardingStatsData
     ) => void;
     prepareForwardingStats: (
         forwarder: MPForwarder,
         event: IUploadObject
     ) => void;
 }
 
+export interface IForwardingStatsData {
```
alexs-mparticle, line 38: Please add comments for what these abbreviated attributes represent.

src/apiClient.ts
```diff
@@ -18,18 +23,32 @@ export interface IAPIClient {
...
+    mid: number;
+    esid: number;
+    n: string;
+    attrs: Dictionary<any>;
```
alexs-mparticle, line 42: Let's avoid using `any`. Either we should make a type for this attribute, or use `Record<string, string>` if it's a simple key/value pair. I would prefer the former if there are a specific attributes that must be passed.

src/apiClient.ts
```diff
@@ -161,37 +180,44 @@ export default function APIClient(
...
-            };
-            const xhr = mpInstance._Helpers.createXHR(xhrCallback);
-            url = mpInstance._Helpers.createServiceUrl(
-                mpInstance._Store.SDKConfig.v1SecureServiceUrl,
-                mpInstance._Store.devToken
-            );
-            data = forwardingStatsData;
+    this.sendSingleForwardingStatsToServer = async (
+        forwardingStatsData: IForwardingStatsData
+    ) => {
```
alexs-mparticle, line 185:
```suggestion
    this.sendSingleForwardingStatsToServer = async (
        forwardingStatsData: IForwardingStatsData
    ): void => {
```

src/apiClient.ts
```diff
@@ -161,37 +180,44 @@ export default function APIClient(
...
+        const { devToken } = mpInstance._Store;
+
+        const uploadUrl = `https://${forwardingDomain}${devToken}/Forwarding`;
+
+        let url: string;
+        const uploader: AsyncUploader = window.fetch
+            ? new FetchUploader(uploadUrl)
+            : new XHRUploader(uploadUrl);
+
+        const fetchPayload: fetchPayload = {
```
alexs-mparticle, line 196: I think we should tag this for refactor since I think a lot of the headers we're passing in are the same and can be set in the `AsyncUploader`

```suggestion
        // https://go.mparticle.com/work/SQDSDKS-6568
        const fetchPayload: fetchPayload = {
```

src/apiClient.ts
```diff
@@ -161,37 +180,44 @@ export default function APIClient(
...
+            mpInstance?.Logger?.verbose(
+                'Successfully sent forwarding stats to mParticle Servers configuration from server'
+            );
+        } else {
+            mpInstance?.Logger?.verbose(
+                'Issue with receiving configuration from server, received HTTP Code of ' +
+                    response.statusText
             );
         }
+
```
alexs-mparticle, line 218: I think this makes things more readable and possibly can remove the need for a comment.

```suggestion
        // This is a fire and forget, so we only need to log the response based on the code, and not return any response body
        let message: string;
        if (response.status === 202) {
            message = 'Successfully sent forwarding stats to mParticle Servers configuration from server';
        } else {
            message = 'Issue with receiving configuration from server, received HTTP Code of ' + response.statusText;
        }
        mpInstance?.Logger?.verbose(message);
```

test/src/tests-forwarders.js
```diff
@@ -1011,7 +1016,8 @@ describe('forwarders', function() {
             true
         );
 
-        mockServer.requests[mockServer.requests.length - 1].url.includes(
+        const calls = fetchMock.calls();
```
alexs-mparticle, line 1019: I think `fetchMock.lastCall()` could reduce this into a single line.

alexs-mparticle (changes requested):

src/apiClient.ts
rmi22186: i ended up moving this to the forwarder.js file, so we lost the typing here.  it seemed to fit better there logically, but i can revert if you think keeping the types is more important than organization

alexs-mparticle (changes requested):

src/forwarders.js
```diff
@@ -771,4 +776,30 @@ export default function Forwarders(mpInstance, kitBlocker) {
...
+                'Content-Type': 'text/plain;charset=UTF-8',
+            },
+        };
+
+        const response = await this.forwarderStatsUploader.upload(fetchPayload);
+
+        let message;
+        // This is a fire and forget, so we only need to log the response based on the code, and not return any response body
+        if (response.status === 202) {
+            message = 'Successfully sent forwarding stats to mParticle Servers';
```
rmi22186, line 796:
```suggestion
            // https://go.mparticle.com/work/SQDSDKS-6670
            message = 'Successfully sent forwarding stats to mParticle Servers';
```

alexs-mparticle (approved):

PR #905, opened 2024-08-23
fix: Remove XHR References from Identity Utils

alexs-mparticle:
## Instructions

 ## Summary
 - Replaces usage of `XHttpRequest` with a `IdentityResponse` Object so we can cache the actual data of the response rather than having our caching logic tightly coupled to XHR.
 - Uses a temporary `xhrIdentityResponseAdapter` to facility the conversion of an XHR response to an `IdentityResponse` object. This will conversion logic eventually be moved into `IdentityApiClient` in a future revision but is necessary to reduce the scope of changes in this PR.

 ## Testing Plan
 - All automated tests should pass and all identity request caching logic should behave as expected 

 - Closes https://go.mparticle.com/work/SQDSDKS-6657

### DIFF
--- a/src/identity-utils.ts
+++ b/src/identity-utils.ts
@@ -2,10 +2,24 @@ import Constants, { ONE_DAY_IN_SECONDS, MILLIS_IN_ONE_SEC } from './constants';
 import { Dictionary, parseNumber, isObject, generateHash } from './utils';
 import { BaseVault } from './vault';
 import Types from './types';
-import { IdentityApiData, UserIdentities, IdentityCallback } from '@mparticle/web-sdk';
+import {
+    IdentityApiData,
+    UserIdentities,
+    IdentityCallback,
+} from '@mparticle/web-sdk';
 import { IdentityAPIMethod } from './identity.interfaces';
+import { IdentityResultBody } from './identity-user-interfaces';
 
 const { Identify, Modify, Login, Logout } = Constants.IdentityMethods;
+const CACHE_HEADER = 'x-mp-max-age' as const;
+
+export interface IIdentityResponse {
+    // https://go.mparticle.com/work/SQDSDKS-6672
+    responseText: IdentityResultBody;
+    status: number;
+    cacheMaxAge: number; // Default: 86400
+    expireTimestamp?: number;
+}
 
 export type IdentityCache = BaseVault<Dictionary<ICachedIdentityCall>>;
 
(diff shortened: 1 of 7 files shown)


test/src/tests-identity-utils.ts
```diff
@@ -23,9 +24,10 @@ import { IdentityApiData } from '@mparticle/web-sdk';
 import Identity from "../../src/identity";
 
 import Constants from '../../src/constants';
-const { Identify, Modify, Login, Logout } = Constants.IdentityMethods;
+const { Identify, Login, Logout } = Constants.IdentityMethods;
 
 import sinon from 'sinon';
+import { IdentityResultBody } from "../../src/identity-user-interfaces";
```
alexs-mparticle, line 30:
```suggestion
// https://go.mparticle.com/work/SQDSDKS-6671
import { IdentityResultBody } from "../../src/identity-user-interfaces";
```

src/identity-utils.ts
```diff
@@ -2,10 +2,23 @@ import Constants, { ONE_DAY_IN_SECONDS, MILLIS_IN_ONE_SEC } from './constants';
...
+import {
+    IdentityApiData,
+    UserIdentities,
+    IdentityCallback,
+} from '@mparticle/web-sdk';
 import { IdentityAPIMethod } from './identity.interfaces';
+import { IdentityResultBody } from './identity-user-interfaces';
 
 const { Identify, Modify, Login, Logout } = Constants.IdentityMethods;
+const CACHE_HEADER = 'x-mp-max-age';
```
jared-dickman, line 14:
```suggestion
const CACHE_HEADER = 'x-mp-max-age' as const;
```

src/identity-utils.ts
```diff
@@ -17,74 +30,101 @@ export type IParseCachedIdentityResponse = (
...
     expireTimestamp: number;
 }
 
+// https://go.mparticle.com/work/SQDSDKS-6568
+// Temporary adapter to convert the XMLHttpRequest response to the IIdentityResponse interface
+export const xhrIdentityResponseAdapter = (
+    possiblyXhr: XMLHttpRequest | IIdentityResponse
+): IIdentityResponse => {
+    // This might be an actual xhr object, so we need to parse it
+    if (!possiblyXhr.hasOwnProperty('expireTimestamp')) {
```
jared-dickman, line 52: suggestion: when i see a `if !() return`, i consider flipping it. this looks like a good candidate, since the else is a 1 line return

something like ```
if (hasProp expireTimestamp) return possiblyXhr 
return {
   ...
}
```

File: src/identity-utils.ts [hunk 4]
```diff
@@ -17,74 +30,101 @@ export type IParseCachedIdentityResponse = (
...
+// https://go.mparticle.com/work/SQDSDKS-6568
+// Temporary adapter to convert the XMLHttpRequest response to the IIdentityResponse interface
+export const xhrIdentityResponseAdapter = (
+    possiblyXhr: XMLHttpRequest | IIdentityResponse
+): IIdentityResponse => {
+    if (possiblyXhr.hasOwnProperty('expireTimestamp')) {
+        // This is likely an IIdentityResponse object, just return it
+        return possiblyXhr as IIdentityResponse;
+    } else {
+        // This might be an actual xhr object, so we need to parse it
```
rmi22186, line 55: instead of saying it *might* be an actual xhr object in this path, I think the comment should be more clear that it IS an xhr object going down this code path.  So maybe we should just say that.
```suggestion
        // If there is no `expireTimestamp`, then it is an XHR object and needs to be parsed.
```

src/identity-utils.ts
```diff
@@ -17,74 +30,101 @@ export type IParseCachedIdentityResponse = (
...
 }
 
 export interface ICachedIdentityCall {
+    // https://go.mparticle.com/work/SQDSDKS-6672
     responseText: string;
     status: number;
     expireTimestamp: number;
 }
 
+// https://go.mparticle.com/work/SQDSDKS-6568
```
rmi22186, line 46: I'm not convinced this ticket clearly explains what should happen to `xhrIdentityResponseAdapter` during the clean up.  Consider adding to the description so that it is clear.

src/identity-utils.ts
```diff
@@ -17,74 +30,101 @@ export type IParseCachedIdentityResponse = (
...
+// Temporary adapter to convert the XMLHttpRequest response to the IIdentityResponse interface
+export const xhrIdentityResponseAdapter = (
+    possiblyXhr: XMLHttpRequest | IIdentityResponse
+): IIdentityResponse => {
+    if (possiblyXhr.hasOwnProperty('expireTimestamp')) {
+        // This is likely an IIdentityResponse object, just return it
```
rmi22186, line 52: It is not clear that the `IIdentityResponse` interface is cached.

Also - when you say `likely`, that implies it isn't the case 100% of the time, but if it goes down this code path, then it in fact *is* `IIdentityResponse`, so I think the comment should be explicit instead of saying `likely`.
```suggestion
        // If there is an `expireTimestamp`, it is an IIdentityResponse object, so just return it.  This indicates it was a previously cached value.
```

einsteinx2 (approved):
Live reviewed with Alex on Zoom

rmi22186 (approved):

PR #908, opened 2024-08-26
fix: Remove XHR References from Parse Identity Response

alexs-mparticle:
## Instructions

 ## Summary
 - Replaces usage of `XHttpRequest` with an IdentityResponse Object within `parseIdentityResponse` using the data of the Identity Response rather than having our logic tightly coupled to XHR.
- Uses a temporary xhrIdentityResponseAdapter to facility the conversion of an XHR response to an IdentityResponse object. This will conversion logic eventually be moved into IdentityApiClient in a future revision but is necessary to reduce the scope of changes in this PR.
- Splits up the scope of work originally started in #905 to reduce the surface area of the change and to make the code review easier to understand.
- This PR will be merged in with #905

 ## Testing Plan
 - Using a sample app, verify that Identity methods continue to function and are properly cached.

 - Closes https://go.mparticle.com/work/SQDSDKS-6658

### DIFF
--- a/src/identity.js
+++ b/src/identity.js
@@ -13,6 +13,7 @@ const { CacheIdentity } = FeatureFlags;
 const { Identify, Modify, Login, Logout } = IdentityMethods;
 import {
     generateDeprecationMessage,
+    HTTP_SUCCESS,
     isEmpty,
     isFunction,
     isObject,
@@ -1461,7 +1462,7 @@ export default function Identity(mpInstance) {
 
     // https://go.mparticle.com/work/SQDSDKS-6355
     this.parseIdentityResponse = function(
-        xhr,
+        identityResponse,
         previousMPID,
         callback,
         identityApiData,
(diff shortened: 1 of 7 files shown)


src/identity.js
```diff
@@ -1487,8 +1487,8 @@ export default function Identity(mpInstance) {
                 'Parsing "' + method + '" identity response from server'
             );
 
-            identityApiResult = xhr.responseText
-                ? JSON.parse(xhr.responseText)
+            identityApiResult = identityResponse.responseText
```
jared-dickman, line 1490: this looks like a case for `||` or `??`

src/identity.js
```diff
@@ -1514,14 +1514,16 @@ export default function Identity(mpInstance) {
                 );
             }
 
-            if (xhr.status === 200) {
+            if (identityResponse.status === 200) {
```
jared-dickman, line 1517: **suggestion** http success constant

rmi22186 (approved):
lgtm!

PR #914, opened 2024-09-04
ci: Update upload-artifact version in CI/CD

rmi22186:
## Instructions

 ## Summary
When doing a release, we came upon an error about `upload-artifacts` being outdated. We used v3 elsewhere in the workflows, this PR is to just update the remaining 2 instances.

 ## Testing Plan
NA - This PR is a workflow update and cannot be tested

 - Closes https://go.mparticle.com/work/SQDSDKS-6734


BrandonStalnaker (approved):
LGTM

PR #926, opened 2024-10-04
feat: Capture Integration Ids and assign to events

alexs-mparticle:
## Instructions

 ## Summary
 - Adds support to capture select Integration Ids passed as query parameters or cookies and add them to Event Custom Flags for processing.
 - Relevant Facebook Documentation: https://developers.facebook.com/docs/marketing-api/conversions-api/parameters/fbp-and-fbc

 ## Testing Plan
 - Using a sample app, pass in Facebook Click IDs via a query parameter, or Facebook Pixel ID as a cookie value and verify they appear in the Custom Flags for Events

 - Closes https://go.mparticle.com/work/SQDSDKS-6806

### DIFF
--- a/src/integrationCapture.ts
+++ b/src/integrationCapture.ts
@@ -3,7 +3,7 @@ import { Dictionary, queryStringParser, getCookies, getHref } from './utils';
 
 const integrationIdMapping: Dictionary<string> = {
     fbclid: 'Facebook.ClickId',
-    _fbp: 'FaceBook.BrowserId',
+    _fbp: 'Facebook.BrowserId',
 };
 
 export default class IntegrationCapture {
@@ -29,11 +29,16 @@ export default class IntegrationCapture {
      * Captures query parameters based on the integration ID mapping.
      */
     public captureQueryParams(): void {
-        const parsedQueryParams = queryStringParser(
-            getHref(),
-            Object.keys(integrationIdMapping)
-        );
+        const parsedQueryParams = this.getQueryParams();
         this.clickIds = { ...this.clickIds, ...parsedQueryParams };
+    }
+
+    /**
+     * Gets the query parameters based on the integration ID mapping.
+     * @returns {Dictionary<string>} The query parameters.
+     */
+    public getQueryParams(): Dictionary<string> {
+        return queryStringParser(getHref(), Object.keys(integrationIdMapping));
     }
 
     /**
--- a/src/integrationCapture.ts
+++ b/src/integrationCapture.ts
@@ -23,23 +23,23 @@ export const facebookClickIdProcessor: IntegrationCaptureProcessorFunction = (
     url: string,
     timestamp?: number,
 ): string => {
-    if (!clickId && !url) {
+    if (!clickId || !url) {
         return '';
     }
 
-    const urlPrefix = url?.split('//')
-    if (!urlPrefix) {
+    const urlSegments = url?.split('//')
+    if (!urlSegments) {
         return '';
     }
 
-    const urlParts = urlPrefix[1].split('/');
-    const hostParts = urlParts[0].split('.');
+    const urlParts = urlSegments[1].split('/');
+    const domainParts = urlParts[0].split('.');
     let subdomainIndex: number = 1;
 
     // The rules for subdomainIndex are for parsing the domain portion
     // of the URL for cookies, but in this case we are parsing the URL 
     // itself, so we can ignore the use of 0 for 'com'
-    if (hostParts.length >= 3) {
+    if (domainParts.length >= 3) {
         subdomainIndex = 2;
     }
 
--- a/src/integrationCapture.ts
+++ b/src/integrationCapture.ts
@@ -1,9 +1,41 @@
 import { SDKEventCustomFlags } from './sdkRuntimeModels';
 import { Dictionary, queryStringParser, getCookies, getHref } from './utils';
 
-const integrationIdMapping: Dictionary<string> = {
-    fbclid: 'Facebook.ClickId',
-    _fbp: 'Facebook.BrowserId',
+
+export const facebookClickIdProcessor = (clickId: string): string => {
+    if (!clickId) {
+        return '';
+    }
+
+    const clickIdParts = clickId.split('.');
+    if (clickIdParts.length === 4) {
+        return clickIdParts[3];
+    };
+
+    return clickId;
+};
+interface ProcessorFunction {
+    (clickId: string): string;
+}
+interface IntegrationIdMapping {
+    [key: string]: {
+        mappingValue: string;
+        processor?: ProcessorFunction;
+    };
+}
+
+const integrationMapping: IntegrationIdMapping = {
+    fbclid: {
+        mappingValue: 'Facebook.ClickId',
+        processor: facebookClickIdProcessor,
+    },
+    _fbp: {
+        mappingValue: 'Facebook.BrowserId',
+    },
+    _fbc: {
+        mappingValue: 'Facebook.ClickId',
+        processor: facebookClickIdProcessor,
+    },
 };
 
 export default class IntegrationCapture {
@@ -13,24 +45,31 @@ export default class IntegrationCapture {
      * Captures Integration Ids from cookies and query params and stores them in clickIds object
      */
     public capture(): void {
-        this.captureCookies();
-        this.captureQueryParams();
+        const queryParams = this.captureQueryParams() || {};
+        const cookies = this.captureCookies() || {};
+
+        if (queryParams['fbclid'] && cookies['_fbc']) {
+            // Exclude fbclid if _fbc is present
+            delete cookies['_fbc'];
+        }
+
+        this.clickIds = { ...this.clickIds, ...queryParams, ...cookies };
     }
 
     /**
      * Captures cookies based on the integration ID mapping.
      */
-    public captureCookies(): void {
-        const cookies = getCookies(Object.keys(integrationIdMapping));
-        this.clickIds = { ...this.clickIds, ...cookies };
+    public captureCookies(): Dictionary<string> {
+        const cookies = getCookies(Object.keys(integrationMapping));
+        return this.applyProcessors(cookies);
     }
 
     /**
      * Captures query parameters based on the integration ID mapping.
      */
-    public captureQueryParams(): void {
-        const parsedQueryParams = this.getQueryParams();
-        this.clickIds = { ...this.clickIds, ...parsedQueryParams };
+    public captureQueryParams(): Dictionary<string> {
+        const queryParams = this.getQueryParams();
+        return this.applyProcessors(queryParams);
     }
 
     /**
@@ -38,7 +77,7 @@ export default class IntegrationCapture {
      * @returns {Dictionary<string>} The query parameters.
      */
     public getQueryParams(): Dictionary<string> {
-        return queryStringParser(getHref(), Object.keys(integrationIdMapping));
+        return queryStringParser(getHref(), Object.keys(integrationMapping));
     }
 
     /**
@@ -53,11 +92,26 @@ export default class IntegrationCapture {
         }
 
         for (const [key, value] of Object.entries(this.clickIds)) {
-            const mappedKey = integrationIdMapping[key];
+            const mappedKey = integrationMapping[key]?.mappingValue;
             if (mappedKey) {
                 customFlags[mappedKey] = value;
             }
         }
         return customFlags;
     }
+
+    private applyProcessors(clickIds: Dictionary<string>): Dictionary<string> {
+        const processedClickIds: Dictionary<string> = {};
+
+        for (const [key, value] of Object.entries(clickIds)) {
+            const processor = integrationMapping[key]?.processor;
+            if (processor) {
+                processedClickIds[key] = processor(value);
+            } else {
+                processedClickIds[key] = value;
+            }
+        }
+
+        return processedClickIds;
+    };
 }
--- a/src/integrationCapture.ts
+++ b/src/integrationCapture.ts
@@ -1,41 +1,58 @@
-import { SDKEventCustomFlags } from "./sdkRuntimeModels";
-import { Dictionary } from "./utils";
+import { SDKEventCustomFlags } from './sdkRuntimeModels';
+import { Dictionary, queryStringParser, getCookies, getHref } from './utils';
 
-const integrationIdMapping = {
-    'ttclid': 'TikTok.ClickId',
-    'fbclid': 'Facebook.ClickId',
-    'fbp': 'FaceBook.BrowserId',
+const integrationIdMapping: Dictionary<string> = {
+    fbclid: 'Facebook.ClickId',
+    _fbp: 'FaceBook.BrowserId',
 };
 
-// TODO: Do we need to put this in the store?
 export default class IntegrationCapture {
-    public clickIds: Dictionary<string> = {};
+    public clickIds: Dictionary<string>;
 
-    public captureQueryParams(): void {
-        // TODO: Make this a utility with a fallback
-        const urlParams = new URLSearchParams(window.location.search);
-
-        Object.keys(integrationIdMapping).forEach((key) => {
-            const value = urlParams.get(key);
-            if (value) {
-                this.clickIds[key] = value;
-            }
-        });
-
-        console.log('captured params', this.clickIds);
+    /**
+     * Captures Integration Ids from cookies and query params and stores them in clickIds object
+     */
+    public capture(): void {
+        this.captureCookies();
+        this.captureQueryParams();
     }
 
+    /**
+     * Captures cookies based on the integration ID mapping.
+     */
+    public captureCookies(): void {
+        const cookies = getCookies(Object.keys(integrationIdMapping));
+        this.clickIds = { ...this.clickIds, ...cookies };
+    }
+
+    /**
+     * Captures query parameters based on the integration ID mapping.
+     */
+    public captureQueryParams(): void {
+        const parsedQueryParams = queryStringParser(
+            getHref(),
+            Object.keys(integrationIdMapping)
+        );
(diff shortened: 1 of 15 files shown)


src/mp-instance.js
```diff
@@ -1370,6 +1376,33 @@ function completeSDKInitialization(apiKey, config, mpInstance) {
...
+    //     mpInstance._Store.configuredClickIds['fbclid'] = fbClickId;
+    // }
+
+    // if (tikTokId) {
+    //     mpInstance._Store.configuredClickIds['ttclid'] = tikTokId;
+    // }
+    // *** End Quick Solution
+
+    // *** Long term solution
+
```
alexs-mparticle, line 1402: remove

src/integrationCapture.ts
```diff
@@ -0,0 +1,63 @@
...
+    public getClickIdsAsCustomFlags(): SDKEventCustomFlags {
+        const customFlags: SDKEventCustomFlags = {};
+
+        if (!this.clickIds) {
+            return customFlags;
+        }
+
+        for (const [key, value] of Object.entries(this.clickIds)) {
+            const mappedKey = integrationIdMapping[key];
+            if (mappedKey) {
```
jared-dickman, line 57: could mappedKey be 0 or an empty string?? in which case this might fail erroneously

src/serverModel.ts
```diff
@@ -320,6 +321,14 @@ export default function ServerModel(
             if (event.hasOwnProperty('toEventAPIObject')) {
                 eventObject = event.toEventAPIObject();
             } else {
+                let customFlags: SDKEventCustomFlags = {...event.customFlags};
+
+                // https://go.mparticle.com/work/SQDSDKS-5053
```
jared-dickman, line 326: **suggestion** put a link to this line in the ticket as well

test/jest/integration-capture.spec.ts
```diff
@@ -0,0 +1,132 @@
+import IntegrationCapture from "../../src/integrationCapture";
+
+function deleteAllCookies() {
+    document.cookie.split(';').forEach(cookie => {
+        const eqPos = cookie.indexOf('=');
+        const name = eqPos > -1 ? cookie.substring(0, eqPos) : cookie;
+        document.cookie = name + '=;expires=Thu, 01 Jan 1970 00:00:00 GMT';
```
jared-dickman, line 7: is this best practice? not sure, seems funky

jared-dickman: is setting the expires time to the past a shortcut for deleting a cookie?

test/src/tests-feature-flags.ts
```diff
@@ -15,6 +16,14 @@ declare global {
     }
 }
 
+function deleteAllCookies() {
```
jared-dickman, line 19: duplicated function? reduce it!

test/jest/integration-capture.spec.ts
alexs-mparticle: Yeah, you'd think Jest would have a better way of mocking or clearing cookies. 😦

test/src/tests-feature-flags.ts
alexs-mparticle: Good catch. I'll move it to our test utilities

src/integrationCapture.ts
alexs-mparticle: Unlikely as this is a hard coded JSON object, but I see your point.

src/integrationCapture.ts
```diff
@@ -0,0 +1,63 @@
+import { SDKEventCustomFlags } from './sdkRuntimeModels';
+import { Dictionary, queryStringParser, getCookies, getHref } from './utils';
+
+const integrationIdMapping: Dictionary<string> = {
+    fbclid: 'Facebook.ClickId',
```
rmi22186, line 5: I'd add a comment stating that these two facebook values are custom flags.

src/serverModel.ts
```diff
@@ -320,6 +321,14 @@ export default function ServerModel(
             if (event.hasOwnProperty('toEventAPIObject')) {
                 eventObject = event.toEventAPIObject();
             } else {
+                let customFlags: SDKEventCustomFlags = {...event.customFlags};
```
alexs-mparticle, line 324: Move this above `event.hasOwnProperty('toEventAPIObject')`

src/utils.ts
```diff
@@ -232,6 +232,87 @@ const mergeObjects = <T extends object>(...objects: T[]): T => {
...
+        }
+    }
+    return results;
+    };
+
+    // Parse cookies from document.cookie
+    const cookies = parseCookies();
+
+    // Filter cookies by keys if provided
+    return filterCookies(cookies, keys);
```
rmi22186, line 307: Great work with this function.  I think this can be a _bit_ more clear

```suggestion
    const parsedCookies: string[] = parseCookies();

    // Filter cookies by keys if provided
    return filterCookies(parsedCookies, keys);
```

brewdente (approved):
I did a read through with @alexs-mparticle and everything looks clean and pretty straightforward.  The tests are comprehensive and verified in his environment that it works.

src/integrationCapture.ts
```diff
@@ -0,0 +1,63 @@
...
+     */
+    public getQueryParams(): Dictionary<string> {
+        return queryStringParser(getHref(), Object.keys(integrationIdMapping));
+    }
+
+    /**
+     * Converts captured click IDs to custom flags for SDK events.
+     * @returns {SDKEventCustomFlags} The custom flags.
+     */
+    public getClickIdsAsCustomFlags(): SDKEventCustomFlags {
```
rmi22186, line 48: Are we confident that all click ids will be stored as custom flags? I understand that this is true for the current project, but I also know that we are doing it this way for certain reasons.  May be worth asking the team if the other click ids they are considering have a different means of being captured just for our knowledge, since I know there is potential for being integration attributes also.  Probably nothing to change for now though.

src/integrationCapture.ts
```diff
@@ -0,0 +1,63 @@
...
+        this.captureCookies();
+        this.captureQueryParams();
+    }
+
+    /**
+     * Captures cookies based on the integration ID mapping.
+     */
+    public captureCookies(): void {
+        const cookies = getCookies(Object.keys(integrationIdMapping));
+        this.clickIds = { ...this.clickIds, ...cookies };
```
rmi22186, line 25: I assume the idea for this is that there may be multiple times in which we need to call `capture()`, and so we want to ensure we are not losing previous click ids?

src/utils.ts
```diff
@@ -232,6 +232,87 @@ const mergeObjects = <T extends object>(...objects: T[]): T => {
...
+        }
+        return window.document.cookie.split(';').map(cookie => cookie.trim());
+    };
+
+    // Helper function to filter cookies by keys
+    const filterCookies = (cookies: string[], keys?: string[]): Dictionary<string> => {
+        const results: Dictionary<string> = {};
+    for (const cookie of cookies) {
+        const [key, value] = cookie.split('=');
+            if (!keys || keys.includes(key)) {
```
rmi22186, line 296:
```suggestion
      if (!keys || keys.includes(key)) {
```

src/utils.ts
```diff
@@ -232,6 +232,87 @@ const mergeObjects = <T extends object>(...objects: T[]): T => {
...
+
+    return {
+        get: function(key: string) {
+            return params[key];
+        },
+    };
+};
+
+// Get cookies as a dictionary
+const getCookies = (keys?: string[]): Dictionary<string> => {
```
rmi22186, line 282: `getCookies` seems to generically named given it is in a `utils` file.  Just given the name, it I would have expected it to just return cookies.  However, it is doing a lot more than just getting the cookies.  I think this is okay for now, but it's not clear to me if this method can be used elsewhere where we are getting and parsing cookies.

This is also true of queryString methods here.  They seem pretty specific to the integrationId feature, rather than being generic methods that are reusable.

Thoughts?

rmi22186 (changes requested):

test/jest/integration-capture.spec.ts
```diff
@@ -0,0 +1,125 @@
+import IntegrationCapture from "../../src/integrationCapture";
+import { deleteAllCookies } from "./utils";
+
+describe('Integration Capture', () => {
+    describe('constructor', () => {
+        it('should initialize with clickIds as undefined', () => {
```
rmi22186, line 6: I go back and forth with saying this should be undefined vs it being defined with a default value of {}.  But not a hill i'm going to die on :)

test/jest/utils.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+            window.document.cookie = 'foo=bar';
+            window.document.cookie = 'baz=qux';
+        });
+
+        afterEach(() => {
+            deleteAllCookies();
+        });
+
+        it('returns all cookies as an object', () => {
+
```
rmi22186, line 19:
```suggestion
```

test/jest/utils.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+            const originalWindow = global.window;
+            delete global.window;
+
+            expect(getCookies()).toEqual({});
+
+            global.window = originalWindow
+        });
+    });
+
+    describe('queryyStringParser', () => {
```
rmi22186, line 59: can you add 2 more tests (I think these are true statements):
1. returns an empty object if there are no query parameters
2. returns an object with all query params if no keys are passed

test/src/tests-feature-flags.ts
```diff
@@ -102,4 +106,66 @@ describe('feature-flags', function() {
...
+            window.document.cookie = 'foo=bar';
+            window.document.cookie = '_fbp=54321';
+            window.document.cookie = 'baz=qux';
+        });
+
+        afterEach(() => {
+            fetchMock.restore();
+            deleteAllCookies();
+            sinon.restore(); // Restore all stubs and spies
+
```
rmi22186, line 127:
```suggestion
```

rmi22186 (changes requested):
very minor nits. looks great!

src/integrationCapture.ts
```diff
@@ -1,9 +1,41 @@
...
+        return clickIdParts[3];
+    };
+
+    return clickId;
+};
+interface ProcessorFunction {
+    (clickId: string): string;
+}
+interface IntegrationIdMapping {
+    [key: string]: {
```
rmi22186, line 21: add new lines between the interfaces for spacing/readability

src/integrationCapture.ts
```diff
@@ -0,0 +1,117 @@
+import { SDKEventCustomFlags } from './sdkRuntimeModels';
+import { Dictionary, queryStringParser, getCookies, getHref } from './utils';
+
+
+export const facebookClickIdProcessor = (clickId: string): string => {
```
bowiec, line 5: It looks like this processing is always returning the `fbclid` value, but we need it to be returning the formatted `fbc` value - as in of the form `fb.1.1554763741205.AbCdEfGhIjKlMnOpQrStUvWxYz1234567890`.

The `_fbc` value should already be formatted like this so we don't need to process it at all and can use it as is. The `fbclid` value is the last segment (from the example, `AbCdEfGhIjKlMnOpQrStUvWxYz1234567890`). We need to populate the first three segments as described in the docs [here](https://developers.facebook.com/docs/marketing-api/conversions-api/parameters/fbp-and-fbc#2--format-clickid).

src/integrationCapture.ts
```diff
@@ -0,0 +1,117 @@
...
+    fbclid: {
+        mappingValue: 'Facebook.ClickId',
+        processor: facebookClickIdProcessor,
+    },
+    _fbp: {
+        mappingValue: 'Facebook.BrowserId',
+    },
+    _fbc: {
+        mappingValue: 'Facebook.ClickId',
+        processor: facebookClickIdProcessor,
```
bowiec, line 37: This isn't needed - we can use `_fbc` directly.

src/integrationCapture.ts
```diff
@@ -0,0 +1,117 @@
...
+
+    /**
+     * Captures Integration Ids from cookies and query params and stores them in clickIds object
+     */
+    public capture(): void {
+        const queryParams = this.captureQueryParams() || {};
+        const cookies = this.captureCookies() || {};
+
+        if (queryParams['fbclid'] && cookies['_fbc']) {
+            // Exclude fbclid if _fbc is present
```
bowiec, line 52: Seems like this should say `Exclude _fbc if fbclid is present`?

src/integrationCapture.ts
```diff
@@ -13,32 +45,39 @@ export default class IntegrationCapture {
      * Captures Integration Ids from cookies and query params and stores them in clickIds object
      */
     public capture(): void {
-        this.captureCookies();
-        this.captureQueryParams();
+        const queryParams = this.captureQueryParams() || {};
+        const cookies = this.captureCookies() || {};
+
+        if (queryParams['fbclid'] && cookies['_fbc']) {
+            // Exclude fbclid if _fbc is present
```
rmi22186, line 52: Please reference the doc where we get this info from.

src/integrationCapture.ts
```diff
@@ -1,9 +1,41 @@
...
+    (clickId: string): string;
+}
+interface IntegrationIdMapping {
+    [key: string]: {
+        mappingValue: string;
+        processor?: ProcessorFunction;
+    };
+}
+
+const integrationMapping: IntegrationIdMapping = {
```
rmi22186, line 27: Love this design pattern!

src/integrationCapture.ts
rmi22186: i think it's this https://developers.facebook.com/docs/marketing-api/conversions-api/parameters/fbp-and-fbc#retrieve-from-fbclid-url-query-parameter

src/integrationCapture.ts
```diff
@@ -0,0 +1,117 @@
...
+    };
+
+    return clickId;
+};
+interface ProcessorFunction {
+    (clickId: string): string;
+}
+interface IntegrationIdMapping {
+    [key: string]: {
+        mappingValue: string;
```
bowiec, line 22: Nit: This is more of a `mappedKey` or maybe `mappedFlag`

test/jest/integration-capture.spec.ts
```diff
@@ -122,4 +201,35 @@ describe('Integration Capture', () => {
...
+            const expectedClickId = '';
+
+            expect(facebookClickIdProcessor(null)).toEqual(expectedClickId);
+            expect(facebookClickIdProcessor(undefined)).toEqual(expectedClickId);
+            expect(facebookClickIdProcessor('')).toEqual(expectedClickId);
+
+            // @ts-ignore
+            expect(facebookClickIdProcessor(NaN)).toEqual(expectedClickId);
+            // @ts-ignore
+            expect(facebookClickIdProcessor(0)).toEqual(expectedClickId);
```
rmi22186, line 232: instead of the // @ts-ignore comment can you do `NaN as ProcessorFunction) and `0 as ProcessorFunction`?

rmi22186 (changes requested):

src/mp-instance.js
```diff
@@ -1336,6 +1338,10 @@ function completeSDKInitialization(apiKey, config, mpInstance) {
             mpInstance._ForwardingStatsUploader.startForwardingStatsTimer();
         }
 
+        if (mpInstance._Helpers.getFeatureFlag(CaptureIntegrationSpecificIds)) {
+            mpInstance._IntegrationCapture.capture();
```
bowiec, line 1342: Will this run after the FB SDK loads? If not, `_fbp`/`_fbc` may not be set yet

src/integrationCapture.ts
alexs-mparticle: Good catch. I think the naming of these two attributes confuses (or confounds) me.

src/integrationCapture.ts
```diff
@@ -1,45 +1,80 @@
...
+// - subdomainIndex is which domain the cookie is defined on ('com' = 0, 'example.com' = 1, 'www.example.com' = 2)
+// - creationTime is the UNIX time since epoch in milliseconds when the _fbc was stored. If you don't save the _fbc cookie, use the timestamp when you first observed or received this fbclid value
+// - <fbclid> is the value for the fbclid query parameter in the page URL.
+// https://developers.facebook.com/docs/marketing-api/conversions-api/parameters/fbp-and-fbc
+export const facebookClickIdProcessor: IntegrationCaptureProcessorFunction = (
+    clickId: string,
+    url: string,
+    timestamp?: number,
+): string => {
+    if (!clickId && !url) {
```
bowiec, line 26: Maybe `!clickId || !url`? As is, this would pass if click ID is unset and URL is set. We probably shouldn't send anything if clickID isn't set.

src/serverModel.ts
```diff
@@ -325,6 +325,7 @@ export default function ServerModel(
 
                 // https://go.mparticle.com/work/SQDSDKS-5053
                 if (mpInstance._Helpers.getFeatureFlag && mpInstance._Helpers.getFeatureFlag(Constants.FeatureFlags.CaptureIntegrationSpecificIds)) {
+                    mpInstance._IntegrationCapture.capture();
```
rmi22186, line 328: Add a comment to explain why you put this here.

src/utils.ts
```diff
@@ -275,6 +285,13 @@ const queryStringParserFallback = (url: string): URLSearchParamsFallback => {
         get: function(key: string) {
             return params[key];
         },
+        forEach: function(callback: (value: string, key: string) => void) {
+            for (var key in params) {
+                if (params.hasOwnProperty(key)) {
+                    callback(params[key], key);
+                }
+            }
```
rmi22186, line 293: good catch that this wasn't supported previously

rmi22186 (changes requested):
tiny nits

rmi22186 (approved):

PR #929, opened 2024-10-08
fix: Handle cases where window.screen is unavailable

alexs-mparticle:
## Instructions

 ## Summary
 - When customers use the Web SDK in an environment that does not support `window`, such as in a service worker, batch creation will throw an exception when trying to detect the screen size.

 ## Testing Plan
 - Attempt to run Web SDK in a service worker

 - Closes https://go.mparticle.com/work/SQDSDKS-6847


rmi22186 (changes requested):
PR this against `development`, otherwise lgtm

rmi22186 (approved):

PR #930, opened 2024-10-09
tests: Fix remaining tests from Identity Refactor

rmi22186:
## Instructions
Refactor tests to use `waitForCondition` and other helpers to get tests to pass now that identity is async.

 ## Testing Plan
 - [X] Was this tested locally? If not, explain why.
 - {explain how this has been tested, and what, if any, additional testing should be done}

 ## Reference Issue (For mParticle employees only.  Ignore if you are an outside contributor)
 - Closes https://go.mparticle.com/work/SQDSDKS-6848

PR #931, opened 2024-10-11
tests: Fix remaining tests from XHR to fetch migration

rmi22186:
## Instructions

 ## Summary
Resolve all breaking tests due to XHR --> window.fetch refactor

 ## Testing Plan
 - [X] Was this tested locally? If not, explain why. Automated test suite run locally

 - Closes https://go.mparticle.com/work/SQDSDKS-6848


rmi22186:
1st commit is approved from @alexs-mparticle from [this PR](https://github.com/mParticle/mparticle-web-sdk/pull/930). We squashed them all into a single commit in order to make rebasing easier.  Then added a commit from breaking tests which used the previous `mockServer` (xhr) and migrated those tests to use `fetchMock`.

alexs-mparticle (approved):

PR #933, opened 2024-10-11
fix: Move ready queue processing after identity request

alexs-mparticle:
## Instructions

 ## Summary
 - Changes the order of when we purge the ready queue so that it comes after the identify request is attempted.
 - Will also continue to clear out the ready queue as part of the initial init flow if an identify request is unnecessary, such as if there is an existing session.
 - This attempts to solve an issue where a User Identity Change should occur before event logs are queued in the ready queue, but will be queued after because the ready queue is purged before the identify request returns.

 ## Testing Plan

 - Closes https://go.mparticle.com/work/SQDSDKS-6835


src/mp-instance.js
```diff
@@ -1361,15 +1362,17 @@ function completeSDKInitialization(apiKey, config, mpInstance) {
         );
     }
 
-    mpInstance._Store.isInitialized = true;
+    // If for some reason the identity call inside of Session Manager does not run
+    // we should clear out the ready queue.
```
rmi22186, line 1366: Your PR description appropriately states when this is the case, rather than "for some reason", so I'd turn that into a comment.

"Will also continue to clear out the ready queue as part of the initial init flow if an identify request is unnecessary, such as if there is an existing session".

src/pre-init-utils.ts
```diff
@@ -0,0 +1,38 @@
+import { isEmpty, isFunction } from './utils';
+
+export const processReadyQueue = (readyQueue): Function[] => {
+    if (!isEmpty(readyQueue)) {
+        readyQueue.forEach(readyQueueItem => {
+            if (isFunction(readyQueueItem)) {
+                // debugger;
```
rmi22186, line 7:
```suggestion
```

src/pre-init-utils.ts
```diff
@@ -0,0 +1,38 @@
...
+                readyQueueItem();
+            } else if (Array.isArray(readyQueueItem)) {
+                // debugger;
```
rmi22186, line 10:
```suggestion
```

test/jest/pre-init-utils.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+            const readyQueue = [['fakeFunction']];
+            processReadyQueue(readyQueue);
+            expect(functionSpy).toHaveBeenCalled();
+        });
+
+        it('should process functions passed as arrays with arguments', () => {
+            const functionSpy = jest.fn();
+            (window.mParticle as any) = {
+                fakeFunction: functionSpy,
+                args: () => {},
```
rmi22186, line 32: Is this `args` function necessary? I haven't pulled this code down, but as I read through the `processReadyQueue` function, the only times I see `args` is where it references the item in the array, as opposed to it being a key on `window.mParticle`.

test/jest/pre-init-utils.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+            const functionSpy = jest.fn();
+            (window.mParticle as any) = {
+                fakeFunction: {
+                    anotherFakeFunction: functionSpy,
+                },
+            };
+            const readyQueue = [['fakeFunction.anotherFakeFunction', 'foo']];
+            processReadyQueue(readyQueue);
+            expect(functionSpy).toHaveBeenCalledWith('foo');
+        });
```
rmi22186, line 49: is this multiple methods?  it seems like it's a single method, but 1 layer deeper.
Should the test instead be:

```suggestion
        it('should process arrays passed as arguments with multiple methods', () => {
            const functionSpy = jest.fn();
            const functionSpy2 = jest.fn();
            (window.mParticle as any) = {
                fakeFunction: functionSpy,
                anotherFakeFunction: functionSpy2,
            };
            const readyQueue = [['fakeFunction', 'foo'], [anotherFakeFunction', 'foo2']];
            processReadyQueue(readyQueue);
            expect(functionSpy).toHaveBeenCalledWith('foo');
            expect(functionSpy).toHaveBeenCalledWith('foo');
        });
```

(This may not run as I just did the suggestion inline here, but I think you get the point)

rmi22186: or actually I think I see what you are doing here. It's just that the test name is incorrect.

I think the suggestion I gave above can be its own test though so actually test multiple methods inside of the process queue when they are in array form.

test/src/tests-forwarders.js
```diff
@@ -3109,22 +3103,13 @@ describe('forwarders', function() {
         window.mParticle.config.rq = [];
 
         mParticle.init(apiKey, window.mParticle.config);
-
-        waitForCondition(() => {
-            return (
-                window.mParticle.getInstance()?._Store?.identityCallInFlight === false
-            );
-        })
-        .then(() => {
```
rmi22186, line 3118: thanks for refactoring this :)

test/jest/pre-init-utils.spec.ts
alexs-mparticle: I'm open to discussing this further. Adding `args` here was the only way I could get it to hit that specific codepath in my tests.

test/src/tests-identities-attributes.ts
```diff
@@ -1223,6 +1237,135 @@ describe('identities and attributes', function() {
...
+            },
+        };
+
+        mParticle.init(apiKey, window.mParticle.config);
+
+        mParticle.logEvent('Test Event 1');
+        mParticle.logEvent('Test Event 2');
+        mParticle.logEvent('Test Event 3');
+
+        expect(mParticle.getInstance()._preInit.readyQueue.length).to.equal(3);
```
rmi22186, line 1323: this test is identical to the test above here but without the readyqueue assertion.  however, i imagine that the above test would also pass if you put this assertion there.  is there a way to differentiate the two?

rmi22186 (changes requested):
one last comment, but looks good otherwise

test/jest/pre-init-utils.spec.ts
rmi22186: let's hop on a zoom about this to walk through

test/src/tests-identities-attributes.ts
alexs-mparticle: I think the two tests should represent the workflow with and without the ready queue, so technically they would be nearly identical. I made a change to the other test to make it a bit more clear. How does that look?

rmi22186 (approved):
lgtm!

test/jest/pre-init-utils.spec.ts
rmi22186: hm. when i comment out line 32, it passes for me.  see screenshot below:

alexs-mparticle: It's weird but commenting out `arg` goes to a different code path that has the same result. It's likely a redundant conditional and could use some clean up. We should definitely sync up on the use cases to make sure nothing is missed.

Issue #840, opened 2024-02-01
Crash when calling upload()


paulrehkugler:
#### Summary

If you call `upload()` in the SDK before queueing an event for processing, the SDK will throw a runtime exception.

#### Details

The `uploader` may be `null` when called on this line. https://github.com/mParticle/mparticle-web-sdk/blob/4740596cb9f19e968e74c4d3c103d50c846d0888/src/mp-instance.js#L577

This is because it has not yet been initialized within `queueEventForBatchedUpload` on the `APIClient`. https://github.com/mParticle/mparticle-web-sdk/blob/4740596cb9f19e968e74c4d3c103d50c846d0888/src/apiClient.ts#L39-L50

The `uploader` is also listed as a nullable type in its type definition. https://github.com/mParticle/mparticle-web-sdk/blob/4740596cb9f19e968e74c4d3c103d50c846d0888/src/apiClient.ts#L13

#### Proposed fix

```ts
// if your JS toolchain supports the optional chaining operator
self._APIClient.uploader?.prepareAndUpload(false, false); 

// if your JS toolchain does not support the optional chaining operator
if (self._APIClient.uploader) {
  self._APIClient.uploader.prepareAndUpload(false, false);
}
```

rmi22186:
Thanks for flagging this @paulrehkugler!.  Our JS toolchain _does_ support the optional chaining operator.  We can make a ticket and add to our backlog, but since you have a proposed fix and since we are open source, you could also open a PR to get credit for it if you are comfortable with that :).  Let me know what you prefer.

rmi22186:
I do think think that a workaround for not is wrapping the `mParticle.upload()` in the `mParticle.ready` function would also work for now before this gets worked on.  You can view docs here for an example: https://docs.mparticle.com/developers/sdk/web/initialization/#listen-for-sdk-initialization

alexs-mparticle:
Marking this as closed as #938 will be released as part of [v2.31.1](https://github.com/mParticle/mparticle-web-sdk/releases/tag/v2.31.1).

If this is still not resolved, please reopen this issue or contact support@mparticle.com.

Thank you!

PR #938, opened 2024-10-22
fix: prevent crash when calling upload() early

mmustafa-tse:
## Summary
 - github issue with description [here](https://github.com/mParticle/mparticle-web-sdk/issues/840)

 ## Testing Plan
 - [Y] Was this tested locally? If not, explain why.
 - Tested with local overrides

 - Closes https://mparticle-eng.atlassian.net/browse/SQDSDKS-6164

### DIFF
--- a/src/mp-instance.js
+++ b/src/mp-instance.js
@@ -590,7 +590,7 @@ export default function mParticleInstance(instanceName) {
                     Constants.NativeSdkPaths.Upload
                 );
             } else {
-                self._APIClient.uploader.prepareAndUpload(false, false);
+                self._APIClient.uploader?.prepareAndUpload(false, false);
             }
         }
     };


src/mp-instance.js
```diff
@@ -590,7 +590,7 @@ export default function mParticleInstance(instanceName) {
                     Constants.NativeSdkPaths.Upload
                 );
             } else {
-                self._APIClient.uploader.prepareAndUpload(false, false);
+                self._APIClient.uploader?.prepareAndUpload(false, false);
```
alexs-mparticle, line 593: Can you write a test that catches this?

mmustafa-tse: just added the test case, thnx!

rmi22186: there exists a case where _APIClient won't yet be available, when there is a /config call in a self hosted world, so let's add a ? to _APIClient also: 
```suggestion
                self._APIClient?.uploader?.prepareAndUpload(false, false);
```

test/src/tests-batchUploader.ts
```diff
@@ -329,6 +329,21 @@ describe('batch uploader', () => {
...
+
+                window.mParticle.init(apiKey, window.mParticle.config);
+
+                const mpInstance = window.mParticle.getInstance();
+                const uploader = mpInstance._APIClient.uploader;
+
+                expect(uploader).to.equal(null)
+
+                // calling upload manually on an a null uploader should not result in an error
+                window.mParticle.upload();
```
alexs-mparticle, line 344: I think you should add an assertion to verify that an error actually doesn't throw.

IIRC the format should be something like:
`expect(() => { window.mParticle.upload()}).not.to.throw(new Error('<ERROR MESSAGE>'))`
or
`expect(() => { window.mParticle.upload()}).not.to.throw('<ERROR MESSAGE>')`

I think the gist is to make sure we're actually catching the right error message or to make sure throwing an error doesn't suspend other testing.

I would timebox this unless @rmi22186 feels strongly about it as well.

rmi22186: yea - i actually was thinking the same thing, assuming that if an error threw, perhaps the test would still pass since there is no `expect`.  however, i did pull the code down yesterday to check and it throws an error and the test fails, so at least it's not a silent error.

here's a recommendation i found online as an example of how this would work, so try this, and if it doesn't work, I think we are good to go with what you currently have, and just add a comment above it that says something like


```
const { expect } = require('chai');
const sinon = require('sinon');

describe('MyFunction', () => {
    it('should not throw an error', () => {
        const myFunction = sinon.stub().returns(true); // Stub the function
        
        expect(() => myFunction()).to.not.throw();
    });
});
```

and if that doesn't work, then put a comment explaining why there is not `expect`
```
 // the following does not throw an error, but if it did, the test would fail because the error is not silent
or something like that. 
```

mmustafa-tse: I was able to get it working and not throw the error with a regex matching a part of the message while specifying the error type to be TypeError, tested on both before and after the fix and it worked as expected

rmi22186 (approved):
I pulled the code down to run the new test undoing the source code change, and it works great

PR #939, opened 2024-10-24
fix: Support capturing click ids as custom flags in commerce events

rmi22186:
## Instructions

 ## Summary
- Commerce events were having the custom flags for click ids supplanted completely by custom flags that were being passed from the user. This PR fixes that. Instead we `extend` the custom flags to include both the custom flags created by `createEventObject` as well as what is passed by the user, with the flags passed by the user taken as priority.

 ## Testing Plan
Added unit tests for ecommerce and page view, as well as an additional for custom events.
Manually tested regular page events, pageViews, and commerce events with and without a custom flag of `Facebook.ClickId`.  All behavior working as expected (passed in custom flag takes priority over captured query param)

 - Closes https://go.mparticle.com/work/SQDSDKS-6889

### DIFF
--- a/src/ecommerce.js
+++ b/src/ecommerce.js
@@ -540,8 +540,8 @@ export default function Ecommerce(mpInstance) {
             baseEvent.CurrencyCode = mpInstance._Store.currencyCode;
             baseEvent.ShoppingCart = [];
             baseEvent.CustomFlags = {
-                ...baseEvent.CustomFlags,
-                ...customFlags,
+                ...(baseEvent.CustomFlags || {}),
+                ...(customFlags || {}),
             };
 
             return baseEvent;
--- a/src/ecommerce.js
+++ b/src/ecommerce.js
@@ -524,7 +524,6 @@ export default function Ecommerce(mpInstance) {
     };
 
     this.createCommerceEventObject = function(customFlags, options) {
-        const { extend } = mpInstance._Helpers;
         var baseEvent;
 
         mpInstance.Logger.verbose(
@@ -540,7 +539,10 @@ export default function Ecommerce(mpInstance) {
 
             baseEvent.CurrencyCode = mpInstance._Store.currencyCode;
             baseEvent.ShoppingCart = [];
-            baseEvent.CustomFlags = extend(baseEvent.CustomFlags, customFlags);
+            baseEvent.CustomFlags = {
+                ...baseEvent.CustomFlags,
+                ...customFlags,
+            };
 
             return baseEvent;
         } else {
--- a/src/ecommerce.js
+++ b/src/ecommerce.js
@@ -524,6 +524,7 @@ export default function Ecommerce(mpInstance) {
     };
 
     this.createCommerceEventObject = function(customFlags, options) {
+        const { extend } = mpInstance._Helpers;
         var baseEvent;
 
         mpInstance.Logger.verbose(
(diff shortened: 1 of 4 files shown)


test/src/tests-integration-capture.ts
```diff
@@ -64,4 +64,206 @@ describe('Integration Capture', () => {
...
         });
     });
+
+    it('should add captured integrations to event custom flags, prioritizing passed in custom flags', async () => {
+        await waitForCondition(hasIdentifyReturned);
+        window.mParticle.logEvent(
+            'Test Event',
+            mParticle.EventType.Navigation,
+            { mykey: 'myvalue' },
+            {'Facebook.ClickId': 'passed-in',}
```
alexs-mparticle, line 74:
```suggestion
            { 'Facebook.ClickId': 'passed-in' },
```

test/src/tests-integration-capture.ts
```diff
@@ -64,4 +64,206 @@ describe('Integration Capture', () => {
...
+        );
+
+        const testEvent = findEventFromRequest(fetchMock.calls(), 'Test Event');
+
+        const initialTimestamp = window.mParticle.getInstance()._IntegrationCapture.initialTimestamp;
```
alexs-mparticle, line 79: I see you're pulling this value but not using it anywhere.

test/src/tests-integration-capture.ts
```diff
@@ -64,4 +64,206 @@ describe('Integration Capture', () => {
...
+            'Facebook.BrowserId': '54321',
+        });
+    });
+
+    it('should add captured integrations to page view custom flags, prioritizing passed in custom flags', async () => {
+        await waitForCondition(hasIdentifyReturned);
+        window.mParticle.logPageView(
+            'Test Page View',
+            {'foo-attr': 'bar-attr'},
+            {'Facebook.ClickId': 'passed-in',}
```
alexs-mparticle, line 116:
```suggestion
            {'Facebook.ClickId': 'passed-in'},
```

test/src/tests-integration-capture.ts
```diff
@@ -8,7 +8,7 @@ const { waitForCondition, fetchMockSuccess, deleteAllCookies, findEventFromReque
 
 const mParticle = window.mParticle;
 
-describe('Integration Capture', () => {
+describe.only('Integration Capture', () => {
```
alexs-mparticle, line 11:
```suggestion
describe('Integration Capture', () => {
```

alexs-mparticle (changes requested):

test/src/tests-integration-capture.ts
rmi22186, line 79:
```suggestion
```

src/ecommerce.js
```diff
@@ -539,7 +540,7 @@ export default function Ecommerce(mpInstance) {
 
             baseEvent.CurrencyCode = mpInstance._Store.currencyCode;
             baseEvent.ShoppingCart = [];
-            baseEvent.CustomFlags = customFlags;
+            baseEvent.CustomFlags = extend(baseEvent.CustomFlags, customFlags);
```
alexs-mparticle, line 543: Can we use spread or the existing merge objects util? I'd like to get rid of the extend helper.

alexs-mparticle (approved):

PR #940, opened 2024-10-26
refactor: Add Type interfaces for cookie sync manager

alexs-mparticle:
## Instructions

 ## Summary
 - Introduces type interfaces for Cookie Sync Manager

 ## Testing Plan
 - N/A Does not change any actual functionality

 - Closes https://go.mparticle.com/work/SQDSDKS-4800

### DIFF
--- a/src/cookieSyncManager.interfaces.ts
+++ b/src/cookieSyncManager.interfaces.ts
@@ -13,7 +13,7 @@ export interface ICookieSyncManager {
         mpidIsNotInCookies: boolean,
         requiresConsent: boolean
     ) => void;
-    privatereplaceAmpWithAmpersand: (string: string) => string;
+    replaceAmpWithAmpersand: (string: string) => string;
     replaceMPID: (string, mpid: MPID) => string;
 
     /**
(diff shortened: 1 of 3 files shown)


src/cookieSyncManager.interfaces.ts
```diff
@@ -0,0 +1,23 @@
...
+        url: string,
+        moduleId: number,
+        mpid: MPID,
+        cookieSyncDates: Dictionary<number>,
+        filteringConsentRuleValues: IConsentRules,
+        mpidIsNotInCookies: boolean,
+        requiresConsent: boolean
+    ) => void;
+    replaceAmpWithAmpersand: (string: string) => string;
+    replaceMPID: (string, mpid: MPID) => string;
```
rmi22186, line 17:
```suggestion
    replaceMPID: (string: string, mpid: MPID) => string;
```

rmi22186 (changes requested):
tiny nit.  otherwise LGTM!

rmi22186 (approved):

PR #944, opened 2024-10-31
refactor: Migrate CookieSyncManager to TypeScript

alexs-mparticle:
## Instructions

 ## Summary
 - Migrate CookieSyncManager to TypeScript and add Jest Unit Tests

 ## Testing Plan
 - Run automated tests and verify that they continue to pass

 - Closes https://go.mparticle.com/work/SQDSDKS-5084

### DIFF
--- a/src/consent.ts
+++ b/src/consent.ts
@@ -38,6 +38,10 @@ export interface SDKConsentApi {
     createConsentState: (consentState?: ConsentState) => ConsentState;
     ConsentSerialization: IConsentSerialization;
     createPrivacyConsent: ICreatePrivacyConsentFunction;
+    isEnabledForUserConsent: (
+        consentRules: IConsentRules,
+        user: IMParticleUser
+    ) => boolean;
 }
 
 export interface IConsentSerialization {
--- a/src/cookieSyncManager.interfaces.ts
+++ b/src/cookieSyncManager.interfaces.ts
@@ -6,7 +6,7 @@ export interface ICookieSyncManager {
     attemptCookieSync: (
         previousMPID: MPID,
         mpid: MPID,
-        mpidIsNotInCookies: boolean
+        mpidIsNotInCookies?: boolean
     ) => void;
     performCookieSync: (
         url: string,
--- a/src/cookieSyncManager.interfaces.ts
+++ b/src/cookieSyncManager.interfaces.ts
@@ -13,11 +13,4 @@ export interface ICookieSyncManager {
         mpidIsNotInCookies: boolean,
         requiresConsent: boolean
     ) => void;
-    replaceAmpWithAmpersand: (string: string) => string;
-    replaceMPID: (string: string, mpid: MPID) => string;
-
-    /**
-     * @deprecated replaceAmp has been deprecated, use replaceAmpersandWithAmp instead
-     */
-    replaceAmp: (string: string) => string;
 }
--- a/src/cookieSyncManager.interfaces.ts
+++ b/src/cookieSyncManager.interfaces.ts
@@ -3,6 +3,19 @@ import { Dictionary } from './utils';
 import { IConsentRules } from './consent';
 
 export type CookieSyncDates = Dictionary<number>; 
+
+export interface IPixelConfiguration {
+    name?: string;
+    moduleId: number;
+    esId?: number;
+    isDebug?: boolean;
+    isProduction?: boolean;
+    settings: Dictionary<string>;
+    frequencyCap: number;
+    pixelUrl: string;
+    redirectUrl: string;
+    filteringConsentRuleValues?: IConsentRules;
+}
 export interface ICookieSyncManager {
     attemptCookieSync: (
         previousMPID: MPID,
--- a/src/cookieSyncManager.interfaces.ts
+++ b/src/cookieSyncManager.interfaces.ts
@@ -1,16 +1,25 @@
-import { MPID } from "@mparticle/web-sdk";
-import { Dictionary } from "./utils";
-import { IConsentRules } from "./consent";
+import { MPID } from '@mparticle/web-sdk';
+import { Dictionary } from './utils';
+import { IConsentRules } from './consent';
 
 export interface ICookieSyncManager {
-    attemptCookieSync: (previousMPID: MPID, mpid: MPID, mpidIsNotInCookies: boolean) => void;
+    attemptCookieSync: (
+        previousMPID: MPID,
+        mpid: MPID,
+        mpidIsNotInCookies: boolean
+    ) => void;
     performCookieSync: (
         url: string,
-        moduleId: number,
+        moduleId: string,
         mpid: MPID,
         cookieSyncDates: Dictionary<number>,
         filteringConsentRuleValues: IConsentRules,
         mpidIsNotInCookies: boolean,
         requiresConsent: boolean
     ) => void;
+    combineUrlWithRedirect: (
+        mpid: MPID,
+        pixelUrl: string,
+        redirectUrl: string
+    ) => string;
 }
--- a/src/cookieSyncManager.interfaces.ts
+++ b/src/cookieSyncManager.interfaces.ts
@@ -2,6 +2,7 @@ import { MPID } from '@mparticle/web-sdk';
 import { Dictionary } from './utils';
 import { IConsentRules } from './consent';
 
+export type CookieSyncDates = Dictionary<number>; 
 export interface ICookieSyncManager {
     attemptCookieSync: (
         previousMPID: MPID,
@@ -12,7 +13,7 @@ export interface ICookieSyncManager {
         url: string,
         moduleId: string,
         mpid: MPID,
-        cookieSyncDates: Dictionary<number>,
+        cookieSyncDates: CookieSyncDates,
         filteringConsentRuleValues: IConsentRules,
         mpidIsNotInCookies: boolean,
         requiresConsent: boolean
--- a/src/cookieSyncManager.ts
+++ b/src/cookieSyncManager.ts
@@ -8,14 +8,6 @@ import { IConsentRules } from './consent';
 const { Messages } = Constants;
 const { InformationMessages } = Messages;
 
-interface PixelConfig {
-    moduleId: string;
-    frequencyCap: number;
-    pixelUrl: string;
-    redirectUrl: string;
-    filteringConsentRuleValues: IConsentRules;
-}
-
 export const DAYS_IN_MILLISECONDS = 1000 * 60 * 60 * 24;
 
 const hasFrequencyCapExpired = (
@@ -107,9 +99,7 @@ export default function CookieSyncManager(
                     persistence[mpid].csd = {};
                 }
 
-                const lastSyncDateForModule = persistence[mpid].csd[moduleId]
-                    ? persistence[mpid].csd[moduleId]
-                    : null;
+                const lastSyncDateForModule = persistence[mpid].csd[moduleId] || null;
 
                 // Check to see if we need to refresh cookieSync
                 if (hasFrequencyCapExpired(lastSyncDateForModule, frequencyCap)) {
@@ -123,7 +113,6 @@ export default function CookieSyncManager(
                         requiresConsent
                     );
                 }
-
             }
         });
     };
--- a/src/cookieSyncManager.ts
+++ b/src/cookieSyncManager.ts
@@ -14,6 +14,10 @@ const hasFrequencyCapExpired = (
     frequencyCap: number,
     lastSyncDate?: number,
 ): boolean => {
+    if (!lastSyncDate) {
+        return true;
+    }
+
     return (
         new Date().getTime() >
         new Date(lastSyncDate).getTime() + frequencyCap * DAYS_IN_MILLISECONDS
@@ -61,7 +65,6 @@ export default function CookieSyncManager(
             // Url for cookie sync pixel
             const pixelUrl = replaceAmpWithAmpersand(pixelSettings.pixelUrl);
 
-            // TODO: Document requirements for redirectUrl
             const redirectUrl = pixelSettings.redirectUrl
                 ? replaceAmpWithAmpersand(pixelSettings.redirectUrl)
                 : null;
@@ -72,8 +75,6 @@ export default function CookieSyncManager(
                 redirectUrl
             );
 
-            // TODO: Is there a historic reason for checking for previousMPID?
-            //       it does not appear to be passed in anywhere
             if (previousMPID && previousMPID !== mpid) {
                 if (persistence && persistence[mpid]) {
                     if (!persistence[mpid].csd) {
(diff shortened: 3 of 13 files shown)


test/jest/cookieSyncManager.spec.ts
```diff
@@ -0,0 +1,454 @@
...
+                testMPID,
+                {},
+                undefined,
+                true,
+                false, 
+            );
+        });
+
+        // QUESTION: What is the purpose of this code path?
+        it('should call performCookieSync with mpid if previousMpid and mpid do not match', () => {
```
rmi22186, line 119: If a cookie sync was recently done, it won't be performed again until a certain time interval (i thinkt his varies per partner).  So if MPID1 is refreshing the page, it will do a cookie sync once until a certain time passes.

If a user logs out, and then logs in, then we would immediately know it's a different user and not have to use that logic since the new user hasn't done a cookie sync yet.

As I write that all out, I realize there is room for improvement.  There is a chance that if MPID1 logs in, cookie syncs, then logs in as MPID2, and cookie syncs, then logs in again as MPID1, i can see it wouldn't consider the time and ONLY looks at MPID change before performing a cookie sync.  Can you make a ticket to address this scenario?  It's a very minor optimization.

test/jest/cookieSyncManager.spec.ts
```diff
@@ -0,0 +1,454 @@
...
+                5,
+                testMPID,
+                {},
+                undefined,
+                true,
+                false, 
+            );
+        });
+
+        it('should call performCookieSync with mpid and csd if previousMpid and mpid do not match', () => {
```
rmi22186, line 147: This test name and the previous test name make it sound like the same test.  but this one has stuff in CSD.  This one is probably more realistic because testMPID1 will have a CSD, before logging in to a new MPID.  So I think you can remove the previous test.

test/jest/cookieSyncManager.spec.ts
```diff
@@ -0,0 +1,454 @@
...
+                    5: 1234567890
+                },
+                undefined,
+                true,
+                false, 
+            );
+        });
+
+        // QUESTION: What is the purpose of this code path?
+        it('should call performCookieSync with mpid if previousMpid and mpid match', () => {
```
rmi22186, line 179: As i mentioned in the long previous comment from the first question, seems like this code can be refactored to better represent when a cookie sync should happen.

1. when cookie sync has never happened yet (assuming it should based on whatever filters/privacy there are)
2. when cookie sync allotted time has lapsed

test/jest/cookieSyncManager.spec.ts
```diff
@@ -0,0 +1,454 @@
...
+                5,
+                testMPID,
+                {},
+                undefined,
+                true,
+                false, 
+            ); 
+        });
+
+        it('should perform a cookie sync if lastSyncDateForModule is past the frequency cap', () => {
```
rmi22186, line 207:
```suggestion
        it('should perform a cookie sync if lastSyncDateForModule is passed the frequency cap', () => {
```

test/jest/cookieSyncManager.spec.ts
rmi22186: The whole MPID equality portion of code can _probably_ be removed because we'd check for criteria 1 and 2 above for whatever the current MPID is

test/jest/cookieSyncManager.spec.ts
```diff
@@ -0,0 +1,454 @@
...
+                {
+                    5: cookieSyncDateInPast 
+                },
+                undefined,
+                true,
+                false, 
+            );
+        });
+
+        it('should perform a cookie sync if lastSyncDateForModule is past the frequency cap even if csd is empty', () => {
```
rmi22186, line 243: If CSD is empty, it means it has never been performed yet.  So there shouldn't be a code path for this.  I think this will be part of the code refactor.

test/jest/cookieSyncManager.spec.ts
```diff
@@ -0,0 +1,454 @@
...
+            const now = new Date().getTime();
+
+            const mockSDK = ({
```
rmi22186, line 246: Super minor nit.  maybe this (and elsewhere in the file) should be `mockMPInstance` instead?

test/jest/cookieSyncManager.spec.ts
```diff
@@ -0,0 +1,454 @@
...
+                    webviewBridgeEnabled: false,
+                    pixelConfigurations: [pixelSettings],
+                },
+                _Persistence: {
+                    setCookieSyncDates: jest.fn(),
+                    getPersistence: jest.fn(),
+                    saveUserCookieSyncDatesToPersistence: jest.fn(),
+                },
+                _Consent: {
+                    isEnabledForUserConsent: jest.fn().mockReturnValue(true),
```
rmi22186, line 422: should this be false?  isEnabled being true makes me think that the cookie sync would happen, and so the value for cookieSyncDates[42] should be defined.  What am i missing here?

rmi22186 (changes requested):
the jest tests are awesome.  THANK YOU.  minor nits mostly i'd say, and some commentary.

src/consent.ts
```diff
@@ -38,6 +38,10 @@ export interface SDKConsentApi {
     createConsentState: (consentState?: ConsentState) => ConsentState;
     ConsentSerialization: IConsentSerialization;
     createPrivacyConsent: ICreatePrivacyConsentFunction;
+    isEnabledForUserConsent: (
+        consentRules: IConsentRules,
+        user: IMParticleUser
+    ) => boolean;
```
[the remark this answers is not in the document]
alexs-mparticle: Updated the ticket and added screenshot.

src/cookieSyncManager.interfaces.ts
```diff
@@ -1,16 +1,25 @@
...
-    attemptCookieSync: (previousMPID: MPID, mpid: MPID, mpidIsNotInCookies: boolean) => void;
+    attemptCookieSync: (
+        previousMPID: MPID,
+        mpid: MPID,
+        mpidIsNotInCookies: boolean
+    ) => void;
     performCookieSync: (
         url: string,
-        moduleId: number,
+        moduleId: string,
```
rmi22186, line 13: when i pull a config down that has pixelConfigs in it, [such as this one ](https://jssdkcdns.mparticle.com/JS/v2/beab3f4d34281d45bfcdbbd7eb21c083/config?env=0), it shows that moduleId comes back as a number.  Any reason for this change?  or is it because a test or another internal use of this is wrong?

rmi22186: oh i see.  you're turning into a string earlier on.  i'm fine with that

src/cookieSyncManager.ts
```diff
@@ -0,0 +1,184 @@
...
+        mpidIsNotInCookies: boolean
+    ): void => {
+        if (!mpid || mpInstance._Store.webviewBridgeEnabled) {
+            return;
+        }
+
+        const { pixelConfigurations } = mpInstance._Store;
+        const persistence = mpInstance._Persistence.getPersistence();
+
+        pixelConfigurations.forEach(pixelSettings => {
```
rmi22186, line 50:
```suggestion
        pixelConfigurations.forEach((pixelSettings: IPixelConfig) => {
```

src/cookieSyncManager.ts
```diff
@@ -0,0 +1,184 @@
...
+import Constants from './constants';
+import { ICookieSyncManager } from './cookieSyncManager.interfaces';
+import { MParticleWebSDK } from './sdkRuntimeModels';
+import { Dictionary, MPID } from '@mparticle/web-sdk';
+import { IConsentRules } from './consent';
+
+const { Messages } = Constants;
+const { InformationMessages } = Messages;
+
+interface PixelConfig {
```
rmi22186, line 11:
```suggestion
interface IPixelConfig {
```

src/cookieSyncManager.ts
```diff
@@ -0,0 +1,184 @@
...
+                    return;
+                }
+
+                if (!persistence[mpid].csd) {
+                    persistence[mpid].csd = {};
+                }
+
+                const lastSyncDateForModule = persistence[mpid].csd[moduleId]
+                    ? persistence[mpid].csd[moduleId]
+                    : null;
```
rmi22186, line 112: not sure why this was every ternary if it's setting to `null` if `undefined`:
```suggestion
                const lastSyncDateForModule = persistence[mpid].csd[moduleId] || null;
```

src/cookieSyncManager.ts
```diff
@@ -0,0 +1,184 @@
...
+                        urlWithRedirect,
+                        moduleId,
+                        mpid,
+                        persistence[mpid].csd,
+                        filteringConsentRuleValues,
+                        mpidIsNotInCookies,
+                        requiresConsent
+                    );
+                }
+
```
rmi22186, line 126:
```suggestion
```

src/cookieSyncManager.ts
```diff
@@ -0,0 +1,184 @@
...
+export const DAYS_IN_MILLISECONDS = 1000 * 60 * 60 * 24;
+
+const hasFrequencyCapExpired = (
+    lastSyncDate: number,
+    frequencyCap: number
+): boolean => {
+    return (
+        new Date().getTime() >
+        new Date(lastSyncDate).getTime() + frequencyCap * DAYS_IN_MILLISECONDS
+    );
```
rmi22186, line 28: `lastSyncDate` should be optional, and the second parameter.  [this code block](https://github.com/mParticle/mparticle-web-sdk/pull/944/commits/b6a0be6bfebb5ffce2dad7cd853dbc2e89cf472d#diff-240c8f092446a9912eb4e9046db7089f14af869309570f62a42cd6e9f0e75c90L108-L143) will call `self.performCookieSync` if there is no `lastSyncDate` and

src/cookieSyncManager.ts
```diff
@@ -0,0 +1,184 @@
...
+        const redirect = redirectUrl ? replaceMPID(redirectUrl, mpid) : '';
+        return url + encodeURIComponent(redirect);
+    };
+
+    // Private
+    this.performCookieSync = (
+        url: string,
+        moduleId: string,
+        mpid: MPID,
+        cookieSyncDates: Dictionary<number>,
```
rmi22186, line 146: we use `cookieSyncDates: Dictionary<number>` in enough places that i think you should just make it its own Type

test/jest/cookieSyncManager.spec.ts
```diff
@@ -0,0 +1,590 @@
...
+    esId: 24053,
+    isDebug: true,
+    isProduction: true,
+    settings: {},
+    frequencyCap: 14,
+    pixelUrl: '',
+    redirectUrl: '',
+};
+
+describe.only('CookieSyncManager', () => {
```
rmi22186, line 18:
```suggestion
describe('CookieSyncManager', () => {
```

test/jest/cookieSyncManager.spec.ts
```diff
@@ -451,4 +493,43 @@ describe('CookieSyncManager', () => {
...
+
+            const cookieSyncManager = new CookieSyncManager(mockSDK);
+
+            const result = cookieSyncManager.combineUrlWithRedirect(
+                '1234',
+                'https://test.com',
+                '?redirect=https://redirect.com&mpid=%%mpid%%',
+            );
+
+            expect(result).toBe('https://test.com%3Fredirect%3Dhttps%3A%2F%2Fredirect.com%26mpid%3D1234');
```
rmi22186, line 569: something looks off to me here.  perhaps it's the example you are using for the redirect url that is causing this.  but the urlWithRedirect is what the browser will load, and in this case, I don't think the browser will use the %3F and know that query parameters are coming after it. maybe it's best to look at some [real examples](https://jssdkcdns.mparticle.com/JS/v2/beab3f4d34281d45bfcdbbd7eb21c083/config?env=0)  (search for `pixelConfig`)and see if you need to make any changes

rmi22186 (changes requested):

test/jest/cookieSyncManager.spec.ts
```diff
@@ -439,7 +439,7 @@ describe.only('CookieSyncManager', () => {
             expect(loggerSpy).toHaveBeenCalledWith('Performing cookie sync');
         });
 
-        it('should return early if the user has not consented to the cookie sync', () => {
+        it.only('should return early if the user has not consented to the cookie sync', () => {
```
rmi22186, line 443: remove .only

test/jest/cookieSyncManager.spec.ts
```diff
@@ -450,7 +450,61 @@ describe.only('CookieSyncManager', () => {
...
+            );
+
+            // Simulate image load
+            mockImage.onload();
+
+            expect(mockImage.src).toBe('');
+            expect(cookieSyncDates[42]).toBeUndefined();
+        });
+
+        it.only('should return early if requiresConsent and mpidIsNotInCookies are both true', () => {
```
rmi22186, line 497: remove .only

test/jest/cookieSyncManager.spec.ts
alexs-mparticle: Agreed. There's a lot of nested logic that could be broken out to make things more readable. I think it should likely be done as part of SQDSDKS-6915

test/jest/cookieSyncManager.spec.ts
alexs-mparticle: I think both tests are valid because they each test two different code paths in what is a somewhat squirrely and complex function. I'd rather change the assertion to make them more distinct.

test/jest/cookieSyncManager.spec.ts
alexs-mparticle: `%3F` is what will be returned when you run the `redirect` url through `encodeURIComponent`.

alexs-mparticle: Not sure if this is a bug or something we didn't consider, but the current logic does not really append the redirect url as a query param. It just concatenates `url` and `redirectUrl` into a single string. From the examples provided, it looks like it's up to the customer/user to add a `?` into the url before concatenation. Is this expected?

src/cookieSyncManager.ts
```diff
@@ -0,0 +1,167 @@
...
+const { Messages } = Constants;
+const { InformationMessages } = Messages;
+
+export const DAYS_IN_MILLISECONDS = 1000 * 60 * 60 * 24;
+
+const hasFrequencyCapExpired = (
+    frequencyCap: number,
+    lastSyncDate?: number,
+): boolean => {
+    return (
```
alexs-mparticle, line 17: Note to self: Return true if `lastSyncDate` is null.

src/persistence.interfaces.ts
```diff
@@ -79,10 +80,8 @@ export interface IPersistenceMinified extends Dictionary {
     // };
 }
 
-export type CookieSyncDate = Dictionary<number>;
-
 export interface IUserPersistenceMinified extends Dictionary {
-    csd: CookieSyncDate; // Cookie Sync Dates // list of timestamps for last cookie sync
+    csd: CookieSyncDates; // Cookie Sync Dates // list of timestamps for last cookie sync
```
rmi22186, line 84:
```suggestion
    csd: CookieSyncDates; // list of timestamps for last cookie syncs per module
```

rmi22186 (approved):

PR #945, opened 2024-11-05
refactor: Add Type interfaces for identity api client

alexs-mparticle:
## Instructions

 ## Summary
 - Create Type Interfaces for Identity API Client

 ## Testing Plan
 - N/A - Defines new type interfaces but they will not be used until we refactor api client

 - Closes https://go.mparticle.com/work/SQDSDKS-6678

### DIFF
--- a/src/identityApiClient.interfaces.ts
+++ b/src/identityApiClient.interfaces.ts
@@ -1,0 +1,46 @@
+import { IdentityApiData, MPID, UserIdentities } from '@mparticle/web-sdk';
+import {
+    IdentityCallback,
+    IIdentityResponse,
+} from './identity-user-interfaces';
+import {
+    IAliasRequest,
+    IAliasCallback,
+    IIdentityRequest,
+    IdentityAPIMethod,
+    IIdentity,
+} from './identity.interfaces';
+import { MParticleWebSDK } from './sdkRuntimeModels';
+
+export interface IIdentityApiClient {
+    sendAliasRequest: (
+        aliasRequest: IAliasRequest,
+        aliasCallback: IAliasCallback
+    ) => Promise<void>;
+    sendIdentityRequest: (
+        identityApiRequest: IIdentityRequest,
+        method: IdentityAPIMethod,
+        callback: IdentityCallback,
+        originalIdentityApiData: IdentityApiData,
+        parseIdentityResponse: IIdentity['parseIdentityResponse'],
+        mpid: MPID,
+        knownIdentities: UserIdentities
+    ) => Promise<void>;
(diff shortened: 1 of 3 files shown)


src/identityApiClient.interfaces.ts
```diff
@@ -0,0 +1,44 @@
...
+import {
+    IAliasRequest,
+    IAliasCallback,
+    IIdentityRequest,
+    IdentityAPIMethod,
+    IIdentity,
+} from './identity.interfaces';
+import { MParticleWebSDK } from './sdkRuntimeModels';
+
+export interface IIdentityApiClient {
```
rmi22186, line 15: I actually find interfaces that are not yet being used to be easier to debug when they _are_ used.  I think a better approach for this would be to PR this into an IdentityAPI refactor feature branch, then do https://mparticle-eng.atlassian.net/browse/SQDSDKS-6679 into the same one.  This way we are more certain these types are correct based on implementation/linting int he IDE, as opposed to thinking it _looks_ fine.  Let me know if that is clear.

src/identityApiClient.interfaces.ts
```diff
@@ -0,0 +1,44 @@
...
+    sendAliasRequest: (
+        aliasRequest: IAliasRequest,
+        aliasCallback: IAliasCallback
+    ) => Promise<void>;
+    sendIdentityRequest: (
+        identityApiRequest: IIdentityRequest,
+        method: IdentityAPIMethod,
+        callback: IdentityCallback,
+        originalIdentityApiData: IdentityApiData,
+        parseIdentityResponse: IIdentity['parseIdentityResponse'],
```
rmi22186, line 25: Should this instead be a `Pick` of some sort?  Maybe this works also, but a `Pick` might be more TS-y?  It's another reason to implement the interface to see if both do work.

rmi22186 (changes requested):
generally lgtm, but see my comment about PR-ing into a feature branch so that the unused interface is appropriately used and shown to be working

src/identityApiClient.interfaces.ts
alexs-mparticle: I normally agree with you but one of the biggest struggles with this migration is the discrepancy between what we _think_ the interfaces should be and what _actually_ is happening in the code.

In the past, when trying to do the migration in one shot, it proved difficult to keep the migration in scope because of this discrepancy. I think the better approach would be to define the intended interfaces, and then update those interfaces if we realized we missed something or plan out a future revision to keep things in scope.

src/identityApiClient.interfaces.ts
rmi22186: sgtm

rmi22186 (approved):

PR #947, opened 2024-11-08
fix: prevent error when forwarding user attributes to kits while kit …

mmustafa-tse:
…blocking enabled and unplanned UAs are allowed

 ## Summary
 - A customer provided that an error is thrown when enabling kit blocking or user attributes in the UI and having unplanned user attributes to be allowed, this result in a conflict where dataPointMatchLookUps for user_attributes will be null since and thus when we try to look into a key of null, the error is thrown.

 ## Testing Plan
 - [Y] Was this tested locally? If not, explain why.
 - Unit tests did not fail and E2E tested

 - Closes https://mparticle-eng.atlassian.net/browse/SQDSDKS-6391

### DIFF
--- a/src/kitBlocking.ts
+++ b/src/kitBlocking.ts
@@ -460,7 +460,7 @@ export default class KitBlocker {
         if (!this.blockUserAttributes) {
             return false
         }
-        var matchedAttributes = this.dataPlanMatchLookups['user_attributes'];
+        const matchedAttributes = this.dataPlanMatchLookups['user_attributes'];
 
         // When additionalProperties is set to true, matchedAttributes 
         // will be a boolean, otherwise it will return an object
@@ -468,7 +468,7 @@ export default class KitBlocker {
             return false
         }
 
-        if (this.blockUserAttributes && typeof matchedAttributes === "object") {
+        if (typeof matchedAttributes === "object") {
             if (matchedAttributes[key] === true) {
                 return false;
             } else {
--- a/src/kitBlocking.ts
+++ b/src/kitBlocking.ts
@@ -476,6 +476,8 @@ export default class KitBlocker {
             }
         }
 
+        // When block unplanned user attributess is enabled and allow unplanned user
+        // attributes is also enabled, allowing unplanned user attribues will be prioritized
         return false;
     }
 
--- a/src/kitBlocking.ts
+++ b/src/kitBlocking.ts
@@ -460,17 +460,23 @@ export default class KitBlocker {
         if (!this.blockUserAttributes) {
             return false
         }
-        if (this.blockUserAttributes) {
-            const matchedAttributes = this.dataPlanMatchLookups['user_attributes'];
-            if (matchedAttributes === true) {
-                return false
-            }
-            if (!matchedAttributes[key]) {
-                return true
+        var matchedAttributes = this.dataPlanMatchLookups['user_attributes'];
+
+        // When additionalProperties is set to true, matchedAttributes 
+        // will be a boolean, otherwise it will return an object
+        if (typeof matchedAttributes === 'boolean' && matchedAttributes) {
+            return false
+        }
+
+        if (this.blockUserAttributes && typeof matchedAttributes === "object") {
+            if (matchedAttributes[key] === true) {
+                return false;
+            } else {
+                return true;
             }
         }
 
-        return false
+        return false;
     }
 
     isIdentityBlocked(key: string) {
--- a/src/kitBlocking.ts
+++ b/src/kitBlocking.ts
@@ -476,8 +476,8 @@ export default class KitBlocker {
             }
         }
 
-        // When block unplanned user attributess is enabled and allow unplanned user
-        // attributes is also enabled, allowing unplanned user attribues will be prioritized
+        // When "Block unplanned user attributes" is enabled and "Allow unplanned user
+        // attributes" is also enabled in the UI, allowing unplanned user attributes will be prioritized
         return false;
     }
 
--- a/test/src/tests-kit-blocking.ts
+++ b/test/src/tests-kit-blocking.ts
@@ -628,7 +628,7 @@ describe('kit blocking', () => {
             let kitBlocker = new KitBlocker(kitBlockerDataPlan, window.mParticle.getInstance());
 
             expect(() => { kitBlocker.isAttributeKeyBlocked('unplannedAttr') }).to.not.throw(TypeError, /Cannot read properties of undefined \(reading 'unplannedAttr'\)/)
-            // allow unplanned user attributes is prioritized when blocking unplanned attributes is also enabled, hence the expected value is false
+            // "Allow unplanned user attributes" is prioritized when blocking unplanned attributes is also enabled, hence the expected value is false
             expect(kitBlocker.isAttributeKeyBlocked('unplannedAttr')).to.equal(false)
             // reset data points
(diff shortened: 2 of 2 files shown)


src/kitBlocking.ts
```diff
@@ -460,17 +460,23 @@ export default class KitBlocker {
...
             return false
         }
-        if (this.blockUserAttributes) {
-            const matchedAttributes = this.dataPlanMatchLookups['user_attributes'];
-            if (matchedAttributes === true) {
-                return false
-            }
-            if (!matchedAttributes[key]) {
-                return true
+        var matchedAttributes = this.dataPlanMatchLookups['user_attributes'];
```
alexs-mparticle, line 463:
```suggestion
        const matchedAttributes = this.dataPlanMatchLookups['user_attributes'];
```

src/kitBlocking.ts
```diff
@@ -460,17 +460,23 @@ export default class KitBlocker {
...
+
+        // When additionalProperties is set to true, matchedAttributes 
+        // will be a boolean, otherwise it will return an object
+        if (typeof matchedAttributes === 'boolean' && matchedAttributes) {
+            return false
+        }
+
+        if (this.blockUserAttributes && typeof matchedAttributes === "object") {
```
alexs-mparticle, line 471: Is checking `this.blockUserAttributes` necessary? We're already checking for it's existence a few lines earlier.

alexs-mparticle (changes requested):

rmi22186 (commented):
There should be a test added to this that fails with the previous code, but passes with the current code.

src/kitBlocking.ts
```diff
@@ -460,17 +460,23 @@ export default class KitBlocker {
...
-            if (matchedAttributes === true) {
-                return false
-            }
-            if (!matchedAttributes[key]) {
-                return true
+        const matchedAttributes = this.dataPlanMatchLookups['user_attributes'];
+
+        // When additionalProperties is set to true, matchedAttributes 
+        // will be a boolean, otherwise it will return an object
+        if (typeof matchedAttributes === 'boolean' && matchedAttributes) {
```
rmi22186, line 467: it's also unclear to me what combination of settings will have `additionalAttributes === true`, and so `matchedAttributes` is a `boolean`. can you share a data plan that results in this?  i've only ever seen `additionalAttributes === false` when `Block unplanned user attributes` is enabled, and `Allow unplanned attributes` is selected

rmi22186: nevermind!  seems to be a caching issue.  i ended up downloading the plans instead to avoid having to wait for cache to bust.

rmi22186:
@mmustafa-tse i reviewed this again and the logic looks good.  please add a unit test.  give `test/src/tests-kit-blocking.ts` a look and if you need help, i'm happy to hop on with you.  this is a pretty low priority bug though, so please address anything else you are working on first.  thanks!

mmustafa-tse:
@rmi22186 @alexs-mparticle - added unit test and confirmed it was passing with my code change in KitBlocking.ts and was failing when testing the old code prior to my changes

test/src/tests-kit-blocking.ts
```diff
@@ -616,6 +617,24 @@ describe('kit blocking', () => {
...
+            window.mParticle.config.kitConfigs.push(forwarderDefaultConfiguration('MockForwarder'));
+            window.mParticle.init(apiKey, window.mParticle.config);
+            
+            //save old data points for reset later
+            const oldDataPoints = dataPlan.dtpn.vers.version_document.data_points;
+            dataPlan.dtpn.vers.version_document.data_points = [];
+            let kitBlocker = new KitBlocker(kitBlockerDataPlan, window.mParticle.getInstance());
+
+            expect(() => { kitBlocker.isAttributeKeyBlocked('unplannedAttr') }).to.not.throw(TypeError, /Cannot read properties of undefined \(reading 'unplannedAttr'\)/)
+
```
rmi22186, line 630: Add another expect for kitBlocker.isAttributeKeyBlocked('unplannedAttr' to equal true)

test/src/tests-kit-blocking.ts
```diff
@@ -616,6 +617,24 @@ describe('kit blocking', () => {
             })
         });
 
+        it('integration test - should not throw an error when unplanned user attributes are allowed and block.ua = true', function(done) {
+            window.mParticle.config.kitConfigs.push(forwarderDefaultConfiguration('MockForwarder'));
+            window.mParticle.init(apiKey, window.mParticle.config);
+            
+            //save old data points for reset later
+            const oldDataPoints = dataPlan.dtpn.vers.version_document.data_points;
+            dataPlan.dtpn.vers.version_document.data_points = [];
```
rmi22186, line 626: Add a comment specifying that when "Allow planned user attributes" is set, that data_points returns an empty array.

rmi22186 (changes requested):
minor comments, otherwise lgtm!

test/src/tests-kit-blocking.ts
mmustafa-tse: added the case and confirmed this should be expected to be false

src/kitBlocking.ts
```diff
@@ -460,17 +460,25 @@ export default class KitBlocker {
...
+            if (matchedAttributes[key] === true) {
+                return false;
+            } else {
+                return true;
             }
         }
 
-        return false
+        // When block unplanned user attributess is enabled and allow unplanned user
+        // attributes is also enabled, allowing unplanned user attribues will be prioritized
```
rmi22186, line 480:
```suggestion
        // When "Block unplanned user attributes" is enabled and "Allow unplanned user
        // attributes" is also enabled in the UI, allowing unplanned user attributes will be prioritized
```

test/src/tests-kit-blocking.ts
```diff
@@ -616,6 +617,26 @@ describe('kit blocking', () => {
             })
         });
 
+        it('integration test - should not throw an error when unplanned user attributes are allowed and block.ua = true', function(done) {
+            window.mParticle.config.kitConfigs.push(forwarderDefaultConfiguration('MockForwarder'));
+            window.mParticle.init(apiKey, window.mParticle.config);
+            
+            // save old data points for reset later
+            const oldDataPoints = dataPlan.dtpn.vers.version_document.data_points;
+            // when allow unplanned user attributes is enabled, the data points returned is an empty array
```
rmi22186, line 626:
```suggestion
            // when "Allow unplanned user attributes" is enabled, the data points returned is an empty array
```

test/src/tests-kit-blocking.ts
```diff
@@ -616,6 +617,26 @@ describe('kit blocking', () => {
...
+            dataPlan.dtpn.vers.version_document.data_points = [];
+            let kitBlocker = new KitBlocker(kitBlockerDataPlan, window.mParticle.getInstance());
+
+            expect(() => { kitBlocker.isAttributeKeyBlocked('unplannedAttr') }).to.not.throw(TypeError, /Cannot read properties of undefined \(reading 'unplannedAttr'\)/)
+            // allow unplanned user attributes is prioritized when blocking unplanned attributes is also enabled, hence the expected value is false
```
rmi22186, line 631:
```suggestion
            // "Allow unplanned user attributes" is prioritized when blocking unplanned attributes is also enabled, hence the expected value is false
```

rmi22186 (approved):

alexs-mparticle (approved):

PR #948, opened 2024-11-08
Refactor/sqdsdks 6679 migrate identity api client typescript

alexs-mparticle:
## Instructions

 ## Summary

 ## Testing Plan


src/identityApiClient.ts
```diff
@@ -0,0 +1,257 @@
...
+
+        const uploadUrl =
+            method === Modify
+                ? uploadServiceUrl + mpid + '/' + method
+                : uploadServiceUrl + method;
+
+        return uploadUrl;
+    };
+
+    this.getIdentityResponseFromFetch = (response, responseBody) => ({
```
rmi22186, line 240:
```suggestion
    this.getIdentityResponseFromFetch = (response, responseBody): IIdentityResponse => ({
```

src/identityApiClient.ts
```diff
@@ -0,0 +1,257 @@
...
+    getIdentityResponseFromXHR: (response: XMLHttpRequest) => IIdentityResponse;
+}
+
+
+export interface IAliasResponseBody {
+    message?: string
+}
+
+export default function IdentityAPIClient(this: IIdentityApiClient, mpInstance) {
+    this.sendAliasRequest = async function(aliasRequest, aliasCallback) {
```
rmi22186, line 45: what's the best practice for when you define an interface like `IIdentityApiClient`...should everything in the signatures also be typed?  i expect then we don't need to add types, in which case we should remove them from the signature of `this.sendIdentityRequest`

whatever we pick, let's be consistent

src/identityApiClient.interfaces.ts
```diff
@@ -1,44 +0,0 @@
-import { IdentityApiData, MPID, UserIdentities } from '@mparticle/web-sdk';
```
rmi22186, line 1: for some reason i thought we were goin to keep the `.interfaces.ts` files to keep each module smaller and to ensure separation of interfaces from code.  if they were temporary because we are migrating, i'm fine with that.  i just want to make sure we are picking a path and sticking with it

src/identityApiClient.ts
```diff
@@ -0,0 +1,257 @@
...
+        }
+
+        const previousMPID = mpid || null;
+        const uploadUrl = this.getUploadUrl(method, mpid);
+
+        const uploader = window.fetch
+            ? new FetchUploader(uploadUrl)
+            : new XHRUploader(uploadUrl);
+
+        // TODO: Write test to verify fetch payload
```
rmi22186, line 157: relevant?  or already done?

src/identityApiClient.ts
```diff
@@ -0,0 +1,257 @@
...
+            headers: {
+                Accept: 'text/plain;charset=UTF-8',
+                'Content-Type': 'application/json',
+                'x-mp-key': mpInstance._Store.devToken,
+            },
+            body: JSON.stringify(identityApiRequest),
+        };
+
+        console.log('fetchPayload', fetchPayload)
+
```
rmi22186, line 171:
```suggestion
```

src/identityApiClient.ts
```diff
@@ -0,0 +1,257 @@
...
+        try {
+            console.log('uploader', uploader);
+
```
rmi22186, line 174:
```suggestion
```

src/identityApiClient.ts
```diff
@@ -0,0 +1,257 @@
...
+            mpInstance._Store.identityCallInFlight = true;
+            const response = await uploader.upload(fetchPayload);
+
+            console.log('response', response);
+
```
rmi22186, line 179:
```suggestion
```

src/identityApiClient.ts
```diff
@@ -0,0 +1,257 @@
...
+            let identityResponse: IIdentityResponse;
+
+            if (response.json) {
+
+                console.log('fetch response')
+
```
rmi22186, line 185:
```suggestion
```

src/identityApiClient.ts
```diff
@@ -0,0 +1,257 @@
...
+            if (response.json) {
+
+                console.log('fetch response')
+                
+                // https://go.mparticle.com/work/SQDSDKS-6568
+                // FetchUploader returns the response as a JSON object that we have to await
+                const responseBody = await response.json();
+
+                console.log('responseBody', responseBody);
+
```
rmi22186, line 191:
```suggestion
```

src/identityApiClient.ts
```diff
@@ -0,0 +1,257 @@
...
+                identityResponse = this.getIdentityResponseFromFetch(
+                    response,
+                    responseBody
+                );
+            } else {
+                console.log('xhr response', response)
+
```
rmi22186, line 198:
```suggestion
```

src/identityApiClient.ts
```diff
@@ -0,0 +1,257 @@
...
+                previousMPID,
+                callback,
+                originalIdentityApiData,
+                method,
+                knownIdentities,
+                false
+            );
+        } catch (err) {
+            console.log('did i error?');
+            console.log('callback', callback);
```
rmi22186, line 220:
```suggestion
```

src/identityApiClient.ts
```diff
@@ -0,0 +1,257 @@
...
+        const { verbose, error } = mpInstance.Logger;
+        const { invokeAliasCallback } = mpInstance._Helpers;
+        const { aliasUrl } = mpInstance._Store.SDKConfig;
+        const { devToken: apiKey } = mpInstance._Store;
+
+        verbose(Messages.InformationMessages.SendAliasHttp);
+
+        // https://go.mparticle.com/work/SQDSDKS-6750
+        const uploadUrl = `https://${aliasUrl}${apiKey}/Alias`;
+        const uploader = window.fetch
```
rmi22186, line 55: add types

src/identityApiClient.ts
```diff
@@ -0,0 +1,257 @@
...
+        ? new FetchUploader(uploadUrl)
+        : new XHRUploader(uploadUrl);
+        
+        
+        // https://go.mparticle.com/work/SQDSDKS-6568
+        const uploadPayload = {
```
rmi22186, line 61: add type

src/identityApiClient.ts
```diff
@@ -0,0 +1,257 @@
...
+            method: 'post',
+            headers: {
+                Accept: 'text/plain;charset=UTF-8',
+                'Content-Type': 'application/json',
+            },
+            body: JSON.stringify(aliasRequest),
+        };
+
+        try {
+            const response = await uploader.upload(uploadPayload);
```
rmi22186, line 71: add type

src/identityApiClient.ts
```diff
@@ -0,0 +1,257 @@
...
+                HTTPCodes.activeIdentityRequest,
+                'There is currently an Identity request processing. Please wait for this to return before requesting again'
+            );
+            return;
+        }
+
+        const previousMPID = mpid || null;
+        const uploadUrl = this.getUploadUrl(method, mpid);
+
+        const uploader = window.fetch
```
rmi22186, line 153: Add types

src/identityApiClient.ts
```diff
@@ -0,0 +1,257 @@
...
+            ? new FetchUploader(uploadUrl)
+            : new XHRUploader(uploadUrl);
+
+        // TODO: Write test to verify fetch payload
+
+        // https://go.mparticle.com/work/SQDSDKS-6568
+        const fetchPayload = {
```
rmi22186, line 160: Add type

src/identityApiClient.ts
```diff
@@ -0,0 +1,257 @@
...
+            body: JSON.stringify(identityApiRequest),
+        };
+
+        console.log('fetchPayload', fetchPayload)
+
+        try {
+            console.log('uploader', uploader);
+
+            mpInstance._Store.identityCallInFlight = true;
+            const response = await uploader.upload(fetchPayload);
```
rmi22186, line 176: Add type

src/identityApiClient.ts
```diff
@@ -0,0 +1,257 @@
...
+
+            let identityResponse: IIdentityResponse;
+
+            if (response.json) {
+
+                console.log('fetch response')
+                
+                // https://go.mparticle.com/work/SQDSDKS-6568
+                // FetchUploader returns the response as a JSON object that we have to await
+                const responseBody = await response.json();
```
rmi22186, line 188: Add type

src/identityApiClient.ts
```diff
@@ -0,0 +1,257 @@
...
+            console.log('did i error?');
+            console.log('callback', callback);
+            mpInstance._Store.identityCallInFlight = false;
+            invokeCallback(callback, HTTPCodes.noHttpCoverage, err);
+            error('Error sending identity request to servers' + ' - ' + err);
+        }
+    };
+
+    this.getUploadUrl = (method, mpid) => {
+        const uploadServiceUrl = mpInstance._Helpers.createServiceUrl(
```
rmi22186, line 228: Add types=

src/identityApiClient.ts
```diff
@@ -0,0 +1,257 @@
...
+            mpInstance._Store.SDKConfig.identityUrl
+        );
+
+        const uploadUrl =
```
rmi22186, line 232: Add type

test/src/tests-identityApiClient.ts
```diff
@@ -0,0 +1,480 @@
...
+        beforeEach(function() {
+            fetchMock.post(urls.events, 200);
+            mockServer = sinon.createFakeServer();
+            mockServer.respondImmediately = true;
+
+            mockServer.respondWith(urls.identify, [
+                200,
+                {},
+                JSON.stringify({ mpid: testMPID, is_logged_in: false }),
+            ]);
```
rmi22186, line 373: mock server is for xhr.  and seems like xhr is not being tested here, only fetchMock.  we should either add the xhr legacy tests, or remove.  i'd vote for the former just because there aren't many tests here.

rmi22186 (changes requested):
looks great overall.  the number of comments is high, but it's mostly the same thing over and over...if we are migrating to ts...let's keep everything typed!

src/identityApiClient.ts
alexs-mparticle: This is a good callout.

While I think we can avoid repeating ourselves because the IDE can infer the type based on the definition and signature, I can see a benefit to repeating the definition in that it puts the definition closer to the actual implementation.

test/src/tests-identityApiClient.ts
alexs-mparticle: This collection of tests was copied directly from the existing alias request tests. I guess we missed it when we refactored.

src/identityApiClient.interfaces.ts
alexs-mparticle: I think my intention was to always have the `.interfaces.ts` as temporary until we migrate completely to TypeScript.

Other than intentionally mirroring native (which splits implementation and interfaces), I think the only reason to keep them separate is if the module is too large or complex and having a separate file makes it easier to compare the interface and implementation. However, in that case, I would opt to refactor or break up the module.

src/uploaders.ts
```diff
@@ -1,6 +1,6 @@
 type HTTPMethod = 'get' | 'post';
 
-export interface fetchPayload {
+export interface IFetchPayload {
```
rmi22186, line 3: good catch

src/identityApiClient.ts
```diff
@@ -36,13 +53,26 @@ export interface IIdentityApiClient {
     getIdentityResponseFromXHR: (response: XMLHttpRequest) => IIdentityResponse;
 }
 
-
 export interface IAliasResponseBody {
-    message?: string
+    message?: string;
+}
+
+interface IdentityApiRequestPayload extends IFetchPayload {
```
rmi22186, line 60: i like this

src/identityApiClient.ts
```diff
@@ -140,7 +173,7 @@ export default function IdentityAPIClient(this: IIdentityApiClient, mpInstance)
 
         if (mpInstance._Store.identityCallInFlight) {
             invokeCallback(
-                callback,
+                callback as Callback,
```
rmi22186, line 176: type here is `Callback` but in the function signature it's `IdentityCallback`...I'm pretty sure it should be the latter

src/identityApiClient.ts
```diff
@@ -216,35 +234,42 @@ export default function IdentityAPIClient(this: IIdentityApiClient, mpInstance)
...
             );
         } catch (err) {
-            console.log('did i error?');
-            console.log('callback', callback);
             mpInstance._Store.identityCallInFlight = false;
-            invokeCallback(callback, HTTPCodes.noHttpCoverage, err);
+            invokeCallback(
+                callback as Callback,
+                HTTPCodes.noHttpCoverage,
+                err.toString()
```
rmi22186, line 241: I think you can mimic [line 144](https://github.com/mParticle/mparticle-web-sdk/pull/948/files/8a2e225a3e69d97b437f21d03c9a156c8561fb32..e8be568fde477988b36ddc4ceb8ddd2a67132d23#diff-3e282016703704717fe073d29ecd32c1bd2505325205cfca20d18fd4c8003030R144-R151) and cast the error as type Error in order to set `err.message` here instead of `err.toString()`

rmi22186 (changes requested):
minor nits

src/identityApiClient.interfaces.ts
alexs-mparticle: Chatted on Zoom and decided that we should remove the `.interface.ts` file after migrating a module, unless the module is longer than 500 lines of code.

src/identityApiClient.ts
```diff
@@ -0,0 +1,284 @@
...
+                identityResponse,
+                previousMPID,
+                callback,
+                originalIdentityApiData,
+                method,
+                knownIdentities,
+                false
+            );
+        } catch (err) {
+            const errorMessage = (err as Error).message || 'Unknown Error';
```
rmi22186, line 237: I think `Unknown error` is too generic. While we will know it is only coming from here if we do a global search for it, it might be more helpful to provide more context.  

I actually think here is a good instance to do something like a `toString` which you had before if there is no message. That will allow us to debug further.  What do you think?

rmi22186 (changes requested):

src/identityApiClient.ts
alexs-mparticle: I think I would prefer `err.toString()` if `error.message` is not available.

rmi22186 (approved):
looks great!

PR #950, opened 2024-11-21
refactor: Migrate Types Module to TypeScript

alexs-mparticle:
## Instructions

 ## Summary

 ## Testing Plan


src/identity.js
```diff
@@ -1623,7 +1622,7 @@ export default function Identity(mpInstance) {
                     self.setForwarderCallbacks(newUser, method);
                 }
 
-                const newIdentitiesByName = getNewIdentitiesByName(
+                const newIdentitiesByName = IdentityType.getNewIdentitiesByName(
```
jared-dickman, line 1625: **praise** name-spacing

src/types.ts
```diff
@@ -0,0 +1,399 @@
...
+    Commerce: 16 as const,
+    Media: 20 as const,
+    UserAttributeChange: 17 as const,
+    UserIdentityChange: 18 as const,
+};
+
+// Dictionary that contains MessageTypes that will
+// trigger an immediate upload.
+export const TriggerUploadType = {
+    [MessageType.Commerce]: 1 as const,
```
jared-dickman, line 26: if these are types, should they have the same value `1`?

alexs-mparticle: This is something that probably needs to be refactored.  Were [using the values as a sort of lookup table](https://github.com/mParticle/mparticle-web-sdk/blob/master/src/batchUploader.ts#L186) in the batchUploader when it might be cleaner to do this as a function instead. But that is likely out of scope for this PR.

rmi22186 (changes requested):
Looks great so far.  Putting this comments in in time for our sync.  Still need to finish going through types.ts and utils.ts, and the tests.

src/kitFilterHelper.ts
```diff
@@ -1,18 +1,17 @@
-import { generateHash } from "./utils";
+import { generateHash, valueof } from "./utils";
 // TODO: https://mparticle-eng.atlassian.net/browse/SQDSDKS-5381
-import { EventTypeEnum, IdentityType } from "./types.interfaces";
-import Constants from './constants';
+import { EventType, IdentityType } from "./types";
 
 export default class KitFilterHelper {
-    static hashEventType(eventType: EventTypeEnum): number {
+    static hashEventType(eventType: valueof<typeof EventType>): number {
```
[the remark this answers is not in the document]
alexs-mparticle: This sets the parameter to explicitly be a value of the type of EventType, meaning it can only be a member of EventType.

src/apiClient.ts
```diff
@@ -135,7 +135,9 @@ export default function APIClient(
             this.queueEventForBatchUpload(event);
         }
 
-        if (event.EventName !== Types.MessageType.AppStateTransition) {
+        // While Event Name is 'usually' a string, there are some cases where it is a number
+        // in that it could be a type of MessageType Enum
+        if (event.EventName as unknown as number !== Types.MessageType.AppStateTransition) {
```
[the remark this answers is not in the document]
alexs-mparticle: Created SQDSDKS-6935 to track.

src/batchUploader.ts
```diff
@@ -167,29 +167,35 @@ export class BatchUploader {
...
     }
 
+    // https://go.mparticle.com/work/SQDSDKS-3720
+    private shouldTriggerImmediateUpload (eventDataType: number): boolean {
+        const priorityEvents = [
+            MessageType.Commerce,
+            MessageType.UserIdentityChange,
+        ] as const;
+
+        return !this.batchingEnabled || priorityEvents.includes(eventDataType as typeof priorityEvents[number]);
```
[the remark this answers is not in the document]
alexs-mparticle: Sort of. `number` is actually being used as a `type` and not as a variable. This is saying that `eventDataType` should be treated as a type of `priorityEvent` which is an array of numbers.

This will transpile to `priorityEvents.includes(eventDataType)` but allows TypeScript to enforce that `eventDataType` is treated as a parameter that is compatible with the members of `priorityEvents`.

src/types.ts
```diff
@@ -0,0 +1,390 @@
...
+    OptOut: 6 as const,
+    AppStateTransition: 10 as const,
+    Profile: 14 as const,
+    Commerce: 16 as const,
+    Media: 20 as const,
+    UserAttributeChange: 17 as const,
+    UserIdentityChange: 18 as const,
+};
+
+export const EventType = {
```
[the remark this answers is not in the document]
alexs-mparticle: From my conversations with @einsteinx2, it sounds like we may be making some architectural changes to our event translation logic. I think rather than making a ticket for sustained engineering, we should only address this for when we decide on what is actually necessary for v3+.

src/types.ts
```diff
@@ -257,14 +259,25 @@ export const IdentityType = {
...
+    ): IdentitiesByType => {
         const newIdentitiesByName: IdentitiesByType = {};
 
         for (const key in newIdentitiesByType) {
-            const identityNameKey = IdentityType.getIdentityName(
-                parseNumber(key)
-            );
-            newIdentitiesByName[identityNameKey] = newIdentitiesByType[key];
+            // IdentityTypes are stored as numbers but are passed in as strings
+            if (IdentityType.getValuesAsStrings().includes(key)) {
```
rmi22186, line 275: I see what you are doing here.  I don' tknow that this needs to be done _every_ time parseIdentity is called. 
 Is there a way we can prevent that from happening...maybe once upon initiation of the identity module and putting that there as a property?  Otherwise I see a map with an includes and a for loop outside of that.  Seems a _bit_ much?  Although these are limited size arrays/objects and I don't imagine identity calls being made that often.

Just some food for thought.

src/types.ts
```diff
@@ -0,0 +1,404 @@
...
+                return 'click';
+            default:
+                return 'unknown';
+        }
+    },
+};
+
+export const ProfileMessageType = {
+    Logout: 3 as const,
+};
```
rmi22186, line 388: small nit.  similar to what you did with trigger uploads - since this is only used in the logout call in `identity.js`, might be worth moving it over there.  or since it's a .js file still, just remembering (or adding a ticket) to move this over there.

src/types.ts
```diff
@@ -0,0 +1,404 @@
...
+
+export const ApplicationTransitionType = {
+    AppInit: 1 as const,
+};
```
rmi22186, line 392: this could be moved to serverModel.ts since it's only called there.  and that's already a .ts file unlike the identity.js comment above

test/jest/types.spec.ts
```diff
@@ -0,0 +1,593 @@
...
+        expect(PromotionView).toEqual(18);
+        expect(PromotionClick).toEqual(19);
+        expect(ProductAddToWishlist).toEqual(20);
+        expect(ProductRemoveFromWishlist).toEqual(21);
+        expect(ProductImpression).toEqual(22);
+    });
+
+    describe('#getName', () => {
+
+    });
```
rmi22186, line 186: i assume this can be removed

test/jest/types.spec.ts
```diff
@@ -0,0 +1,593 @@
...
+        expect(Other).toEqual(0);
+        expect(CustomerId).toEqual(1);
+        expect(Facebook).toEqual(2);
+        expect(Twitter).toEqual(3);
+        expect(Google).toEqual(4);
+        expect(Microsoft).toEqual(5);
+        expect(Yahoo).toEqual(6);
+        expect(Email).toEqual(7);
+
+        // There is no value for 8
```
rmi22186, line 224: Context - 8 used to be `Alias` and it was a concept prior to Identity being implemented. After Identity was implemented, aliasing users was a new concept which supplanted the identity type.

test/jest/types.spec.ts
```diff
@@ -0,0 +1,593 @@
...
+
+        it('returns false if the identity type is not valid', () => {
+            const { isValid } = IdentityType;
+
+            expect(isValid(NaN)).toBe(false);
+            expect(isValid('invalid' as unknown as number)).toBe(false);
+        });
+    });
+
+    describe('#getNewIdentitesByName', () => {
```
rmi22186, line 306:
```suggestion
    describe('#getNewIdentitiesByName', () => {
```

test/jest/types.spec.ts
```diff
@@ -0,0 +1,593 @@
...
+                [Email]: 'bar@gmail.com',
+            };
+
+            expect(getNewIdentitiesByName(newIdentitiesByType)).toEqual({
+                customerid: 'foo',
+                email: 'bar@gmail.com',
+            });
+        });
+
+        it('returns an empty object if the identity type is not found', () => {
```
rmi22186, line 321: I think this test name is a bit misleading.  If an object is full of identity types that aren't real, then an empty object is returned.  Really this test should perhaps have 1 valid type, and the test should be "removes invalid identity types..."

test/jest/types.spec.ts
```diff
@@ -0,0 +1,593 @@
...
+            expect(getName(Other7)).toBe('Other ID');
+            expect(getName(Other8)).toBe('Other ID');
+            expect(getName(Other9)).toBe('Other ID');
+            expect(getName(Other10)).toBe('Other ID');
+            expect(getName(MobileNumber)).toBe('Other ID');
+            expect(getName(PhoneNumber2)).toBe('Other ID');
+            expect(getName(PhoneNumber3)).toBe('Other ID');
+
+            expect(getName(NaN)).toBe('Other ID');
+        });
```
rmi22186, line 401: I think you uncovered a bug here.  all these should probably have an actual Name that's not Other Id.  It's likely that the `getName` switch statement never was updated after these new IdentityTypes were added.

rmi22186 (changes requested):

src/types.ts
alexs-mparticle: Same as above. I think this should stay in here since it's a data type.

test/jest/types.spec.ts
alexs-mparticle: Created https://go.mparticle.com/work/SQDSDKS-6942 to track

rmi22186 (approved):

PR #955, opened 2024-12-04
feat: Add support for Google Click Ids

alexs-mparticle:
## Instructions

 ## Summary
 - Add support for Google Click Ids

 ## Testing Plan
 - Run through automated tests. Manually configure sample app to use Integration Capture setting in backend and add relevant `gclid`, `gbraid`, `wbraid` values to query parameters.

 - Closes https://go.mparticle.com/work/SQDSDKS-6947

### DIFF
--- a/test/jest/integration-capture.spec.ts
+++ b/test/jest/integration-capture.spec.ts
@@ -46,7 +46,7 @@ describe('Integration Capture', () => {
         it('should pass all clickIds to clickIds object', () => {
             jest.spyOn(Date, 'now').mockImplementation(() => 42);
 
-            const url = new URL('https://www.example.com/?fbclid=12345&');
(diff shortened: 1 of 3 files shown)


test/jest/integration-capture.spec.ts
```diff
@@ -62,9 +62,33 @@ describe('Integration Capture', () => {
...
                 _fbp: '54321',
+                gclid: '54321',
+                gbraid: '67890',
+                wbraid: '09876',
             }); 
         });
 
+        describe('Google Click Ids', () => {
+            it('should capture Google specific click ids', () => {
+                jest.spyOn(Date, 'now').mockImplementation(() => 42);
```
rmi22186, line 73: time shouldn't be needed since there is no processor

rmi22186 (approved):

PR #957, opened 2024-12-05
feat: Add support for Google Click Ids

alexs-mparticle:
## Instructions

 ## Summary
 - Add support for Google Click Ids

 ## Testing Plan
 - Run through automated tests. Manually configure sample app to use Integration Capture setting in backend and add relevant gclid, gbraid, wbraid values to query parameters.

 - Closes https://go.mparticle.com/work/SQDSDKS-6947

### DIFF
--- a/src/integrationCapture.ts
+++ b/src/integrationCapture.ts
@@ -56,6 +56,7 @@ interface IntegrationIdMapping {
 }
 
 const integrationMapping: IntegrationIdMapping = {
+    // Facebook / Meta
     fbclid: {
         mappedKey: 'Facebook.ClickId',
         processor: facebookClickIdProcessor,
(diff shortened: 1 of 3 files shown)


rmi22186 (approved):
LGTM!  Making it modular from the beginning really simplifies adding new url parameter click ids.

samdozor (approved):
looks good.

one higher level concern: we're parsing essentially random data within `queryStringParser` - who knows what crazy URLs there are out there. I see that we rolled our own parser for when `URLSearchParams` isn't there (seems like maybe 1-3% of users?). Have you considered having a try/catch to make sure we don't crash if given a totally garbage url?

I looked up the tests for node's internal url.js implementation and there is some wacky stuff it handles, which we do not.

alexs-mparticle:
@samdozor This is a great callout. I went ahead and created [SQDSDKS-6989](https://go.mparticle.com/work/SQDSDKS-6989) so we can work on this.

PR #958, opened 2024-12-06
ci: Add issue autorespond and close workflow

rmi22186:
## Instructions

 ## Summary
 - This uses a reusable workflow to automatically comment and close issues in the web sdk.

 ## Testing Plan
Tested in a personal github repository and it works there/

 - Closes https://go.mparticle.com/work/ci/SQDSDKS-6965

### DIFF
--- a/.github/workflows/issue-autorespond-and-close.yml
+++ b/.github/workflows/issue-autorespond-and-close.yml
@@ -6,8 +6,8 @@ on:
       - opened
 
 jobs:
-  invoke_reusable_workflow:
-    uses: mparticle/mparticle-workflows/.github/workflows/auto-respond-close-issue.yml@master
+  auto-respond-and-close-issue:
+    uses: mparticle/mparticle-workflows/.github/workflows/auto-respond-close-issue.yml@main
     with:
       issue_number: ${{ github.event.issue.number }}
       repository: ${{ github.repository }}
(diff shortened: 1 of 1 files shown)


.github/workflows/issue-autorespond-and-close.yml
```diff
@@ -0,0 +1,14 @@
+name: Auto respond and close issue
+
+on:
+  issues:
+    types:
+      - opened
+
+jobs:
+  invoke_reusable_workflow:
+    uses: mparticle/mparticle-workflows/.github/workflows/auto-respond-close-issue.yml@master
```
alexs-mparticle, line 10: This should be either `stable` or `main`.

```suggestion
    uses: mparticle/mparticle-workflows/.github/workflows/auto-respond-close-issue.yml@main
```

.github/workflows/issue-autorespond-and-close.yml
```diff
@@ -0,0 +1,14 @@
+name: Auto respond and close issue
+
+on:
+  issues:
+    types:
+      - opened
+
+jobs:
+  invoke_reusable_workflow:
```
alexs-mparticle, line 9: Nitpick: This job name is too generic. I know it's the only one in the file (for now) but I think we should give it something more appropriate in case we ever add new jobs in the future.

```suggestion
  auto-respond-and-close-issue:
```

alexs-mparticle (changes requested):

.github/workflows/issue-autorespond-and-close.yml
rmi22186: good point.

alexs-mparticle (approved):

PR #959, opened 2024-12-06
refactor: Update Error handling for Identity API Client

alexs-mparticle:
## Instructions

 ## Summary
 - Updates error handling and reporting for Identity API Client
 - Adds more detailed messaging and better reflects actual Identity API Error Responses.

 ## Testing Plan
 - Run SDK in sample app and disable networking. Manually modify identity requests with bad data and verify that errors are properly reported in developer console.

 - Closes https://go.mparticle.com/work/SQDSDKS-6932

### DIFF
--- a/src/identityApiClient.ts
+++ b/src/identityApiClient.ts
@@ -134,7 +134,11 @@ export default function IdentityAPIClient(
                         // response.json will always exist on a fetch, but can only be
                         // await-ed when the response is not empty, otherwise it will
                         // throw an error.
-                        aliasResponseBody = await response.json();
+                        try {
+                            aliasResponseBody = await response.json();
+                        } catch (e) {
+                            verbose('No response body for Alias request');
+                        }
                     } else {
                         // https://go.mparticle.com/work/SQDSDKS-6568
                         // XHRUploader returns the response as a string that we need to parse
--- a/src/identityApiClient.ts
+++ b/src/identityApiClient.ts
@@ -126,7 +126,7 @@ export default function IdentityAPIClient(
                     message = 'Received Alias Response from server: ' + JSON.stringify(response.status);
                     break;
 
-                // Our Identity API will return a 400 error if there is an issue with the requeest body
+                // Our Alias Request API will return a 400 error if there is an issue with the request body
                 // such as if the body is empty or one of the attirbutes is missing or malformed
                 // A 400 will return an error in the response body and will go through the happy path to report the error
                 case HTTP_BAD_REQUEST:
--- a/src/identityApiClient.ts
+++ b/src/identityApiClient.ts
@@ -1,4 +1,4 @@
-import Constants, { HTTP_ACCEPTED, HTTP_OK } from './constants';
+import Constants, { HTTP_ACCEPTED, HTTP_BAD_REQUEST, HTTP_OK } from './constants';
 import {
     AsyncUploader,
     FetchUploader,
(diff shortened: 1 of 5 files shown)


src/identityApiClient.ts
```diff
@@ -197,33 +233,62 @@ export default function IdentityAPIClient(
...
+                        }
+
+                    } else {
+                        message = 'Received Identity Response from server: ';
+                        message += JSON.stringify(identityResponse.responseText);
+                    }
+
+                    break;
+                    
+                // 401 and 403 have no bodies and should be rejected outright
```
alexs-mparticle, line 283: Note: Add specific comments as to what will cause a 401 or 403

src/identityApiClient.ts
```diff
@@ -197,33 +233,62 @@ export default function IdentityAPIClient(
...
             const response: Response = await uploader.upload(fetchPayload);
 
             let identityResponse: IIdentityResponse;
+            let message: string;
+
+            switch (response.status) {
+                case HTTP_ACCEPTED:
+                case HTTP_OK:
+
+                // 400 error will has a body and will go through the happy path to report the error
```
alexs-mparticle, line 248: Note: Add context to what would cause a 400

src/identityApiClient.ts
```diff
@@ -99,55 +114,76 @@ export default function IdentityAPIClient(
...
+                            response.status;
+                        
+                        if (errorResponse?.code) {
+                            message += ' - ' + errorResponse.code;
+                        }
+                    }
+
+                    break;
+                    
+                // 401 and 403 have no bodies and should be rejected outright
```
alexs-mparticle, line 171: Note: Rephrase to say 'All other error codes will hit this path'

src/identityApiClient.ts
```diff
@@ -99,55 +114,76 @@ export default function IdentityAPIClient(
...
+
+                        // https://go.mparticle.com/work/SQDSDKS-6670
+                        message =
+                            'Successfully sent forwarding stats to mParticle Servers';
                     }
-                    message =
-                        'Issue with sending Alias Request to mParticle Servers, received HTTP Code of ' +
-                        response.status;
+
+
```
rmi22186, line 153:
```suggestion
```

src/identityApiClient.ts
rmi22186: Please update this

src/identityApiClient.ts
rmi22186: Please update this

src/identityApiClient.ts
```diff
@@ -99,55 +114,76 @@ export default function IdentityAPIClient(
...
+                    if (response.json) {
+                        // HTTP responses of 202, 200, and 403 do not have a response.
+                        // response.json will always exist on a fetch, but can only be
+                        // await-ed when the response is not empty, otherwise it will
+                        // throw an error.
+                        try {
+                            aliasResponseBody = await response.json();
+                        } catch (e) {
+                            verbose('The request has no response body');
+                        }
```
rmi22186, line 138: Previously, we had this message because we weren't doing it inside of a switch statement, so it technically allowed for all other http status codes to be .  However, This is a bit confusing to me.  We say 202, 200, and 403 won't have a body, and so this switch statement will _always_ throw an error and go to the catch statement, right?

It also seems like `aliasResponseBody` is only ever used for grabbing the errors, and not in a successful alias request.  I feel like this function is still confusing. I think it should handle 200 and 202 separately from errors.

rmi22186 (changes requested):

src/identityApiClient.ts
rmi22186: Doing it in this switch case requires us to know that the only responses coming back are 202, 200, and 403. otherwise, it goes to the default and throws an error.  ultimately everything just calls `invokeAliasCallback` which is the main concern, but I have a small worry that there may be an error for a 20X that we aren't catching, and then we are telling the customer via a thrown Error that there's a 204, or 203.  not sure if these are realistic use cases, but I think that's what the code was doing in the past via verbose, and it not necessarily being an error.  Perhaps we should sync w Napo to get an idea of _every_ status code that can come back

src/identityApiClient.ts
```diff
@@ -99,55 +114,76 @@ export default function IdentityAPIClient(
...
+                        // XHRUploader returns the response as a string that we need to parse
+                        const xhrResponse = (response as unknown) as XMLHttpRequest;
+
+                        aliasResponseBody = xhrResponse.responseText
+                            ? JSON.parse(xhrResponse.responseText)
+                            : '';
+
+                        // https://go.mparticle.com/work/SQDSDKS-6670
+                        message =
+                            'Successfully sent forwarding stats to mParticle Servers';
```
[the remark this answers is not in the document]
alexs-mparticle: 😆

src/identityApiClient.ts
```diff
@@ -99,55 +114,72 @@ export default function IdentityAPIClient(
...
-                    message =
-                        'Successfully sent forwarding stats to mParticle Servers';
+                    message = 'Received Alias Response from server: ' + JSON.stringify(response.status);
                     break;
-                default:
-                    // 400 has an error message, but 403 doesn't
-                    if (aliasResponseBody?.message) {
-                        errorMessage = aliasResponseBody.message;
+
+                // Our Identity API will return a 400 error if there is an issue with the requeest body
```
rmi22186, line 129:
```suggestion
                // Our Alias Request API will return a 400 error if there is an issue with the request body
```

rmi22186 (changes requested):
minor nit

src/identityApiClient.ts
```diff
@@ -99,55 +114,72 @@ export default function IdentityAPIClient(
...
+                    message = 'Received Alias Response from server: ' + JSON.stringify(response.status);
                     break;
-                default:
-                    // 400 has an error message, but 403 doesn't
-                    if (aliasResponseBody?.message) {
-                        errorMessage = aliasResponseBody.message;
+
+                // Our Alias Request API will return a 400 error if there is an issue with the request body
+                // such as if the body is empty or one of the attirbutes is missing or malformed
+                // A 400 will return an error in the response body and will go through the happy path to report the error
```
rmi22186, line 131:
```suggestion
                // Our Alias Request API will 400 if there is an issue with the request body (ie timestamps are too far 
                // in the past or MPIDs don't exist).
                // A 400 will return an error in the response body and will go through the happy path to report the error
```

rmi22186 (approved):
last comment, then lgtm

PR #962, opened 2025-01-07
refactor: Create Interfaces for Events

alexs-mparticle:
## Instructions

 ## Summary
 - Create Interface Definitions for Events

 ## Testing Plan
 - N/A - This PR only defines new interfaces

 - Closes https://go.mparticle.com/work/SQDSDKS-5593

### DIFF
--- a/src/events.interfaces.ts
+++ b/src/events.interfaces.ts
@@ -19,11 +19,6 @@ import { CommonEventData } from '@mparticle/event-models';
 // Supports wrapping event handlers functions that will ideally return a specific type
 type EventHandlerFunction<T> = (element: HTMLLinkElement | HTMLFormElement) => T;
 
-// User options specified during the checkout process
-// e.g., FedEx, DHL, UPS for delivery options;
-// Visa, MasterCard, AmEx for payment options.
-type CheckoutOption = string;
-
 export interface IEvents {
     addEventHandler(
         domEvent: string,
@@ -35,7 +30,12 @@ export interface IEvents {
     logAST(): void;
     logCheckoutEvent(
         step: number,
-        option?: CheckoutOption,
+
+        // User options specified during the checkout process
+        // e.g., FedEx, DHL, UPS for delivery options;
+        // Visa, MasterCard, AmEx for payment options.
+        option?: string,
+
         attrs?: SDKEventAttrs,
         customFlags?: SDKEventCustomFlags
     ): void;
--- a/src/events.interfaces.ts
+++ b/src/events.interfaces.ts
@@ -1,0 +1,77 @@
+import {
+    Callback,
+    Dictionary,
+    SDKEventAttrs,
+    SDKEventOptions,
+    TransactionAttributes,
+} from '@mparticle/web-sdk';
+import {
+    BaseEvent,
+    SDKEventCustomFlags,
+    SDKProduct,
+    SDKProductImpression,
+    SDKPromotion,
+} from './sdkRuntimeModels';
+import { valueof } from './utils';
+import { EventType, ProductActionType, PromotionActionType } from './types';
+
+type dataFunction = (element: HTMLElement) => Dictionary<string>;
+type dataObject = Dictionary<string>;
+
+export interface IEvents {
+    addEventHandler(
+        domEvent: string,
+        selector: string | Node,
+        eventName: string,
+
+        // QUESTION: In what cases would data be a function?
+        data: dataObject | dataFunction,
+
+        eventType: valueof<typeof EventType>
+    ): void;
+    logCheckoutEvent(
+        step: number,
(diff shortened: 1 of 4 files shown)


src/events.interfaces.ts
```diff
@@ -0,0 +1,77 @@
...
+        data: dataObject | dataFunction,
+
+        eventType: valueof<typeof EventType>
+    ): void;
+    logCheckoutEvent(
+        step: number,
+        option?: SDKEventOptions,
+        attrs?: SDKEventAttrs,
+        customFlags?: SDKEventCustomFlags
+    );
```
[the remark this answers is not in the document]
rmi22186: several of the methods below do not have `void` and should

src/events.interfaces.ts
```diff
@@ -0,0 +1,77 @@
...
+    logPromotionEvent(
+        promotionType: valueof<typeof PromotionActionType>,
+        promotion: SDKPromotion,
+        attrs?: SDKEventAttrs,
+        customFlags?: SDKEventCustomFlags,
+        eventOptions?: SDKEventOptions
+    );
+    logPurchaseEvent(
+        product: SDKProduct,
+        transactionAttributes?: TransactionAttributes,
```
rmi22186, line 63: https://github.com/mParticle/mparticle-web-sdk/blob/master/src/events.js#L189-L192

swap the order of these arguments. it should match `logRefundEvent` below

src/events.interfaces.ts
```diff
@@ -0,0 +1,77 @@
...
+    );
+    logRefundEvent(
+        transactionAttributes: TransactionAttributes,
+        product: SDKProduct,
+        attrs?: SDKEventAttrs,
+        customFlags?: SDKEventCustomFlags
+    );
+    startTracking(callback: Callback): void;
+    stopTracking(): void;
+
```
rmi22186, line 75: this file is missing logAST

rmi22186: https://github.com/mParticle/mparticle-web-sdk/blob/master/src/events.js#L98

src/events.interfaces.ts
```diff
@@ -0,0 +1,77 @@
...
+type dataObject = Dictionary<string>;
+
+export interface IEvents {
+    addEventHandler(
+        domEvent: string,
+        selector: string | Node,
+        eventName: string,
+
+        // QUESTION: In what cases would data be a function?
+        data: dataObject | dataFunction,
```
rmi22186, line 28: This predates me all the way to the very first commit of the SDK. 
https://github.com/mParticle/mparticle-web-sdk/commit/050d15a1e9bf8f6a04b883ecc4c8404edbd00ee8#diff-95045420909226e4347b6b716fe6ae6fdfb25642d6571953b756744578d93180R657-R730

Unclear to me why a function would be passed.  Nothing in docs about this.  Bonus points if you can figure out why :)

src/events.interfaces.ts
rmi22186: and `logCommerceEvent`.  same file as above

rmi22186 (changes requested):

src/events.interfaces.ts
alexs-mparticle: So I was able to do some digging and this was intended to allow customers to wrap event handlers and pass them in as callbacks to `logEvent`. As we don't have this documented anywhere, I wonder if it's something that is even used. I think we should consider deprecating this function unless there's a good reason to keep it.

From what I can tell, we don't even have any tests around this.

rmi22186 (changes requested):
a couple more comments

src/events.interfaces.ts
```diff
@@ -0,0 +1,84 @@
...
+import { CommonEventData } from '@mparticle/event-models';
+
+// Supports wrapping event handlers functions that will ideally return a specific type
+type EventHandlerFunction<T> = (element: HTMLLinkElement | HTMLFormElement) => T;
+
+export interface IEvents {
+    addEventHandler(
+        domEvent: string,
+        selector: string | Node,
+        eventName: EventHandlerFunction<string> | string,
```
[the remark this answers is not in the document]
alexs-mparticle: The usage in the handler function treats both `eventName` and `data` as a possible function:

```
                self.logEvent({
                    messageType: Types.MessageType.PageEvent,
                    name:
                        typeof eventName === 'function'
                            ? eventName(element)
                            : eventName,
                    data: typeof data === 'function' ? data(element) : data,
                    eventType: eventType || Types.EventType.Other,
                });
```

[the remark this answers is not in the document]
rmi22186: i stand corrected!  strang

rmi22186 (approved):

PR #963, opened 2025-01-07
refactor: Simplify cookie sync manager

rmi22186:
## Instructions

 ## Summary
This PR makes the cookie sync manager less cognitively complex by:
1. Not triggering cookie sync when `pixelUrl` is empty
2. Removing the logic between previousMPID and current mpid.  To simplify, we always test to see if the new mpid has a last sync date that has expired or not
3. Simplify `performCookieSync` by removing a lot of logic from it to `attemptCookieSync`
4. Reduce nested if statements throughout `attemptCookieSync`.

 ## Testing Plan
Unit tests continue to run successfully.  Additional tests added to the jest tests, as well as removing any tests that dealt with previous MPID.

 - Closes https://go.mparticle.com/work/SQDSDKS-6915

### DIFF
--- a/src/cookieSyncManager.ts
+++ b/src/cookieSyncManager.ts
@@ -46,22 +46,6 @@ export interface ICookieSyncManager {
     ) => string;
 }
 
-const shouldPerformCookieSync = (
-    frequencyCap: number,
-    lastSyncDate?: number
-): boolean => {
-    // If there is no lastSyncDate, then there is no previous cookie sync, so we should sync the cookie
-    if (!lastSyncDate) {
-        return true;
-    }
-
-    // Otherwise, compare the last sync date to determine if it should do a cookie sync again
-    return (
-        new Date().getTime() >
-        new Date(lastSyncDate).getTime() + frequencyCap * DAYS_IN_MILLISECONDS
-    );
-};
-
 export default function CookieSyncManager(
     this: ICookieSyncManager,
     mpInstance: MParticleWebSDK
@@ -73,7 +57,9 @@ export default function CookieSyncManager(
         mpid: MPID,
         mpidIsNotInCookies?: boolean
     ): void => {
-        if (!mpid || mpInstance._Store.webviewBridgeEnabled) {
+        const { pixelConfigurations, webviewBridgeEnabled } = mpInstance._Store;
+
+        if (!mpid || webviewBridgeEnabled) {
             return;
         }
 
@@ -83,7 +69,6 @@ export default function CookieSyncManager(
             return;
         }
 
-        const { pixelConfigurations } = mpInstance._Store;
         pixelConfigurations.forEach((pixelSettings: IPixelConfiguration) => {
             // set requiresConsent to false to start each additional pixel configuration
             // set to true only if filteringConsenRuleValues.values.length exists
@@ -100,6 +85,10 @@ export default function CookieSyncManager(
             } = pixelSettings;
             const { values } = filteringConsentRuleValues || {};
             
+            if (isEmpty(pixelUrl) && isEmpty(redirectUrl)) {
+                return;
+            }
+
             if (!isEmpty(values)) {
                 requiresConsent = true;
             }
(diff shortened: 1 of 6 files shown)


test/jest/cookieSyncManager.spec.ts
```diff
@@ -13,10 +13,12 @@ const pixelSettings: IPixelConfiguration = {
     isProduction: true,
     settings: {},
     frequencyCap: 14,
-    pixelUrl: '',
-    redirectUrl: '',
+    pixelUrl: 'https://test.com',
+    redirectUrl: '?redirect=https://redirect.com&mpid=%%mpid%%',
```
alexs-mparticle, line 17: If we are populating this object with default values, we should have a test to handle the expected behavior if they are missing. I don't recall if I did that.

test/jest/cookieSyncManager.spec.ts
```diff
@@ -329,35 +240,57 @@ describe('CookieSyncManager', () => {
...
                 testMPID,
                 {},
-                undefined,
                 true,
                 false, 
             );
         });
 
-        it('should sync cookies when there was not a previous cookie-sync', () => {
+        it.only('should sync cookies when there was not a previous cookie-sync', () => {
```
alexs-mparticle, line 255: 🦤 

```suggestion
        it('should sync cookies when there was not a previous cookie-sync', () => {
```

test/jest/cookieSyncManager.spec.ts
```diff
@@ -396,6 +329,7 @@ describe('CookieSyncManager', () => {
                     pixelConfigurations: [pixelSettings],
                 },
                 _Persistence: {
+                    // FIXME: why is this here?  it doesn't exist in the SDK (it is used elsewhere too)
```
alexs-mparticle, line 332: Let me know if you need to pair on this.

src/cookieSyncManager.ts
```diff
@@ -104,63 +123,39 @@ export default function CookieSyncManager(
...
-                    );
-                }
-                return;
-            } else {
-                if (!persistence || !persistence[mpid]) {
-                    return;
-                }
-
+            // set up csd object if it doesn't exist
+            if (persistence && persistence[mpid]) {
```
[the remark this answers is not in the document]
rmi22186: i like this.  i'll have to move the new lines into the foreach block though because of when some of the variables are available.

test/jest/cookieSyncManager.spec.ts
```diff
@@ -496,60 +406,73 @@ describe('CookieSyncManager', () => {
...
 
-            const loggerSpy = jest.fn();
-
             const mockMPInstance = ({
                 _Store: {
                     webviewBridgeEnabled: false,
                     pixelConfigurations: [pixelSettings],
                 },
                 _Persistence: {
+                    // FIXME: why is this here?  it doesn't exist in the SDK (it is used elsewhere too)
```
alexs-mparticle, line 415: Is this still relevant?

test/jest/cookieSyncManager.spec.ts
```diff
@@ -496,60 +406,73 @@ describe('CookieSyncManager', () => {
...
                 },
             } as unknown) as MParticleWebSDK;
 
             const cookieSyncManager = new CookieSyncManager(mockMPInstance);
 
-            let cookieSyncDates = [];
+            jest.useFakeTimers();
+
+            // Mock the Date constructor
+            // const mockDate = new Date(0); // Epoch time: 0
```
alexs-mparticle, line 438: Does this still need to be commented out?

src/cookieSyncManager.ts
```diff
@@ -67,100 +70,83 @@ export default function CookieSyncManager(
...
+                return;
+            }
 
             // Url for cookie sync pixel
-            const pixelUrl = replaceAmpWithAmpersand(pixelSettings.pixelUrl);
-
-            const redirectUrl = pixelSettings.redirectUrl
-                ? replaceAmpWithAmpersand(pixelSettings.redirectUrl)
+            const finalPixelUrl = replaceAmpWithAmpersand(pixelUrl);
+            const finalDirectUrl = redirectUrl
```
alexs-mparticle, line 120: I think these should be renamed from `final` to `modified` or `sanitized`. In my experience "final" is never "final".

rmi22186: true story

test/jest/cookieSyncManager.spec.ts
rmi22186: as we discussed, i have removed this comment as well as the line in question below it and elsewhere

alexs-mparticle (changes requested):

src/cookieSyncManager.ts
```diff
@@ -67,137 +53,108 @@ export default function CookieSyncManager(
...
+            // because a cookie sync can only occur once a user either consents or doesn't.
+            // we should not check if it's enabled if the user has a blank consent
+            if (requiresConsent && mpidIsNotInCookies) {
+                return;
+            }
 
-            // Tells you how often we should do a cookie sync (in days)
-            const frequencyCap = pixelSettings.frequencyCap;
+            const { isEnabledForUserConsent } = mpInstance._Consent;
+            if (!isEnabledForUserConsent(filteringConsentRuleValues, mpInstance.Identity.getCurrentUser())) {
```
rmi22186, line 101:
```suggestion
            const { isEnabledForUserConsent } = mpInstance._Consent;

            if (!isEnabledForUserConsent(filteringConsentRuleValues, mpInstance.Identity.getCurrentUser())) {
```

alexs-mparticle (approved):

PR #964, opened 2025-01-15
refactor: Create Interfaces for eCommerce

alexs-mparticle:
## Instructions

 ## Summary
 - Defines TypeScript interfaces for eCommerce

 ## Testing Plan
 - N/A - these are type definitions

 - Closes https://go.mparticle.com/work/SQDSDKS-5678

### DIFF
--- a/src/ecommerce.interfaces.ts
+++ b/src/ecommerce.interfaces.ts
@@ -187,7 +187,7 @@ export interface IECommerce extends IECommerceShared {
         promotion: Promotion
     ): void;
     extractTransactionId(
-        attributes: ExtractedTransactionId;
+        attributes: ExtractedTransactionId,
         productAction: ProductAction
     ): void;
     generateExpandedEcommerceName(eventName: string, plusOne: boolean): string;
--- a/src/ecommerce.interfaces.ts
+++ b/src/ecommerce.interfaces.ts
@@ -120,11 +120,45 @@ export interface SDKECommerceAPI extends IECommerceShared {
     ): void;
 }
 
+interface ExtractedActionAttributes {
+    "Affiliation"?: string;
+    "Coupon Code"?: string;
+    "Total Amount"?: number;
+    "Shipping Amount"?: string;
+    "Tax Amount"?: string;
+    "Checkout Option"?: string;
+    "Checkout Step"?: number;
+    "Transaction ID"?: string;
+}
+interface ExtractedProductAttributes {
+    "Coupon Code"?: string;
+    "Brand"?: string;
+    "Category"?: string;
+    "Name"?: string;
+    "Id"?: string;
+    "Item Price"?: string;
+    "Quantity"?: string;
+    "Position"?: number;
+    "Variant"?: string;
+    "Total Product Amount": string;
+}
+interface ExtractedPromotionAttributes {
+    "Id"?: string;
+    "Creative"?: string;
+    "Name"?: string;
+    "Position"?: number;
+}
+interface ExtractedTransactionId {
+    "Transaction ID"?: string;
+}
+
 // Used for the private `_Ecommerce` namespace
 export interface IECommerce extends IECommerceShared {
     buildProductList(event: SDKEvent, product: Product | Product[]): Product[];
     convertProductActionToEventType(
         productActionType: valueof<typeof ProductActionType>
+
+        // https://go.mparticle.com/work/SQDSDKS-4801
     ): typeof CommerceEventType | typeof EventType | null;
     convertPromotionActionToEventType(
         promotionActionType: valueof<typeof PromotionActionType>
@@ -141,19 +175,19 @@ export interface IECommerce extends IECommerceShared {
     expandProductImpression(commerceEvent: CommerceEvent): SDKEvent[];
     expandPromotionAction(commerceEvent: CommerceEvent): SDKEvent[];
     extractActionAttributes(
-        attributes: Record<string, string | number>,
+        attributes: ExtractedActionAttributes,
         productAction: ProductAction
     ): void;
     extractProductAttributes(
-        attributes: Record<string, string | number>,
+        attributes: ExtractedProductAttributes,
         product: Product
     ): void;
     extractPromotionAttributes(
-        attributes: Record<string, string | number>,
+        attributes: ExtractedPromotionAttributes,
         promotion: Promotion
     ): void;
     extractTransactionId(
-        attributes: Record<string, string | number>,
+        attributes: ExtractedTransactionId;
         productAction: ProductAction
     ): void;
     generateExpandedEcommerceName(eventName: string, plusOne: boolean): string;
--- a/src/ecommerce.interfaces.ts
+++ b/src/ecommerce.interfaces.ts
@@ -36,7 +36,7 @@ interface IECommerceShared {
         brand?: string,
         position?: number,
         couponCode?: string,
-        attributes?: SDKEventAttrs,
+        attributes?: SDKEventAttrs
     ): SDKProduct | null;
     createImpression(name: string, product: Product): SDKImpression | null;
     createPromotion(
@@ -64,11 +64,6 @@ export interface SDKCart {
 
 // Used for the public `eCommerce` namespace
 export interface SDKECommerceAPI extends IECommerceShared {
-    /*
-     * @deprecated
-     */
-    Cart: SDKCart;
-
     logCheckout(
         step: number,
         option?: string,
@@ -101,6 +96,11 @@ export interface SDKECommerceAPI extends IECommerceShared {
     /*
      * @deprecated
      */
+    Cart: SDKCart;
+
+    /*
+     * @deprecated
+     */
     logPurchase(
         transactionAttributes: TransactionAttributes,
         product: SDKProduct | SDKProduct[],
@@ -108,6 +108,7 @@ export interface SDKECommerceAPI extends IECommerceShared {
         attrs?: SDKEventAttrs,
         customFlags?: SDKEventCustomFlags
     ): void;
+
     /*
      * @deprecated
      */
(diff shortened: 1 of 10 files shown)


src/ecommerce.interfaces.ts
```diff
@@ -0,0 +1,160 @@
...
+        clearCart?: boolean,
+        attrs?: SDKEventAttrs,
+        customFlags?: SDKEventCustomFlags
+    ): void;
+    setCurrencyCode(code: string): void;
+}
+
+// Used for the private `_Ecommerce` namespace
+export interface IECommerce extends IECommerceShared {
+    buildProductList(event: any, product: Product | Product[]): Product[];
```
rmi22186, line 118:
```suggestion
    buildProductList(event: SDKEvent, product: Product | Product[]): Product[];
```

src/ecommerce.interfaces.ts
```diff
@@ -0,0 +1,160 @@
...
+    convertProductActionToEventType(
+        productActionType: valueof<typeof ProductActionType>
+    ): typeof CommerceEventType | typeof EventType | null;
```
rmi22186, line 121: Interesting. i looked at the code and you're right that we are returning one instance of `EventType.Unknown`, that is probably a bug and could e replaced with ProductActionType.Unknown.  I'll let you decide if you want to put that in as a bug fix before this, or after this. I go back and forth about putting the fix in as part of this PR since this should only be for interface creation.

src/ecommerce.interfaces.ts
```diff
@@ -0,0 +1,160 @@
...
+    convertPromotionActionToEventType(
+        promotionActionType: valueof<typeof PromotionActionType>
+    ): typeof CommerceEventType | null;
+    convertTransactionAttributesToProductAction(
+        transactionAttributes: TransactionAttributes,
+        productAction: ProductAction
+    ): void;
```
rmi22186, line 128: Since I don't think either of us like the fact that this is a `void` and mutates state, I'd recommend making a ticket for a future refactor as tech debt.

src/ecommerce.interfaces.ts
```diff
@@ -0,0 +1,160 @@
...
+    createCommerceEventObject(
+        customFlags: SDKEventCustomFlags,
+        options?: SDKEventOptions
+    ): SDKEvent | null;
+    expandProductAction(commerceEvent: CommerceEvent): SDKEvent[];
+    expandProductImpression(commerceEvent: CommerceEvent): SDKEvent[];
+    expandPromotionAction(commerceEvent: CommerceEvent): SDKEvent[];
+    extractActionAttributes(
+        attributes: Record<string, string | number>,
```
rmi22186, line 137: Can you add an interfacte/type for these product attributes to be used here and in the other `extract` methods below?  Could be a new ticket also

rmi22186: i also think the same comment above about `void`/mutation vs returning a new object can apply to the `extract` methods as well, so make a note of that in the ticket

test/src/tests-identities-attributes.ts
```diff
@@ -803,7 +803,7 @@ describe('identities and attributes', function() {
             product2 = mParticle.eCommerce.createProduct('Android', 'SKU2', 1);
         waitForCondition(hasIdentifyReturned)
         .then(() =>  {
-        mParticle.eCommerce.Cart.add([product1, product2]);
+        mParticle.eCommerce.Cart.add([product1, product2], null);
```
rmi22186, line 806:
```suggestion
        mParticle.eCommerce.Cart.add([product1, product2], false);
```

rmi22186: i'm surprised that it's accepting `null`

test/src/tests-identity.ts
```diff
@@ -4799,11 +4799,11 @@ describe('identity', function() {
             waitForCondition(hasIdentifyReturned)
             .then(() => {
 
-            const product = mParticle.eCommerce.createProduct(
+            const product: Product = mParticle.eCommerce.createProduct(
```
rmi22186, line 4802: I assume this and the below should all be `SDKProduct` and not `Product`
```suggestion
            const product: SDKProduct = mParticle.eCommerce.createProduct(
```

rmi22186 (changes requested):
looks great!

src/ecommerce.interfaces.ts
```diff
@@ -0,0 +1,160 @@
...
+        name: string,
+        sku: string | number,
+        price: string | number,
+        quantity?: string | number,
+        variant?: string,
+        category?: string,
+        brand?: string,
+        position?: number,
+        couponCode?: string,
+        attributes?: Record<string, string | number>
```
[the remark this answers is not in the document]
alexs-mparticle: Good callout

src/ecommerce.interfaces.ts
alexs-mparticle: I think we should do it as part of the actual migration of eCommerce, which we already have a ticket for: https://go.mparticle.com/work/SQDSDKS-4801

src/ecommerce.interfaces.ts
```diff
@@ -96,28 +96,69 @@ export interface SDKECommerceAPI extends IECommerceShared {
...
     logPurchase(
         transactionAttributes: TransactionAttributes,
         product: SDKProduct | SDKProduct[],
         clearCart?: boolean,
         attrs?: SDKEventAttrs,
         customFlags?: SDKEventCustomFlags
     ): void;
+    /*
+     * @deprecated
+     */
```
rmi22186, line 113: for consistency with spacing, either add a line before the deprecation on line 111 or remove the new line at 100

src/ecommerce.interfaces.ts
```diff
@@ -96,28 +96,69 @@ export interface SDKECommerceAPI extends IECommerceShared {
...
+    "Checkout Step"?: number;
+    "Transaction ID"?: string;
+}
+interface ExtractedProductAttributes {
+    "Coupon Code"?: string;
+    "Brand"?: string;
+    "Category"?: string;
+    "Name"?: string;
+    "Id"?: string;
+    "Item Price"?: string;
```
rmi22186, line 139: a customer is allowed to add this as a string or number, but then we parse it into a number

https://github.com/mParticle/mparticle-web-sdk/blob/master/src/ecommerce.js#L267-L273

src/ecommerce.interfaces.ts
```diff
@@ -96,28 +96,69 @@ export interface SDKECommerceAPI extends IECommerceShared {
...
+    "Quantity"?: string;
```
rmi22186, line 140: a customer is allowed to add it as a string or number, but then we parse it into a number https://github.com/mParticle/mparticle-web-sdk/blob/master/src/ecommerce.js#L282-L287

src/ecommerce.interfaces.ts
```diff
@@ -96,28 +96,69 @@ export interface SDKECommerceAPI extends IECommerceShared {
...
+    "Position"?: number;
+    "Variant"?: string;
+    "Total Product Amount": string;
```
rmi22186, line 143: https://github.com/mParticle/mparticle-web-sdk/blob/master/src/ecommerce.js#L299 makes this a number

```suggestion
    "Total Product Amount": number;
```

src/ecommerce.interfaces.ts
```diff
@@ -96,28 +96,69 @@ export interface SDKECommerceAPI extends IECommerceShared {
...
     ): void;
-    setCurrencyCode(code: string): void;
+}
+
+interface ExtractedActionAttributes {
+    "Affiliation"?: string;
+    "Coupon Code"?: string;
+    "Total Amount"?: number;
+    "Shipping Amount"?: string;
+    "Tax Amount"?: string;
```
rmi22186, line 128: this gets sanitized and is a number - https://github.com/mParticle/mparticle-web-sdk/blob/master/src/ecommerce.js#L33-L37
```suggestion
    "Tax Amount"?: number;
```

src/ecommerce.interfaces.ts
```diff
@@ -96,28 +96,69 @@ export interface SDKECommerceAPI extends IECommerceShared {
...
         customFlags?: SDKEventCustomFlags
     ): void;
-    setCurrencyCode(code: string): void;
+}
+
+interface ExtractedActionAttributes {
+    "Affiliation"?: string;
+    "Coupon Code"?: string;
+    "Total Amount"?: number;
+    "Shipping Amount"?: string;
```
rmi22186, line 127: this gets sanitized and is a number - https://github.com/mParticle/mparticle-web-sdk/blob/master/src/ecommerce.js#L28-L31

src/ecommerce.js
```diff
@@ -103,8 +105,11 @@ export default function Ecommerce(mpInstance) {
                 return Types.CommerceEventType.ProductRemoveFromCart;
             case Types.ProductActionType.RemoveFromWishlist:
                 return Types.CommerceEventType.ProductRemoveFromWishlist;
+
+            // https://go.mparticle.com/work/SQDSDKS-4801
             case Types.ProductActionType.Unknown:
                 return Types.EventType.Unknown;
+
```
rmi22186, line 112:
```suggestion
```

src/ecommerce.interfaces.ts
rmi22186: sounds good.  i see it linked in comments now.

test/src/tests-identity.ts
rmi22186: this comment is still pending

rmi22186 (changes requested):

src/ecommerce.js
alexs-mparticle: The space is intentional to show that the specific case relates to the worklink above it. IMO it makes it easier to read and highlights the relationship between the case and the worklink. I can remove if you feel strongly about it.

src/sdkRuntimeModels.ts
```diff
@@ -150,7 +156,7 @@ export interface SDKProduct {
     Position?: number;
     CouponCode?: string;
     TotalAmount?: number;
-    Attributes?: { [key: string]: string };
+    Attributes?: Record<string, unknown> | undefined;
```
alexs-mparticle, line 159: Note to self: Can we remove `undefined` because this is an optional attribute.

alexs-mparticle, line 159: `creatProduct` technically assigns this to `null` if there are no attributes. Probably should be a future refactor to have it be `undefined`.

```suggestion
    Attributes?: Record<string, unknown> | null;
```

src/ecommerce.interfaces.ts
```diff
@@ -36,7 +36,7 @@ interface IECommerceShared {
         brand?: string,
         position?: number,
         couponCode?: string,
-        attributes?: SDKEventAttrs,
+        attributes?: SDKEventAttrs
```
rmi22186, line 39: ,

src/sdkToEventsApiConverter.ts
```diff
@@ -339,7 +339,7 @@ export function convertProducts(
             price: sdkProduct.Price,
             quantity: sdkProduct.Quantity,
             coupon_code: sdkProduct.CouponCode,
-            custom_attributes: sdkProduct.Attributes,
+            custom_attributes: sdkProduct.Attributes as Record<string, string>,
```
rmi22186, line 342: see if this should be string, unknown

alexs-mparticle: `Record<string, string>` doesn't work because `EventsApi.Product` requires `custom_attributes` to be `[key: string]: string`

rmi22186 (approved):

PR #966, opened 2025-01-17
feat: Include consent query parameters for cookie sync

rmi22186:
## Instructions

 ## Summary
* If a consent string is not appended to GMP Cookie Syncs, Google will exclude these users from audiences.  TCF (Transparency and Consent Framework) has specific implementation instructions that all cookie sync customers follow in order to include if someone has or has not consented
* All Consent Management Platforms (such as OneTrust), are required to implement a function on `window.__tcfapi` to return TCF strings which contains what a user has/has not consented.  This gets appended to the end of a cookie sync.
* We call the [API using getInAppTCData](https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/TCFv2/IAB%20Tech%20Lab%20-%20Consent%20string%20and%20vendor%20list%20formats%20v2.md#getinapptcdata), which then [returns](https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/TCFv2/IAB%20Tech%20Lab%20-%20CMP%20API%20v2.md#tcdata) a `tcData` object, and we only need the keys `gdprApplies` and `tcString`

 ## Testing Plan
Added unit tests

 - Closes https://go.mparticle.com/work/SQDSDKS-6991


src/cookieSyncManager.ts
```diff
@@ -113,13 +126,18 @@ export default function CookieSyncManager(
...
-
-            self.performCookieSync(
-                fullUrl,
-                moduleId.toString(),
-                mpid,
-                cookieSyncDates
-            );
+            const moduleIdString = moduleId.toString();
+
+            if (isTcfApiAvailable()) {
```
alexs-mparticle, line 131: I see that this function calls with `performCookieSync` but seems to only define a callback and appends strings to the url. I see why you're doing it this way, but I wonder if it would be more readable (and possibly prevent side effects) if we break out the callback logic and have it simply return a string that then gets passed into `performCookieSync`. It might make things more obvious and less coupled.

src/cookieSyncManager.ts
```diff
@@ -159,4 +204,12 @@ export const isLastSyncDateExpired = (
...
\ No newline at end of file
+};
+
+export const isTcfApiAvailable = (): boolean => {
+    if (typeof window.__tcfapi === 'function') {
+        return true;
+    } else {
+        return false;
+    }
+}
```
alexs-mparticle, line 215: I think we may have a util that checks if something is a function (I could be wrong). But if we do, you can probably make this a slick one-liner. 😎 

```suggestion
export const isTcfApiAvailable = (): boolean =>  isFunction(window.__tcfapi)';
```

alexs-mparticle (changes requested):

src/cookieSyncManager.ts
```diff
@@ -146,6 +155,10 @@ export default function CookieSyncManager(
         mpid: MPID,
         cookieSyncDates: CookieSyncDates,
     ): void => {
+        console.log(url)
+        console.log(moduleId);
+        console.log(mpid);
+        console.log(cookieSyncDates)
```
rmi22186, line 161: had left these here bc a test is failing that i couldn't sort out last night.

test/jest/cookieSyncManager.spec.ts
```diff
@@ -611,82 +608,77 @@ describe('CookieSyncManager', () => {
...
+                await appendGdprConsentUrl(mockUrl, mockMpInstance.Logger);
+            } catch(e) {
+                expect(errorLoggerSpy).toHaveBeenCalledWith('Test Error');
+            }
+        });
 
-            // Ensure the error was logged (if applicable)
-            expect(errorLoggerSpy).toHaveBeenCalledWith('Test Error');
+        describe.only('#integration test', () => {
+            it('should handle errors properly when calling attemptCookieSync and the callback fails', () => {
```
rmi22186, line 643: This test fails for some reason.  I'd like to discuss on zoom

src/cookieSyncManager.ts
```diff
@@ -159,4 +202,30 @@ export const isLastSyncDateExpired = (
...
+                    resolve(`${url}&gdpr=${gdprApplies}&gdpr_consent=${tcString}`);
+                } else {
+                    resolve(url); // No GDPR data; fallback to original URL
+                }
+            }
+
+            // `getInAppTCData` is the function name
+            // 2 is the version of TCF (2.2 as of 1/22/2025)
+            // callback 
+            window.__tcfapi('getInAppTCData', 2, tcfAPICallBack);
```
alexs-mparticle, line 225: Might be a nitpick on my part but I think this makes it more readable and obvious.

```suggestion
            window.__tcfapi(
                'getInAppTCData',  // the function name
                2,                              // Version of TCF (2.2 as of 1/22/2025)
                tcfAPICallBack,       // callback function
            );
```

alexs-mparticle (changes requested):

src/cookieSyncManager.ts
```diff
@@ -1,12 +1,14 @@
...
     Dictionary,
     isEmpty,
-    createCookieSyncUrl,
+    createInitialCookieSyncUrl,
+    isFunction,
 } from './utils';
 import Constants from './constants';
 import { MParticleWebSDK } from './sdkRuntimeModels';
-import { MPID } from '@mparticle/web-sdk';
+import { Logger, MPID } from '@mparticle/web-sdk';
```
alexs-mparticle, line 9:
```suggestion
import { MPID } from '@mparticle/web-sdk';
```

src/cookieSyncManager.ts
```diff
@@ -15,6 +17,12 @@ export const DAYS_IN_MILLISECONDS = 1000 * 60 * 60 * 24;
 
 export type CookieSyncDates = Dictionary<number>; 
 
+declare global {
+    interface Window {
+        __tcfapi: any;
```
alexs-mparticle, line 22: Can we come up with a simple definition?

test/jest/cookieSyncManager.spec.ts
```diff
@@ -553,4 +556,126 @@ describe('CookieSyncManager', () => {
...
+            _Persistence: {
+                saveUserCookieSyncDatesToPersistence: jest.fn(),
+            },
+            Identity: {
+                getCurrentUser: jest.fn(),
+            },
+        } as unknown) as MParticleWebSDK;
+
+        beforeEach(() => {
+            cookieSyncManager = new CookieSyncManager(mockMpInstance);
```
alexs-mparticle, line 587: Is this actually being used after the last reactor we discussed?

test/jest/cookieSyncManager.spec.ts
```diff
@@ -553,4 +556,126 @@ describe('CookieSyncManager', () => {
...
+            expect(isTcfApiAvailable()).toBe(true)
+        });
+    });
+
+    describe('#appendGdprConsentUrl', () => {
+        const verboseLoggerSpy = jest.fn();
+        const errorLoggerSpy = jest.fn();
+        const mockUrl = 'https://example.com/cookie-sync';
+
+        let cookieSyncManager: any;
```
alexs-mparticle, line 572: Do we still need this?

alexs-mparticle (changes requested):

PR #967, opened 2025-01-21
test: Remove set timeouts from tests and other cleanup

rmi22186:
## Instructions

 ## Summary
In this PR, there is a lot of cleanup
* The main refactor is in cookie sync.  I removed the `setTimeout` from the mock for `onload` and have it immediately call the callback, to avoid having setTImeout elsewhere in the cookie sync manager which has historically been very flakey. 
* removed `Should` and replaced with `expect`
* used `hasConfigurationReturned` from Utils to keep things DRY
* cleaned up function() {} to () => {} in some places

There is plenty more to do, but I was already going beyond the scope of this jira ticket.

 ## Testing Plan

 - Closes https://go.mparticle.com/work/SQDSDKS-6916

### DIFF
--- a/test/src/tests-cookie-syncing.ts
+++ b/test/src/tests-cookie-syncing.ts
@@ -6,7 +6,12 @@ import { MParticleWebSDK } from '../../src/sdkRuntimeModels';
 import { IMParticleUser } from '../../src/identity-user-interfaces';
 import { IPixelConfiguration } from '../../src/cookieSyncManager';
 import { IConsentRules } from '../../src/consent';
-const { fetchMockSuccess, waitForCondition, hasIdentifyReturned } = Utils;
+const {
+    fetchMockSuccess,
+    waitForCondition,
+    hasIdentifyReturned,
+    hasConfigurationReturned
+} = Utils;
 
 const { setLocalStorage, MockForwarder, getLocalStorage } = Utils;
 
@@ -32,7 +37,6 @@ declare global {
 const mParticle = window.mParticle;
 
 describe('cookie syncing', function() {
-    const timeout = 100;
     // Have a reference to createElement function to reset after all cookie sync
     // tests have run
     const originalCreateElementFunction = window.document.createElement;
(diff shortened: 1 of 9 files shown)


test/src/tests-cookie-syncing.ts
```diff
@@ -1011,57 +934,48 @@ describe('cookie syncing', function() {
...
-                .getInstance()
-                .Consent.createConsentState()
-                .addGDPRConsentState(
-                    'foo purpose 1',
-                    mParticle.getInstance().Consent.createGDPRConsent(false)
-                );
-
-            mParticle.Identity.getCurrentUser().setConsentState(
-                falseConsentState
+        await waitForCondition(hasIdentifyReturned)
```
alexs-mparticle, line 937:
```suggestion
        await waitForCondition(hasIdentifyReturned);
```

test/src/tests-cookie-syncing.ts
```diff
@@ -45,11 +49,17 @@ describe('cookie syncing', function() {
...
+                // Add an `onload` mock that simulates the browser loading the image
+                Object.defineProperty(element, 'onload', {
+                    set(callback) {
+                    // Automatically invoke the callback to simulate the `load` event
+                        callback(new Event('load'));
+                    },
+                });
+            }
+
+        return element;
```
rmi22186, line 61: indent

test/src/tests-cookie-syncing.ts
```diff
@@ -381,11 +343,9 @@ describe('cookie syncing', function() {
             ._Consent.isEnabledForUserConsent(filteringConsentRuleValues, null);
 
         expect(enabled).to.not.be.ok;
-
-        done();
     });
 
-    it("should disable cookie sync if 'Do Not Forward' when 'Consent Rejected' is selected and user consent is rejected", function(done) {
+    it("should disable cookie sync if 'Do Not Forward' when 'Consent Rejected' is selected and user consent is rejected", function() {
```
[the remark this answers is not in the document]
rmi22186: yea let's do that in the future.  i already did stuff that's out of scope...and it's getting a bit late :)

alexs-mparticle (approved):

PR #968, opened 2025-01-22
refactor: Migrate mParticle Instance to TypeScript

alexs-mparticle:
## Instructions

 ## Summary
 - Migrates mP Instance to TypeScript
 - Splits out mP Instance and Instance Manager Definitions

 ## Testing Plan

 - Closes https://go.mparticle.com/work/SQDSDKS-4804
 - Relies on #969 to be merged in first

### DIFF
--- a/src/configAPIClient.ts
+++ b/src/configAPIClient.ts
@@ -2,7 +2,6 @@ import { DataPlanConfig } from '@mparticle/web-sdk';
 import {
     BooleanStringLowerCase,
     DataPlanResult,
-    MParticleWebSDK,
     SDKEventCustomFlags,
     SDKInitConfig,
 } from './sdkRuntimeModels';
@@ -14,6 +13,7 @@ import {
     XHRUploader,
 } from './uploaders';
 import { IPixelConfiguration } from './cookieSyncManager';
+import { IMParticleWebSDKInstance } from './mp-instance';
 
 export interface IKitConfigs extends IKitFilterSettings {
     name: string;
@@ -82,7 +82,7 @@ export interface IConfigResponse {
 export interface IConfigAPIClient {
     apiKey: string;
     config: SDKInitConfig;
-    mpInstance: MParticleWebSDK;
+    mpInstance: IMParticleWebSDKInstance;
     getSDKConfiguration: () => Promise<IConfigResponse>;
 }
 
@@ -113,7 +113,7 @@ export default function ConfigAPIClient(
     this: IConfigAPIClient,
     apiKey: string,
     config: SDKInitConfig,
-    mpInstance: MParticleWebSDK
+    mpInstance: IMParticleWebSDKInstance
 ): void {
     const baseUrl = 'https://' + mpInstance._Store.SDKConfig.configUrl;
     const { isDevelopmentMode } = config;
--- a/src/mp-instance.ts
+++ b/src/mp-instance.ts
@@ -84,7 +84,6 @@ export interface IMParticleWebSDKInstance extends MParticleWebSDK {
     _Store: IStore;
     _instanceName: string;
     _preInit: IPreInit;
-
 }
 
 const { Messages, HTTPCodes, FeatureFlags } = Constants;
--- a/src/mp-instance.ts
+++ b/src/mp-instance.ts
@@ -16,31 +16,76 @@
 //  Uses portions of code from jQuery
 //  jQuery v1.10.2 | (c) 2005, 2013 jQuery Foundation, Inc. | jquery.org/license
 
-import Types from './types';
+import { EventType, IdentityType, CommerceEventType, PromotionActionType, ProductActionType, MessageType } from './types';
 import Constants from './constants';
-import APIClient from './apiClient';
+import APIClient, { IAPIClient } from './apiClient';
 import Helpers from './helpers';
 import NativeSdkHelpers from './nativeSdkHelpers';
-import CookieSyncManager from './cookieSyncManager';
-import SessionManager from './sessionManager';
+import CookieSyncManager, { ICookieSyncManager } from './cookieSyncManager';
+import SessionManager, { ISessionManager } from './sessionManager';
 import Ecommerce from './ecommerce';
-import Store from './store';
+import Store, { IStore } from './store';
 import Logger from './logger';
 import Persistence from './persistence';
 import Events from './events';
 import Forwarders from './forwarders';
-import ServerModel from './serverModel';
+import ServerModel, { IServerModel } from './serverModel';
 import ForwardingStatsUploader from './forwardingStatsUploader';
 import Identity from './identity';
-import Consent from './consent';
+import Consent, { IConsent } from './consent';
 import KitBlocker from './kitBlocking';
 import ConfigAPIClient from './configAPIClient';
 import IdentityAPIClient from './identityApiClient';
-import { isFunction } from './utils';
+import { isFunction, valueof } from './utils';
 import { LocalStorageVault } from './vault';
 import { removeExpiredIdentityCacheDates } from './identity-utils';
 import IntegrationCapture from './integrationCapture';
-import { processReadyQueue } from './pre-init-utils';
+import { IPreInit, processReadyQueue } from './pre-init-utils';
+import { BaseEvent, MParticleWebSDK, SDKHelpersApi } from './sdkRuntimeModels';
+import { Dictionary, SDKEventAttrs } from '@mparticle/web-sdk';
+import { IIdentity } from './identity.interfaces';
+import { IEvents } from './events.interfaces';
+import { IECommerce } from './ecommerce.interfaces';
+import { INativeSdkHelpers } from './nativeSdkHelpers.interfaces';
+import { IPersistence } from './persistence.interfaces';
+
+export interface IErrorLogMessage {
+    message?: string;
+    name?: string;
+    stack?: string;
+}
+
+export interface IErrorLogMessageMinified {
+    m?: string;
+    s?: string;
+    t?: string;
+}
+
+export type IntegrationDelays = Dictionary<boolean>;
+
+// https://go.mparticle.com/work/SQDSDKS-6949
+export interface IMParticleWebSDKInstance extends MParticleWebSDK {
+    // Private Properties
+    _APIClient: IAPIClient;
+    _Consent: IConsent;
+    _CookieSyncManager: ICookieSyncManager;
+    _Ecommerce: IECommerce;
+    _Events: IEvents;
+    _Forwarders: any; // https://go.mparticle.com/work/SQDSDKS-5767
+    _ForwardingStatsUploader: ForwardingStatsUploader;
+    _Helpers: SDKHelpersApi;
+    _Identity: IIdentity;
+    _IdentityAPIClient: typeof IdentityAPIClient;
+    _IntegrationCapture: IntegrationCapture;
+    _NativeSdkHelpers: INativeSdkHelpers;
+    _Persistence: IPersistence;
+    _SessionManager: ISessionManager;
+    _ServerModel: IServerModel;
+    _Store: IStore;
+    _instanceName: string;
+    _preInit: IPreInit;
+
+}
 
 const { Messages, HTTPCodes, FeatureFlags } = Constants;
 const { ReportBatching, CaptureIntegrationSpecificIds } = FeatureFlags;
@@ -59,8 +104,8 @@ const { StartingInitialization } = Messages.InformationMessages;
  * @class mParticle & mParticleInstance
  */
 
-export default function mParticleInstance(instanceName) {
-    var self = this;
+export default function mParticleInstance(this: IMParticleWebSDKInstance, instanceName: string) {
+    const self = this;
     // These classes are for internal use only. Not documented for public consumption
     this._instanceName = instanceName;
     this._NativeSdkHelpers = new NativeSdkHelpers(this);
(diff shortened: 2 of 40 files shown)


src/mp-instance.ts
```diff
@@ -16,31 +16,98 @@
...
+    reset(instance: IMParticleWebSDKInstance): void;
+    ready(f: Function): void;
+    isInitialized(): boolean;
+    getEnvironment(): valueof<typeof Constants.Environment>;
+    getVersion(): string;
+    setAppVersion(version: string): void;
+    setAppName(name: string): void;
+    startTrackingLocation(callback?: Callback): void;
+    stopTrackingLocation(): void;
+    logError(error: IErrorLogMessage | string, attrs?: any): void;
```
alexs-mparticle, line 100:
```suggestion
    logError(error: IErrorLogMessage | string, attrs?: SDKEventAttrs): void;
```

src/mp-instance.ts
```diff
@@ -16,31 +16,87 @@
...
+export interface IMParticleWebSDKInstance extends MParticleWebSDK {
+    _instanceName: string;
+    _preInit: IPreInit;
+    
+    _APIClient: IAPIClient;
+    _CookieSyncManager: ICookieSyncManager;
+    _Consent: IConsent;
+    _Ecommerce: IECommerce;
+    _Events: IEvents;
+    _Forwarders: any;
```
alexs-mparticle, line 76: Maybe make a sustained Eng ticket?

src/mp-instance.ts
```diff
@@ -571,10 +628,10 @@ export default function mParticleInstance(instanceName) {
 
         self._Events.logEvent(
             {
-                messageType: Types.MessageType.PageView,
+                messageType: MessageType.PageView,
                 name: eventName,
-                data: attrs,
-                eventType: Types.EventType.Unknown,
+                data: attrs as { [key: string]: string },
```
alexs-mparticle, line 633: Make this SDK Attributes

src/sdkRuntimeModels.ts
```diff
@@ -217,14 +199,62 @@ export interface MParticleWebSDK {
...
+    _isTestEnv?: boolean;
+
+    EventType: valueof<typeof EventType>;
+    CommerceEventType: valueof<typeof CommerceEventType>;
+    PromotionType: valueof<typeof PromotionActionType>;
+    ProductActionType: valueof<typeof ProductActionType>;
+
     isIOS?: boolean;
-    sessionManager: Pick<ISessionManager, 'getSession'>; // https://go.mparticle.com/work/SQDSDKS-6949
+    identifyRequest: IdentityApiData;
```
alexs-mparticle, line 234: This is only used for tests and should be removed from interface. We may need to refactor tests. 😢

src/sdkRuntimeModels.ts
```diff
@@ -217,14 +199,62 @@ export interface MParticleWebSDK {
...
+
+    Store: IStore;
+    MPSideloadedKit: IMPSideloadedKit;
+
+    configurePixel(config: IPixelConfiguration): void;
+    getEnvironment(): valueof<typeof Constants.Environment>;
+    getInstance(instanceName?: string): IMParticleWebSDKInstance;
+    getVersion(): string;
+    isInitialized(): boolean;
```
alexs-mparticle, line 243: Move to core

src/serverModel.ts
```diff
@@ -110,7 +110,7 @@ export interface IProductV2DTO {
     ps: number;
     cc: string | number;
     tpa: number;
-    attrs: Record<string, unknown> | undefined;
+    attrs: Record<string, string> | null;
```
alexs-mparticle, line 113: Can we make this a type?

src/serverModel.ts
```diff
@@ -181,7 +181,7 @@ function convertProductToV2DTO(product: SDKProduct): IProductV2DTO {
         ps: parseNumber(product.Position),
         cc: parseStringOrNumber(product.CouponCode),
         tpa: parseNumber(product.TotalAmount),
-        attrs: product.Attributes,
+        attrs: product.Attributes as Record<string, string> | null,
```
alexs-mparticle, line 184: Can we make this a type?

src/configAPIClient.ts
```diff
@@ -113,7 +114,7 @@ export default function ConfigAPIClient(
     this: IConfigAPIClient,
     apiKey: string,
     config: SDKInitConfig,
-    mpInstance: MParticleWebSDK
+    mpInstance: IMParticleWebSDKInstance
```
rmi22186, line 117: on line 88, i see `mpInstance: MParticleWebSDK`.  Should this now be `mpInstance: IMParticleWebSDKInstance` also?  If I do a global search for `mpInstance: MParticleWebSDK` there are several other occurrences.  Should these all be changed to `mpInstance: IMParticleWebSDKInstance`?

src/pre-init-utils.ts
```diff
@@ -1,5 +1,16 @@
+import { IPixelConfiguration } from './cookieSyncManager';
+import { MPForwarder } from './forwarders.interfaces';
+import { IntegrationDelays } from './mp-instance';
 import { isEmpty, isFunction } from './utils';
 
+export interface IPreInit {
+    readyQueue: Function[];
```
rmi22186, line 7: [based on the usage below](https://github.com/mParticle/mparticle-web-sdk/pull/968/files#diff-da27b40568cc9938393028855aed1db2e8fc2b9e7bb803504e270241477d178cR19-R21), this can be an array of functions, or an array of arrays.  I believe the second half is when someone tries to call `mParticle.something` before `something` fully loads (we stub these in the snippet).  might take some more digging into, but this is what i recall.

src/sdkRuntimeModels.ts
```diff
@@ -161,51 +159,36 @@ export interface SDKProduct {
...
-    _preInit: any; // TODO: Set up API
-    _instances?: Dictionary<MParticleWebSDK>;
-    _isTestEnv?: boolean;
     _resetForTests(
         MPConfig?: SDKInitConfig,
         keepPersistence?: boolean,
-        instance?: MParticleWebSDK
+        instance?: IMParticleWebSDKInstance,
     ): void;
+    configurePixel(config: IPixelConfiguration): void;
```
rmi22186, line 177: note to self...i can't quite remember when this is ever called from `mParticle.configurePixel`.

src/sdkRuntimeModels.ts
```diff
@@ -161,51 +159,36 @@ export interface SDKProduct {
...
     endSession(): void;
-    identifyRequest: IdentityApiData;
     init(apiKey: string, config: SDKInitConfig, instanceName?: string): void;
     _getActiveForwarders(): MPForwarder[];
+    _getIntegrationDelays(): IntegrationDelays;
```
rmi22186, line 181: ditto to the above comment - this is probably just because we have a setter, and so added a getter as well, but i don't think it's used anywhere

src/sdkRuntimeModels.ts
```diff
@@ -217,14 +200,49 @@ export interface MParticleWebSDK {
...
-    logLevel: string;
-    ProductActionType: SDKProductActionType;
+    isInitialized(): boolean;
+    ProductActionType: valueof<typeof ProductActionType>;
+    ready(f: Function): void;
+    reset(instance: IMParticleWebSDKInstance): void;
+    setAppName(name: string): void;
+    setAppVersion(version: string): void;
+    setOptOut(isOptingOut: boolean): void;
+    startTrackingLocation(callback?: Callback): void;
```
rmi22186, line 216: Can you add a work item to audit callbacks and create interfaces for each of them?

src/sdkRuntimeModels.ts
```diff
@@ -217,14 +200,49 @@ export interface MParticleWebSDK {
...
+    stopTrackingLocation(): void;
     generateHash(value: string): string;
+    setIntegrationAttribute(integrationModuleId: number, attrs: IntegrationAttribute): void;
+    getIntegrationAttributes(integrationModuleId: number): IntegrationAttribute;
```
rmi22186, line 220: I think this should be `IntegrationAttributes`

src/sdkRuntimeModels.ts
```diff
@@ -271,6 +289,8 @@ export interface SDKInitConfig
 
     // https://go.mparticle.com/work/SQDSDKS-6460
     identityCallback?: IdentityCallback;
+
+    rq?: Function[];
```
rmi22186, line 293: same comment as the above regarding `readyQueue`

src/store.ts
```diff
@@ -121,10 +121,16 @@ function createSDKConfig(config: SDKInitConfig): SDKConfig {
 //       to TypeScript
 export type ServerSettings = Dictionary;
 export type SessionAttributes = Dictionary;
-export type IntegrationAttributes = Dictionary<Dictionary<string>>;
 
-type WrapperSDKTypes = 'flutter' | 'none';
-interface WrapperSDKInfo {
+// FIXME: DefinitelyTyped Batch definition defines this as a string but
```
rmi22186, line 125: I don't think this comment still applies based on my memory of us discussing this, right?.   we pass in a number but it gets set as a string bc JS is JS. the `number` portion is when we are using the method mParticle.setIntegrationAttribute(_moduleid_, 'foo')

test/src/tests-identities-attributes.ts
```diff
@@ -799,11 +800,12 @@ describe('identities and attributes', function() {
...
     it('should get cart products', function(done) {
-        const product1 = mParticle.eCommerce.createProduct('iPhone', 'SKU1', 1),
-            product2 = mParticle.eCommerce.createProduct('Android', 'SKU2', 1);
+        const product1: SDKProduct = mParticle.eCommerce.createProduct('iPhone', 'SKU1', 1, 1);
+        const product2: SDKProduct = mParticle.eCommerce.createProduct('Android', 'SKU2', 1, 1);
+
         waitForCondition(hasIdentifyReturned)
         .then(() =>  {
-        mParticle.eCommerce.Cart.add([product1, product2], false);
+        mParticle.eCommerce.Cart.add([product1, product2], null);
```
rmi22186, line 808: this shouldn't matter since it's deprecated, but i believe the 2nd argument shoudl be true/false bc it indicated if it should log the event or not.  but this test will be removed in the future...so nbd

rmi22186 (changes requested):
thanks for your patience!

src/sdkRuntimeModels.ts
alexs-mparticle: Not sure about that. Looking at the actual implementation, it returns a single `IntegrationAttribute`:

```
    this.getIntegrationAttributes = function(integrationId) {
        if (self._Store.integrationAttributes[integrationId]) {
            return self._Store.integrationAttributes[integrationId];
        } else {
            return {};
        }
    };
```

src/configAPIClient.ts
alexs-mparticle: I think I may have missed a few so I just did another pass to make sure.

src/sdkRuntimeModels.ts
alexs-mparticle: Since it's technically "private" or "internal" can we kill it?

rmi22186: i'd say no to killing just because it's publicly accessible...who knows if someone is actively using it :/

src/sdkRuntimeModels.ts
rmi22186: [Setting it sets an object](https://github.com/mparticle-integrations/mparticle-javascript-integration-adobe/blob/65cc56d8c1a2caa0a2fc412eefd93b7073211fbd/packages/AdobeServer/src/AdobeServerSideKit.js#L90-L94).  Note the `else` above, it is returning an empty object, as opposed to a `null` if it were just returning a single item.

test/src/tests-identity.ts
```diff
@@ -4799,7 +4799,7 @@ describe('identity', function() {
             waitForCondition(hasIdentifyReturned)
             .then(() => {
 
-            const product: SDKProduct = mParticle.eCommerce.createProduct(
+            const product = mParticle.eCommerce.createProduct(
```
rmi22186, line 4802: add Type Product like above

src/sdkRuntimeModels.ts
alexs-mparticle: Per our conversation, I am more correct than you. 😉

src/sdkRuntimeModels.ts
```diff
@@ -217,14 +200,52 @@ export interface MParticleWebSDK {
...
+    // https://go.mparticle.com/work/SQDSDKS-7063
+    startTrackingLocation(callback?: Callback): void;
+
+    stopTrackingLocation(): void;
     generateHash(value: string): string;
+    setIntegrationAttribute(integrationModuleId: number, attrs: IntegrationAttribute): void;
+    getIntegrationAttributes(integrationModuleId: number): IntegrationAttribute;
+}
+
+// https://go.mparticle.com/work/SQDSDKS-4805
```
rmi22186, line 226: not sure if this comment needs to be here.  not a big deal and can be removed in the next PR

src/mp-instance.ts
```diff
@@ -16,31 +16,76 @@
...
+    _IdentityAPIClient: typeof IdentityAPIClient;
+    _IntegrationCapture: IntegrationCapture;
+    _NativeSdkHelpers: INativeSdkHelpers;
+    _Persistence: IPersistence;
+    _SessionManager: ISessionManager;
+    _ServerModel: IServerModel;
+    _Store: IStore;
+    _instanceName: string;
+    _preInit: IPreInit;
+
```
rmi22186, line 87:
```suggestion
```

rmi22186 (approved):

PR #969, opened 2025-01-22
refactor: Migrate mParticle Instance Imports

alexs-mparticle:
## Instructions

 ## Summary
 - Revises imports of mP Instance interfaces

 ## Testing Plan


test/src/tests-kit-blocking.ts
```diff
@@ -617,26 +617,6 @@ describe('kit blocking', () => {
             })
         });
 
-        it('integration test - should not throw an error when unplanned user attributes are allowed and block.ua = true', function(done) {
```
alexs-mparticle, line 620: Why was this removed?

PR #973, opened 2025-02-04
refactor: Migrate mParticle Instance Manager to TypeScript

alexs-mparticle:
## Instructions

 ## Summary

 ## Testing Plan

### DIFF
--- a/src/mparticle-instance-manager.ts
+++ b/src/mparticle-instance-manager.ts
@@ -1,9 +1,13 @@
 import Polyfill from './polyfill';
-import Types from './types';
+import Types, { CommerceEventType, EventType, ProductActionType, PromotionActionType } from './types';
 import Constants from './constants';
-import mParticleInstance from './mp-instance.js';
+import mParticleInstance, { IMParticleWebSDKInstance } from './mp-instance.js';
 import _BatchValidator from './mockBatchCreator';
 import MPSideloadedKit from './sideloadedKit';
+import { IMParticleInstanceManager } from './sdkRuntimeModels';
+import { IStore } from './store';
+import { Dictionary } from '@mparticle/web-sdk';
+import { valueof } from './utils';
 
 if (!Array.prototype.forEach) {
     Array.prototype.forEach = Polyfill.forEach;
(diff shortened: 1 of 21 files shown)


src/mparticle-instance-manager.ts
```diff
@@ -17,21 +21,26 @@ if (!Array.prototype.filter) {
...
 
+// https://go.mparticle.com/work/SQDSDKS-6768
 if (!Array.isArray) {
+    // @ts-ignore
     Array.prototype.isArray = Polyfill.isArray;
 }
 
-function mParticle() {
-    var self = this;
+// Question: Should we just change this to mParticleInsanceManager?
```
rmi22186, line 30: I think that's fine.  we end up instantiating it below, so this is the correct name
```suggestion
// Question: Should we just change this to mParticleInstanceManager?
```

src/mparticle-instance-manager.ts
```diff
@@ -17,21 +21,26 @@ if (!Array.prototype.filter) {
...
+function mParticle(this: IMParticleInstanceManager, instanceName?: string) {
```
rmi22186, line 31: i don't think this needs an `instanceName` as an argument

src/mparticle-instance-manager.ts
```diff
@@ -17,21 +21,26 @@ if (!Array.prototype.filter) {
...
-    this.PromotionType = Types.PromotionActionType;
-    this.ProductActionType = Types.ProductActionType;
-    this.MPSideloadedKit = MPSideloadedKit;
+    this.EventType = EventType as unknown as valueof<typeof EventType>;
+    this.CommerceEventType = CommerceEventType as unknown as valueof<typeof CommerceEventType>;
+    this.PromotionType = PromotionActionType as unknown as valueof<typeof PromotionActionType>;
+    this.ProductActionType = ProductActionType as unknown as valueof<typeof ProductActionType>;
+
+    // QUESTION:  Should this be a 'new' instance?
+    this.MPSideloadedKit =  MPSideloadedKit as unknown as valueof<typeof MPSideloadedKit>;
```
rmi22186, line 43: can you make a jira for this? i'd have to spend a bit of time (that i don't have right now) sorting through this one.

rmi22186: Nope.  I also don't think it should be a valueof and just remain MPSideloadedKit

take a look at these docs in terms of how a customer uses it:
https://docs.mparticle.com/developers/client-sdks/web/kits/#including-the-sideloaded-kit

src/mparticle-instance-manager.ts
```diff
@@ -480,11 +489,14 @@ function mParticle() {
     };
 }
 
-var mparticleInstance = new mParticle();
+// Can we get rid of this and just rename the class to mparticleInstanceManager?
+const mparticleInstance = new mParticle();
```
rmi22186, line 493: i don't know about getting rid of it, but renaming it per my agreement above would be fine.

test/src/tests-batchUploader_2.ts
```diff
@@ -3,8 +3,8 @@ import Utils from './config/utils';
 import { expect } from 'chai';
 import _BatchValidator from '../../src/mockBatchCreator';
 import fetchMock from 'fetch-mock/esm/client';
-import { IMParticleInstanceManager } from '../../src/mparticle-instance-manager';
 import { CustomEvent } from '@mparticle/event-models';
+import { IMParticleInstanceManager } from '../../src/sdkRuntimeModels';
```
rmi22186, line 7: nit: very minor, but move back to line 6 to reduce line changes?  same across a few of the below files

rmi22186 (changes requested):

src/mparticle-instance-manager.ts
```diff
@@ -480,11 +489,14 @@ function mParticle() {
...
 
 if (typeof window !== 'undefined') {
     window.mParticle = mparticleInstance;
+
```
alexs-mparticle, line 497: Note to self, create a comment explaining why we call this `window.mParticle`.

src/mparticle-instance-manager.ts
```diff
@@ -480,11 +487,15 @@ function mParticle() {
     };
 }
 
-var mparticleInstance = new mParticle();
+const mParticleManager = new mParticleInstanceManager();
 
 if (typeof window !== 'undefined') {
-    window.mParticle = mparticleInstance;
+    // mParticle is the global object used to access the SDK and predates instance manager
+    // it is now used to access both the default instance and the instance manager.
```
rmi22186, line 494:
```suggestion
    // mParticle is the global object used to access the SDK and predates instance manager,
    // when mParticle was a singleton.  We now support multiple instances. Calling methods
    // on mParticle directly will access the default instance, but mParticle can also be used 
    // as the instance manager in self hosted mode.
```

rmi22186 (approved):

PR #974, opened 2025-02-04
fix: Update new mParticleInstance instantiation

alexs-mparticle:
## Instructions

 ## Summary
 - Found a bug where creating a new instance of the web sdk will use the `apiKey` as the instance name rather than the actual instance name. This is because the constructor signature is not being used correctly.

 ## Testing Plan
 - Using a sample app, create multiple instances of the web sdk and verify that the instance names are strings and not api keys.

 - Closes https://go.mparticle.com/work/SQDSDKS-7067


rmi22186 (approved):

rmi22186 (changes requested):
this could actually use a test somewhere where you are creating an instance of `foo-instance`, then check that the `._instances` has the key `foo-instance` instead of the apikey.

src/sdkRuntimeModels.ts
```diff
@@ -163,6 +163,7 @@ export interface SDKProduct {
 
 export interface MParticleWebSDK {
     addForwarder(mockForwarder: MPForwarder): void;
+    _instanceName: string;
```
alexs-mparticle, line 166: This will be part of #968

rmi22186: Can you clarify wha tthis means?  I don't see it in that PR

rmi22186: Oh, I think I get it.  Basically, it will also be included in the TS migration but under IMParticleWebSDKInstance, right?  Just didn't see it yet.

rmi22186 (approved):
LGTM!

PR #976, opened 2025-02-10
refactor: Migrate Forwarder Tests to TypeScript

alexs-mparticle:
## Instructions

 ## Summary
 - Migrates Forwarder Tests to TypeScript
 - Verifies imported interfaces and types to ensure they are properly implemented.

 ## Testing Plan
 - Verify automated tests pass and the repo builds properly
 - Manually confirm Identity and Event Types are accessible via the `mParticle` object in the dev console.

 - Closes https://go.mparticle.com/work/SQDSDKS-5767

### DIFF
--- a/src/mp-instance.ts
+++ b/src/mp-instance.ts
@@ -127,10 +127,10 @@ export default function mParticleInstance(this: IMParticleWebSDKInstance, instan
 
     // required for forwarders once they reference the mparticle instance
     this.IdentityType = IdentityType;
-    this.EventType = EventType as unknown as valueof<typeof EventType>;
-    this.CommerceEventType = CommerceEventType as unknown as valueof<typeof CommerceEventType>;
-    this.PromotionType = PromotionActionType as unknown as valueof<typeof PromotionActionType>;
-    this.ProductActionType = ProductActionType as unknown as valueof<typeof ProductActionType>;
+    this.EventType = EventType;
+    this.CommerceEventType = CommerceEventType;
+    this.PromotionType = PromotionActionType;
+    this.ProductActionType = ProductActionType;
 
 
     this._Identity = new Identity(this);
--- a/test/src/tests-forwarders.ts
+++ b/test/src/tests-forwarders.ts
@@ -10,8 +10,16 @@ import {
     MessageType,
 } from './config/constants';
 import { expect } from 'chai';
+import { IMParticleInstanceManager, SDKEvent, SDKInitConfig } from '../../src/sdkRuntimeModels';
+import { IMParticleUser, UserAttributes } from '../../src/identity-user-interfaces';
+import { IdentityType } from '../../src/types';
+import { IntegrationAttribute } from '../../src/store';
+import { IConsentRules } from '../../src/consent';
+import { UserIdentities } from '@mparticle/web-sdk';
 
-const { findEventFromRequest,
+
+const {
+    findEventFromRequest,
     waitForCondition,
     fetchMockSuccess,
     getForwarderEvent,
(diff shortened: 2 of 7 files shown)


src/mp-instance.ts
```diff
@@ -126,11 +126,12 @@ export default function mParticleInstance(this: IMParticleWebSDKInstance, instan
     this._IntegrationCapture = new IntegrationCapture();
 
     // required for forwarders once they reference the mparticle instance
-    this.IdentityType = IdentityType;
-    this.EventType = EventType as unknown as valueof<typeof EventType>;
+    this.IdentityType = IdentityType as unknown as valueof<typeof IdentityType>;
+    // this.EventType = EventType as unknown as valueof<typeof EventType>;
+    this.EventType = EventType;
```
rmi22186, line 131: Which one works here?

src/mp-instance.ts
```diff
@@ -126,11 +126,12 @@ export default function mParticleInstance(this: IMParticleWebSDKInstance, instan
...
     this.CommerceEventType = CommerceEventType as unknown as valueof<typeof CommerceEventType>;
     this.PromotionType = PromotionActionType as unknown as valueof<typeof PromotionActionType>;
-    this.ProductActionType = ProductActionType as unknown as valueof<typeof ProductActionType>;
+    this.ProductActionType = ProductActionType;
```
rmi22186, line 134: similarly - did we not need the `as unknown as valueof<...>` for all of these after all?

rmi22186 (changes requested):
Looking good so far...

test/src/tests-forwarders.ts
```diff
@@ -2084,7 +2084,7 @@ describe('forwarders', function() {
         window.mParticle.config.kitConfigs.push(config1);
 
         mParticle.init(apiKey, window.mParticle.config);
-        await waitForCondition(hasIdentityCallInflightReturned);
+        await waitForCondition(() => window.mParticle.getInstance()?._Store?.identityCallInFlight === false);
```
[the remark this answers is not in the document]
alexs-mparticle: I think this a mistake on my part.

test/src/tests-forwarders.ts
```diff
@@ -3469,11 +3385,12 @@ describe('forwarders', function() {
                 .then(() => {
                 const mpInstance = window.mParticle.getInstance();
 
-                expect(mpInstance._Store.isUsingSideloadedKits).to.be.undefined;
+                // QUESTION: Where is this coming from?
+                // expect(mpInstance._Store.isUsingSideloadedKits).to.be.undefined;
```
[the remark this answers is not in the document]
alexs-mparticle: IIRC I added `sideloadedKitsCount` when we were trying to add metrics around side loaded kits. Not sure if that helps add context.

I say we probably should just remove this line since it's not a real method.

src/mp-instance.ts
alexs-mparticle: I noticed that when actually using the assigned values with `unknown as valueof`, it was treating them as `any`. This way, in the tests, we can validate the actual values.

test/src/tests-forwarders.ts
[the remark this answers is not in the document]
rmi22186: yea i agree with removing

test/src/tests-forwarders.ts
```diff
@@ -1253,22 +1251,18 @@ describe.only('forwarders', function() {
...
         event.should.have.property('UserAttributes');
         event.UserAttributes.should.not.have.property('gender');
-
-        done();
-        });
     });
 
-    // FIXME: Passes in isolation, but fails when run with other tests
-    it.skip('should filter user identities from forwarder on init and bring customerid as first ID', async () => {
+    it('should filter user identities from forwarder on init and bring customerid as first ID', async () => {
```
rmi22186, line 1265: so this passes now? 🤞

rmi22186 (changes requested):
one question otherwise looks great

rmi22186 (approved):

PR #981, opened 2025-02-12
fix: Support numbers in vault

rmi22186:
## Instructions

 ## Summary
Allow the vault to save numbers properly.

 ## Testing Plan
Unit tests added

 - Closes https://go.mparticle.com/work/SQDSDKS-7074

### DIFF
--- a/src/vault.ts
+++ b/src/vault.ts
@@ -1,5 +1,5 @@
 import { Logger } from '@mparticle/web-sdk';
-import { isEmpty } from './utils';
+import { isEmpty, isNumber } from './utils';
 
 export interface IVaultOptions {
     logger?: Logger;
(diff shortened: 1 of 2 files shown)


test/src/vault.spec.ts
```diff
@@ -168,6 +201,36 @@ describe('Vault', () => {
                     JSON.stringify(testArray)
                 );
             });
+
+            it('should store a string', () => {
```
alexs-mparticle, line 205: Please write test cases for retrieving a string and number as well.

src/vault.ts
```diff
@@ -42,9 +42,14 @@ export abstract class BaseVault<StorableItem> {
...
     public store(item: StorableItem): void {
+        let stringifiedItem: string;
         this.contents = item;
 
-        const stringifiedItem = !isEmpty(item) ? JSON.stringify(item) : '';
+        if (isNumber(item)) {
+            stringifiedItem = JSON.stringify(item);
+        } else {
+            stringifiedItem = !isEmpty(item) ? JSON.stringify(item) : '';
+        }
```
alexs-mparticle, line 52: I _think_ this might be cleaner and more readable, but I'm not adamant about it.

```suggestion
        if (isNumber(item) || !isEmpty(item)) {
            stringifiedItem = JSON.stringify(item);
        } else {
            stringifiedItem = '';
        }
```

alexs-mparticle (changes requested):

alexs-mparticle (approved):

PR #983, opened 2025-02-13
fix: Add failsafes for queryStringParser

alexs-mparticle:
## Instructions

 ## Summary
 - Adds handlers to catch potentially malformed urls in `queryStringParser` fallback

 ## Testing Plan
 - Run automated tests
 - Set up sample app and pass in malformed strings into url as query parameter

 - Closes https://go.mparticle.com/work/SQDSDKS-6989

### DIFF
--- a/test/jest/utils.spec.ts
+++ b/test/jest/utils.spec.ts
@@ -148,7 +148,7 @@ describe('Utils', () => {
             });
 
             it('should handle non-standard characters or malformed urls', () => {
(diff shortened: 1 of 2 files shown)


test/jest/utils.spec.ts
```diff
@@ -146,6 +146,27 @@ describe('Utils', () => {
 
                 expect(queryStringParser(url, [])).toEqual(expectedResult);
             });
+
+            it('should handle non-standard characters or malformed urls', () => {
```
rmi22186, line 150: I ran this test against the previous code in src/utils and it passes.  I would expect the sample to not pass, but maybe I'm doing something wrong?

rmi22186: using `a%AFc` as a value should cause the previous code to fail, but current code to pass.

rmi22186 (approved):

PR #985, opened 2025-02-17
fix: Add Rokt Click ID for Integration Capture

alexs-mparticle:
## Instructions

 ## Summary
 - Adds Rokt Click Id to Integration Capture as a Custom Flag
 - https://docs.rokt.com/developers/integration-guides/web/advanced/rokt-id-tag/#test-the-rokt-id

 ## Testing Plan
 - Run automated tests
 - Use a sample app to verify capture of Rokt Click Id via url parameter

 - Closes https://go.mparticle.com/work/SQDSDKS-7078

### DIFF
--- a/src/integrationCapture.ts
+++ b/src/integrationCapture.ts
@@ -99,6 +99,12 @@ const integrationMapping: IntegrationIdMapping = {
         output: IntegrationOutputs.CUSTOM_FLAGS,
     },
 
+    // Rokt
+    rtid: {
+        mappedKey: 'Rokt.ClickId',
+        output: IntegrationOutputs.CUSTOM_FLAGS,
+    },
+
     // TIKTOK
     ttclid: {
         mappedKey: 'TikTok.Callback',
--- a/src/integrationCapture.ts
+++ b/src/integrationCapture.ts
@@ -105,12 +105,12 @@ const integrationMapping: IntegrationIdMapping = {
 
     // Rokt
     rtid: {
-        mappedKey: 'rokt_id',
+        mappedKey: 'passbackconversiontrackingid',
         output: IntegrationOutputs.INTEGRATION_ATTRIBUTES,
         moduleId: 1277,
     },
     RoktTransactionId: {
-        mappedKey: 'rokt_id',
+        mappedKey: 'passbackconversiontrackingid',
         output: IntegrationOutputs.INTEGRATION_ATTRIBUTES,
         moduleId: 1277,
     },
--- a/src/integrationCapture.ts
+++ b/src/integrationCapture.ts
@@ -1,4 +1,5 @@
 import { SDKEventCustomFlags } from './sdkRuntimeModels';
+import { IntegrationAttributes } from './store';
 import {
     Dictionary,
     queryStringParser,
@@ -53,16 +54,19 @@ export const facebookClickIdProcessor: IntegrationCaptureProcessorFunction = (
 // Integration outputs are used to determine how click ids are used within the SDK
 // CUSTOM_FLAGS are sent out when an Event is created via ServerModel.createEventObject
 // PARTNER_IDENTITIES are sent out in a Batch when a group of events are converted to a Batch
+// INTEGRATION_ATTRIBUTES are stored on the event level but are also passed up to the Batch
 
 const IntegrationOutputs = {
     CUSTOM_FLAGS: 'custom_flags',
     PARTNER_IDENTITIES: 'partner_identities',
+    INTEGRATION_ATTRIBUTES: 'integration_attributes',
 } as const;
 
 interface IntegrationMappingItem {
     mappedKey: string;
     output: valueof<typeof IntegrationOutputs>;
     processor?: IntegrationCaptureProcessorFunction;
+    moduleId?: number;
 }
 
 interface IntegrationIdMapping {
@@ -101,8 +105,9 @@ const integrationMapping: IntegrationIdMapping = {
 
     // Rokt
     rtid: {
-        mappedKey: 'Rokt.ClickId',
-        output: IntegrationOutputs.CUSTOM_FLAGS,
+        mappedKey: 'rokt_id',
+        output: IntegrationOutputs.INTEGRATION_ATTRIBUTES,
+        moduleId: 1277,
     },
 
     // TIKTOK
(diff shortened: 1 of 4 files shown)


test/src/tests-integration-capture.ts
```diff
@@ -71,13 +72,11 @@ describe('Integration Capture', () => {
...
-            'Facebook.BrowserId': '54321',
-            'GoogleEnhancedConversions.Gclid': '234',
-            'GoogleEnhancedConversions.Gbraid': '6574',
-            'GoogleEnhancedConversions.Wbraid': '1234111',
-        });
+        expect(testEvent.data.custom_flags['Facebook.ClickId'], 'Facebook Click Id').to.equal(`fb.1.${initialTimestamp}.1234`);
+        expect(testEvent.data.custom_flags['Facebook.BrowserId'], 'Facebook Browser Id').to.equal('54321');
+        expect(testEvent.data.custom_flags['GoogleEnhancedConversions.Gclid'], 'Google Enhanced Conversions Gclid').to.equal('234');
+        expect(testEvent.data.custom_flags['GoogleEnhancedConversions.Gbraid'], 'Google Enhanced Conversions Gbraid').to.equal('6574');
+        expect(testEvent.data.custom_flags['GoogleEnhancedConversions.Wbraid'], 'Google Enhanced Conversions Wbraid').to.equal('1234111');
```
rmi22186, line 79: any reason for the formatting change?  should you add roktid to this

src/integrationCapture.ts
```diff
@@ -180,6 +193,37 @@ export default class IntegrationCapture {
...
+                if (!isEmpty(mappingKey)) {
+                    const moduleId = this.filteredIntegrationAttributeMappings[key]?.moduleId;
+                    if (moduleId && !mappedClickIds[moduleId]) {
+                        mappedClickIds[moduleId] = { [mappingKey]: value };
+                    }
+                }
+            }
+        }
+        return mappedClickIds;
+    }
```
rmi22186, line 224: is there any reason you can't use getClickIds and pass in the `filteredIntegrationAttributeMappings` as the 2nd argument like you do in lines 176-192?

rmi22186 (changes requested):

test/src/tests-integration-capture.ts
alexs-mparticle: Not exactly. This is to make it easier to diagnose failing tests when a specific click ID is broken or missing. I realized when we were using an object comparison, Karma/Chai wasn't really reporting which specific click ID was missing.

src/integrationCapture.ts
alexs-mparticle: The difference is the shape of the data. For example, `getClickIds` returns a simple dictionary and therefore can be made into a simple utility function:
```
{
    clickID1: 'value',
    clickID2: 'value',
    clickID3: 'value',
}
```

But this is very specific to integration attributes, and is more of a hash map:
```
{
    "<moduleId>": {
        "clickId": "clickIdValue"
    }
}
```

I could potentially move this into a separate utility function, but it may not be necessary since we're only using it in one place. I created the previous utility function because it was being used in two different places.

src/integrationCapture.ts
```diff
@@ -169,6 +189,21 @@ export default class IntegrationCapture {
...
+    public captureLocalStorage(): Dictionary<string> {
+        let localStorageItems: Dictionary<string> = {};
+        for (const key in integrationMapping) {
+            const localStorageItem = localStorage.getItem(key);
+            if (localStorageItem) {
+                localStorageItems[key] = localStorageItem;
+            }
+        }
+
+        return this.applyProcessors(localStorageItems, getHref(), this.initialTimestamp);
```
rmi22186, line 204: any reason we aren't including getHref(), and initialTimestamp in `captureCookies`?

src/integrationCapture.ts
```diff
@@ -53,16 +54,19 @@ export const facebookClickIdProcessor: IntegrationCaptureProcessorFunction = (
 // Integration outputs are used to determine how click ids are used within the SDK
 // CUSTOM_FLAGS are sent out when an Event is created via ServerModel.createEventObject
 // PARTNER_IDENTITIES are sent out in a Batch when a group of events are converted to a Batch
+// INTEGRATION_ATTRIBUTES are stored on the event level but are also passed up to the Batch
```
rmi22186, line 57: can you clarify more what "stored on the event level" means?  It's not stored on events at all, but passed on the batch as you say in the end of the comment

src/integrationCapture.ts
```diff
@@ -130,14 +148,29 @@ export default class IntegrationCapture {
...
         // Exclude _fbc if fbclid is present
         // https://developers.facebook.com/docs/marketing-api/conversions-api/parameters/fbp-and-fbc#retrieve-from-fbclid-url-query-parameter
         if (queryParams['fbclid'] && cookies['_fbc']) {
             delete cookies['_fbc'];
         }
 
-        this.clickIds = { ...this.clickIds, ...queryParams, ...cookies };
+        // If both rtid and RoktTransactionId are present, prioritize rtid
+        // If RoktTransactionId is present in both cookies and localStorage,
+        // prioritize localStorage
```
rmi22186, line 161: I can't remember if you said there were actually docs for this, or if this was shared internally from Ben.  If there are docs, I'd link similar to line 153-154 above

test/jest/integration-capture.spec.ts
```diff
@@ -191,6 +207,112 @@ describe('Integration Capture', () => {
...
+
+                localStorage.setItem('RoktTransactionId', '54321');
+
+                const integrationCapture = new IntegrationCapture();
+                integrationCapture.capture();
+
+                expect(integrationCapture.clickIds).toEqual({
+                    RoktTransactionId: '54321',
+                });
+
```
rmi22186, line 257:
```suggestion
```

src/integrationCapture.ts
```diff
@@ -99,6 +103,18 @@ const integrationMapping: IntegrationIdMapping = {
         output: IntegrationOutputs.CUSTOM_FLAGS,
     },
 
+    // Rokt
+    rtid: {
+        mappedKey: 'passbackconversiontrackingid',
```
rmi22186, line 108: is this key documented anywhere?

rmi22186 (changes requested):
mostly nits.

src/integrationCapture.ts
rmi22186, line 57:
```suggestion
// INTEGRATION_ATTRIBUTES are stored initially on the SDKEvent level but then is added to the Batch when the batch is created
```

rmi22186 (approved):

rmi22186 (approved):

PR #987, opened 2025-02-24
fix: Update TOS local storage key

rmi22186:
## Instructions

 ## Summary
Updating the key from `mp-time` to `mprtcl-tos` to be consistent with other localstorage prefixes:

image

 ## Testing Plan
Updated unit tests and tested in a test app to get the above screenshot

 - Closes https://go.mparticle.com/work/SQDSDKS-7105

### DIFF
--- a/src/foregroundTimeTracker.ts
+++ b/src/foregroundTimeTracker.ts
@@ -9,7 +9,7 @@ export default class ForegroundTimeTracker {
     public totalTime: number = 0;
 
     constructor(timerKey: string) {
-        this.localStorageName = `mp-time-${timerKey}`;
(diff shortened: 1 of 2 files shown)


src/foregroundTimeTracker.ts
```diff
@@ -9,7 +9,7 @@ export default class ForegroundTimeTracker {
     public totalTime: number = 0;
 
     constructor(timerKey: string) {
-        this.localStorageName = `mp-time-${timerKey}`;
+        this.localStorageName = `mprtcl-tos-${timerKey}`;
```
alexs-mparticle, line 12: Let's create a work item to create a local storage prefix constant or generator.

alexs-mparticle (approved):

PR #991, opened 2025-03-11
feat: Fire background AST events on page transitions

rmi22186:
## Instructions

 ## Summary
We want to capture TOS when a user closes a window/moves tabs.  Currently if a user comes to a page and do not perform any mParticle actions, (but does scroll up and down and reads a page), then closes the window, no events are fired for TOS to be captured.  Adding AST background events upon page resolves this.  It is added to the end of batch.

This PR is incomplete, but 90% of the way there currently, so I am putting it in draft.  This will be feature flagged.

 ## Testing Plan
Locally tested with test app and integration tests.  Some tests are still currently failing due to using karma...I haven't tried jest yet but that's my next step.

 - Closes https://go.mparticle.com/work/SQDSDKS-7118


src/batchUploader.ts
```diff
@@ -129,21 +131,68 @@ export class BatchUploader {
...
+        if (now - this.lastASTEventTime < this.AST_DEBOUNCE_MS) {
+            return true;
+        }
+
+        this.lastASTEventTime = now;
+
+        return false;
+    }
+
+    private createBackgroundASTEvent(): SDKEvent {
```
rmi22186, line 145: I will DRY this and the code in [servermodels](https://github.com/mParticle/mparticle-web-sdk/blob/master/src/serverModel.ts#L340-L390) up, but wanted to get a POC going first.

alexs-mparticle: Make a sustained eng ticket please.

src/sdkToEventsApiConverter.ts
```diff
@@ -489,11 +489,14 @@ export function convertAST(
     const commonEventData: EventsApi.CommonEventData = convertBaseEventData(
         sdkEvent
     );
+
+    // Determine the transition type based on IsBackgroundAST flag
+    const transitionType = sdkEvent.IsBackgroundAST
+        ? EventsApi.ApplicationStateTransitionEventDataApplicationTransitionTypeEnum.applicationBackground
+        : EventsApi.ApplicationStateTransitionEventDataApplicationTransitionTypeEnum.applicationInitialized;
+
```
alexs-mparticle, line 497:
```suggestion
    const { applicationBackground, applicationInitialized } = EventsApi.ApplicationStateTransitionEventDataApplicationTransitionTypeEnum;
    const transitionType = sdkEvent.IsBackgroundAST ? applicationBackground : applicationInitialized;

```

test/src/tests-batchUploader.ts
```diff
@@ -394,6 +395,263 @@ describe('batch uploader', () => {
...
+                    console.log(lastEvent)
+                    expect(lastEvent.event_type).to.equal('application_state_transition');
+                    expect(lastEvent.data.application_transition_type).to.equal('application_background');
+
+                    // Clean up
+                    beaconSpy.restore();
+                    clock.restore();
+                });
+
+                // TODO: this test is failing right now.  there doesn't seem to be a way for karma to deal with beforeunload events properly
```
alexs-mparticle, line 472: Can we just get rid of it or is this a sustained eng project?

src/batchUploader.ts
```diff
@@ -129,21 +131,68 @@ export class BatchUploader {
...
+            UserAttributes: {},
+            UserIdentities: [],
+            SessionStartDate: _Store.sessionStartDate?.getTime() || now,
+            Debug: _Store.SDKConfig.isDevelopmentMode,
+            CurrencyCode: null,
+            ExpandedEventCount: 0,
+            ActiveTimeOnSite: this.mpInstance._timeOnSiteTimer?.getTimeInForeground() || 0,
+            IsBackgroundAST: true
+        };
+    }
```
alexs-mparticle, line 169:
```suggestion
        const { _Store, Identity, _timeOnSiteTimer  } = this.mpInstance;
        const { sessionId, deviceId, sessionStartDate, SDKConfig } = _Store;
        return {
            EventDataType: MessageType.AppStateTransition,
            EventName: 'Application Exit',
            EventCategory: EventType.Other,
            Timestamp: now,
            SessionId: sessionId,
            MPID: Identity.getCurrentUser()?.getMPID(),
            DeviceId: deviceId,
            IsFirstRun: false,
            SourceMessageId: '',
            SDKVersion: Constants.sdkVersion,
            CustomFlags: {},
            UserAttributes: {},
            UserIdentities: [],
            SessionStartDate: sessionStartDate?.getTime() || now,
            Debug: SDKConfig.isDevelopmentMode,
            CurrencyCode: null,
            ExpandedEventCount: 0,
            ActiveTimeOnSite: _timeOnSiteTimer?.getTimeInForeground() || 0,
            IsBackgroundAST: true
        };
    }
```

src/batchUploader.ts
```diff
@@ -129,21 +131,68 @@ export class BatchUploader {
         return offlineStoragePercentage >= rampNumber;
     }
 
+    private shouldDebounceAST(): boolean {
+        const now = Date.now();
+        if (now - this.lastASTEventTime < this.AST_DEBOUNCE_MS) {
+            return true;
+        }
+
+        this.lastASTEventTime = now;
```
alexs-mparticle, line 140: At first glance like it's just doing a check but it's actually doing a side effect. I personally would modify `lastASTEventTime` outside of the function. If you need this side effect in the function, I would rename the function to `checkLastDebounceASTAndUpdate()` or something similar.

src/batchUploader.ts
```diff
@@ -129,21 +131,68 @@ export class BatchUploader {
...
     // Adds listeners to be used trigger Navigator.sendBeacon if the browser
     // loses focus for any reason, such as closing browser tab or minimizing window
     private addEventListeners() {
         const _this = this;
 
+        const handleExit = () => {
+            // Check for debounce before creating and queueing event
+            if (_this.shouldDebounceAST()) {
+                return;
+            }
```
alexs-mparticle, line 180: Based on my previous suggestion, to avoid the side effect, I would do this:

```suggestion
            if (_this.shouldDebounceAST()) {
                this.lastASTEventTime = new Date.now();
                return;
            }
```

src/batchUploader.ts
```diff
@@ -129,21 +131,68 @@ export class BatchUploader {
...
         // visibility change is a document property, not window
         document.addEventListener('visibilitychange', () => {
-            _this.prepareAndUpload(false, _this.isBeaconAvailable());
-        });
-        window.addEventListener('beforeunload', () => {
-            _this.prepareAndUpload(false, _this.isBeaconAvailable());
-        });
-        window.addEventListener('pagehide', () => {
-            _this.prepareAndUpload(false, _this.isBeaconAvailable());
+            if (document.visibilityState === 'hidden') {
```
alexs-mparticle, line 190: Nitpick: I get the appeal to reduce the footprint of this listener, but since `handleExit()` isn't being reused at the moment, I don't see a point in making it a convenience method. It also serves to highlight that this specific listener is doing something different and prevents the need to dig too deep for what's going on.

alexs-mparticle (changes requested):

src/batchUploader.ts
rmi22186: i'm not sure i understand this comment.  handleExist is being used 3x in this file.  so that's why it should be its own method to be DRY

alexs-mparticle: Disregard this comment. I didn't realize at the time it was being used in multiple places.

src/batchUploader.ts
rmi22186: looks like i actually do need this here ebcause if i try to call `shouldDebounceAST` in a unit test without it, it won't have the lastASTEventTime and it will always return `false`.  so i'll add this back and rename

src/batchUploader.ts
```diff
@@ -129,21 +130,70 @@ export class BatchUploader {
...
+        return {
+            EventDataType: MessageType.AppStateTransition,
+            EventName: 'Application Exit',
+            EventCategory: EventType.Other,
+            Timestamp: now,
+            SessionId: sessionId,
+            MPID: Identity.getCurrentUser()?.getMPID(),
+            DeviceId: deviceId,
+            IsFirstRun: false,
+            SourceMessageId: '',
```
alexs-mparticle, line 158: I think this should use `utils.generateUniqueId` to generate an ID.

src/event-creator.ts
```diff
@@ -0,0 +1,34 @@
...
+    return {
+        EventDataType: messageType,
+        EventName: eventName,
+        EventCategory: EventType.Other,
+        Timestamp: now,
+        SessionId: sessionId,
+        MPID: Identity.getCurrentUser()?.getMPID(),
+        DeviceId: deviceId,
+        IsFirstRun: false,
+        SourceMessageId: '',
```
alexs-mparticle, line 23: Same here. Use `utils.generateUniqueId`

test/jest/batchUploader.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+    });
+
+    afterEach(() => {
+        jest.useRealTimers();
+    });
+
+    describe('shouldDebounceAST', () => {
+        it('should return false for first call', () => {
+            // First call should not be debounced
+            const firstCall = (batchUploader as any).shouldDebounceAndUpdateLastASTTime();
```
alexs-mparticle, line 43: Are you setting `batchUploader as any` because the method is private? Does this really need to be private?

If we set this as `any`, I think we lose the benefit of type safety on this object in the tests. I think a cleaner approach would be to the function and use the proper return types.

```
const shouldDebounceAndUpdateLastASTTime: boolean  = (batchUploader as any).shouldDebounceAndUpdateLastASTTime;
```

This would likely clean up your tests and reduce the need to repeat `batchUploader as any`.

If that doesn't work, I would consider creating a `mockBatchUploader`:

```
const mockBatchUploader = batchUploader as any;
``` 

...similar to how we've done workarounds for type hints for invalid values.

alexs-mparticle (changes requested):

src/event-creator.ts
rmi22186: this file actually shouldn't be here.  removing.

src/batchUploader.ts
rmi22186: good catch

src/user-utils.ts
```diff
@@ -23,3 +26,49 @@ export function hasMPIDChanged(
...
+    if (event.MPID && event.MPID === user.getMPID()) {
+        return;
+    }
+
+    event.MPID = user.getMPID();
+    event.ConsentState = user.getConsentState();
+    event.UserAttributes = user.getAllUserAttributes();
+
+    const userIdentities = user.getUserIdentities().userIdentities;
+    const dtoUserIdentities = {};
```
alexs-mparticle, line 54: Please assign the necessary types.

src/user-utils.ts
```diff
@@ -23,3 +26,49 @@ export function hasMPIDChanged(
...
+    for (const identityKey in userIdentities) {
+        const identityType = Types.IdentityType.getIdentityType(identityKey);
+        if (identityType !== false) {
+            dtoUserIdentities[identityType] = userIdentities[identityKey];
+        }
+    }
+
+    const validUserIdentities = [];
```
alexs-mparticle, line 62: Same here. I would like us to use type declarations for these empty objects.

alexs-mparticle (changes requested):

src/batchUploader.ts
```diff
@@ -129,21 +131,80 @@ export class BatchUploader {
...
+            CustomFlags: {},
+            EventAttributes: {},
+            SessionStartDate: sessionStartDate?.getTime() || now,
+            Debug: SDKConfig.isDevelopmentMode,
+            ActiveTimeOnSite: _timeOnSiteTimer?.getTimeInForeground() || 0,
+            IsBackgroundAST: true
+        };
+
+        appendUserInfo(getCurrentUser(), event as SDKEvent);
+        return event as SDKEvent;
```
alexs-mparticle, line 170:
```suggestion
        } as SDKEvent;

        appendUserInfo(getCurrentUser(), event);
        return event;
```

src/batchUploader.ts
```diff
@@ -129,21 +131,80 @@ export class BatchUploader {
...
+            const {
+                _Helpers: { getFeatureFlag },
+            } = this.mpInstance;
+
+            const flagValue = getFeatureFlag(Constants.FeatureFlags.AstBackgroundEvents);
+            if (flagValue) {
+                console.log('Feature flag value:', flagValue, 'type:', typeof flagValue);
+            }
+
+            if (getFeatureFlag(Constants.FeatureFlags.AstBackgroundEvents)) {
```
alexs-mparticle, line 189: I would destructure `Constants` and `FeatureFlags` at the top of the file, so this becomes more readable.
```suggestion
            if (getFeatureFlag(AstBackgroundEvents)) {
```

src/serverModel.ts
```diff
@@ -397,7 +352,7 @@ export default function ServerModel(
             // FIXME: Remove duplicate occurence
             eventObject.CurrencyCode = mpInstance._Store.currencyCode;
             var currentUser = user || mpInstance.Identity.getCurrentUser();
-            self.appendUserInfo(currentUser, eventObject as SDKEvent);
+            appendUserInfo(currentUser, eventObject as SDKEvent);
```
alexs-mparticle, line 355: Nit: I'd move the `as SDKEvent` to the actual declaration.

alexs-mparticle (approved):

PR #992, opened 2025-03-11
feat: Add listeners for signaling when kit is ready

alexs-mparticle:
## Instructions

 ## Summary
 - Adds a listener to Core SDK that kits can use to publish their ready status for subscribers.

 ## Testing Plan
 - Use a sample app (Ideally React or an SPA) to publish and subscribe ready status.

 - Closes https://go.mparticle.com/work/SQDSDKS-7130

### DIFF
--- a/src/forwarders.js
+++ b/src/forwarders.js
@@ -71,7 +71,6 @@ export default function Forwarders(mpInstance, kitBlocker) {
                     );
                     if (!forwarder.initialized) {
                         mpInstance._preInit.kitListeners[forwarder.id] = {
-                            isInitialized: true,
                             isReady: false,
                             callbackQueue: [],
                         };
--- a/src/forwarders.js
+++ b/src/forwarders.js
@@ -70,6 +70,12 @@ export default function Forwarders(mpInstance, kitBlocker) {
                         forwarder.userAttributeFilters
                     );
                     if (!forwarder.initialized) {
+                        mpInstance._preInit.kitListeners[forwarder.id] = {
+                            isInitialized: true,
+                            isReady: false,
+                            callbackQueue: [],
+                        };
+
                         forwarder.logger = mpInstance.Logger;
                         forwarder.init(
                             forwarder.settings,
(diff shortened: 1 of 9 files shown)


src/forwarders.js
```diff
@@ -70,6 +70,12 @@ export default function Forwarders(mpInstance, kitBlocker) {
                         forwarder.userAttributeFilters
                     );
                     if (!forwarder.initialized) {
+                        mpInstance._preInit.kitListeners[forwarder.id] = {
+                            isInitialized: true,
```
rmi22186, line 74: can prob remove
```suggestion
```

src/mp-instance.ts
```diff
@@ -1267,6 +1269,31 @@ export default function mParticleInstance(this: IMParticleWebSDKInstance, instan
         }
     };
     // Used by our forwarders
```
rmi22186, line 1271:
```suggestion
    // addForwarder is called by each individual mp kit
    // kitready is called by each individual mp kit
    // onKitReady is called by the partner who is importing mparticle
```

src/mp-instance.ts
```diff
@@ -1267,6 +1269,31 @@ export default function mParticleInstance(this: IMParticleWebSDKInstance, instan
...
+    this.onKitReady = function(moduleId, callback) {
```
rmi22186, line 1272: queue this if `isInitialized` is false

src/mp-instance.ts
```diff
@@ -1267,6 +1269,31 @@ export default function mParticleInstance(this: IMParticleWebSDKInstance, instan
...
+        const kitListener = self._preInit.kitListeners[moduleId];
+
+        if (!kitListener) {
+            self.Logger.error('Kit listener not found for moduleId: ' + moduleId);
+            return;
+        }
```
rmi22186, line 1278: I would DRY this with a method so that it's a helper function

alexs-mparticle: I think that work should be done as part of https://go.mparticle.com/work/SQDSDKS-4475

PR #994, opened 2025-03-13
feat: Add Rokt Wrapper

alexs-mparticle:
## Instructions

 ## Summary
 - Introduces a wrapper class for the Rokt SDK to allow integrations and partners to access Rokt functionality via the mParticle Web SDK.

 ## Testing Plan
 - Using a sample app, attempt to access the `window.mParticle.Rokt.selectPlacements()` method.
 - Add the [Rokt Web SDK](https://docs.rokt.com/developers/integration-guides/web/best-practices/) to your sample app and attempt to attach the launcher through `window.mParticle.Rokt.attachLauncher()`.
 - `selectPlacement` requests should be queued until the launcher is attached, which will then trigger Rokt's `selectPlacements` logic.

### DIFF
--- a/src/roktManager.ts
+++ b/src/roktManager.ts
@@ -33,19 +33,9 @@ export default class RoktManager {
         this.launcher = null;
     }
 
-    public attachLauncher(launcher: IRoktLauncher): Promise<void> {
-        return new Promise<void>((resolve) => {
-            if (!launcher) {
-                this.queueMessage({
-                    methodName: 'attachLauncher',
-                    payload: launcher,
-                });
-            } else {
-                this.setLauncher(launcher);
-                this.processMessageQueue();
-            }
-            resolve();
-        });
+    public attachLauncher(launcher: IRoktLauncher): void {
+        this.setLauncher(launcher);
+        this.processMessageQueue();
     }
 
     public selectPlacements(options: IRoktSelectPlacementsOptions): Promise<IRoktSelection> {
--- a/src/roktManager.ts
+++ b/src/roktManager.ts
@@ -4,19 +4,20 @@ export interface IRoktPartnerAttributes {
 }
 
 // https://docs.rokt.com/developers/integration-guides/web/library/select-placements-options
-export interface ISelectPlacementsOptions {
+export interface IRoktSelectPlacementsOptions {
     attributes: IRoktPartnerAttributes;
     identifier?: string;
 }
 
-export interface ISelection {
-    placementId?: string;
-    status?: string;
-    error?: string;
+interface IRoktPlacement {}
+
+export interface IRoktSelection {
+    close: () => void;
+    getPlacements: () => Promise<IRoktPlacement[]>;
 }
 
 export interface IRoktLauncher {
-    selectPlacements: (options: ISelectPlacementsOptions) => Promise<ISelection>;
+    selectPlacements: (options: IRoktSelectPlacementsOptions) => Promise<IRoktSelection>;
 }
 
 export interface IRoktMessage {
@@ -25,7 +26,7 @@ export interface IRoktMessage {
 }
 
 export default class RoktManager {
-    private launcher: IRoktLauncher | null = null;
+    public launcher: IRoktLauncher | null = null;
     private messageQueue: IRoktMessage[] = [];
 
     constructor() {
@@ -47,13 +48,13 @@ export default class RoktManager {
         });
     }
 
-    public selectPlacements(options: ISelectPlacementsOptions): Promise<ISelection> {
+    public selectPlacements(options: IRoktSelectPlacementsOptions): Promise<IRoktSelection> {
         if (!this.launcher) {
             this.queueMessage({
                 methodName: 'selectPlacements',
                 payload: options,
             });
-            return Promise.resolve({});
+            return Promise.resolve({} as IRoktSelection);
         }
 
         try {
--- a/src/roktManager.ts
+++ b/src/roktManager.ts
@@ -1,7 +1,9 @@
+// https://docs.rokt.com/developers/integration-guides/web/library/attributes
 export interface IRoktPartnerAttributes {
     [key: string]: string | number | boolean | undefined | null;
 }
 
+// https://docs.rokt.com/developers/integration-guides/web/library/select-placements-options
 export interface ISelectPlacementsOptions {
     attributes: IRoktPartnerAttributes;
     identifier?: string;
--- a/src/roktManager.ts
+++ b/src/roktManager.ts
@@ -1,0 +1,83 @@
+export interface IRoktPartnerAttributes {
+    [key: string]: string | number | boolean | undefined | null;
+}
+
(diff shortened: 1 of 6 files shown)


src/roktManager.ts
```diff
@@ -0,0 +1,83 @@
...
+export interface ISelectPlacementsOptions {
+    attributes: IRoktPartnerAttributes;
+    identifier?: string;
+}
+
+export interface ISelection {
+    placementId?: string;
+    status?: string;
+    error?: string;
+}
```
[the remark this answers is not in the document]
rmi22186: i see https://docs.rokt.com/developers/integration-guides/web/library/integration-launcher#select-placements which shows that yes, it does return `Promise<selection>` but if you click on `selection` in this page, it brings you here which doesn't have much
https://docs.rokt.com/developers/integration-guides/web/library/selection

src/roktManager.ts
```diff
@@ -0,0 +1,83 @@
...
+    selectPlacements: (options: ISelectPlacementsOptions) => Promise<ISelection>;
+}
+
+export interface IRoktMessage {
+    methodName: string;
+    payload: any;
+}
+
+export default class RoktManager {
+    private launcher: IRoktLauncher | null = null;
```
rmi22186, line 26: should this be public?  the reason we put this here also is so that a customer could access it right?

rmi22186 (changes requested):

test/jest/roktManager.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+            expect(roktManager['messageQueue'][0].payload).toBe(null);
+
+            roktManager.attachLauncher(launcher);
+            expect(roktManager['launcher']).not.toBeNull();
+            expect(roktManager['messageQueue'].length).toBe(0);
+        });
+    });
+
+    describe('#selectPlacements', () => {
+        it('should call selectPlacements', () => {
```
rmi22186, line 51: in general for all these tests, i think it gets confusing when you are saying selectPlacements gets called.  we should be more explicit that one `selectPlacements` gets mapped to anotheri `selectPlacements` call think being explicit is good here.  `should call launcher.selectPlacements when roktManager.selectPlacements is called`

```suggestion
        it('should map a IRoktLauncher selectPlacement to a launcher selectPlacements call', () => {
```

test/jest/roktManager.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+            const launcher = {
+                selectPlacements: jest.fn()
+            } as unknown as IRoktLauncher;
```
rmi22186, line 54: I don't think there's a reason to make it `unknown`

```suggestion
            const launcher: IRoktLauncher = {
                selectPlacements: jest.fn()
            }
```

test/jest/roktManager.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+            roktManager.attachLauncher(launcher);
+            const options = {
+                attributes: {}
+            } as ISelectPlacementsOptions;
+
+            roktManager.selectPlacements(options);
+            expect(launcher.selectPlacements).toHaveBeenCalledWith(options);
+        });
+
+        it('should call selectPlacements with any passed in attributes', () => {
```
rmi22186, line 65:
```suggestion
        it('should call IRoktLauncher selectPlacements with any passed in attributes to roktManager.selectPlacements', () => {
```

rmi22186 (commented):
minor comments

src/roktManager.ts
```diff
@@ -4,19 +4,20 @@ export interface IRoktPartnerAttributes {
...
     identifier?: string;
 }
 
-export interface ISelection {
-    placementId?: string;
-    status?: string;
-    error?: string;
+interface IRoktPlacement {}
+
+export interface IRoktSelection {
```
rmi22186, line 14: i'd add links for each interface that we're pulling for easier searching in the future
https://docs.rokt.com/developers/integration-guides/web/library/selection

rmi22186: maybe there should be a jira ticket for ensuring these interfaces eventually get fully typed since they aren't all here right now?

test/jest/roktManager.spec.ts
alexs-mparticle: I think "map an IRoktLauncher selectPlacement" is too much of an implementation detail for this specific tests, but I agree that it should more explicit to referenc the launcher. I'll revise.

src/roktManager.ts
```diff
@@ -0,0 +1,86 @@
...
+            if (!launcher) {
+                this.queueMessage({
+                    methodName: 'attachLauncher',
+                    payload: launcher,
+                });
+            } else {
+                this.setLauncher(launcher);
+                this.processMessageQueue();
+            }
+            resolve();
```
rmi22186, line 47: no need for this to be a promise

src/roktManager.ts
```diff
@@ -0,0 +1,86 @@
...
+        this.launcher = null;
+    }
+
+    public attachLauncher(launcher: IRoktLauncher): Promise<void> {
+        return new Promise<void>((resolve) => {
+            if (!launcher) {
+                this.queueMessage({
+                    methodName: 'attachLauncher',
+                    payload: launcher,
+                });
```
rmi22186, line 42: this can be removed.  the expectation is there is no launcher and we are setting.

test/jest/roktManager.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+        it('should attach a launcher', () => {
+            const launcher = {} as IRoktLauncher;
+            roktManager.attachLauncher(launcher);
+            expect(roktManager['launcher']).not.toBeNull();
+        });
+
+        it('should process the message queue if a launcher is attached', () => {
+            const launcher = {
+                selectPlacements: jest.fn()
+            } as unknown as IRoktLauncher;
```
rmi22186, line 30: Can this just be
```suggestion
            const launcher: IRoktLauncher = {
                selectPlacements: jest.fn()
            };
```

or does that cause issues somewhere?

test/jest/roktManager.spec.ts
```diff
@@ -24,26 +24,21 @@ describe('RoktManager', () => {
...
-        });
-
         it('should process the message queue if a launcher is attached', () => {
-            const launcher = jest.fn() as unknown as IRoktLauncher;
+            const launcher = {
+                selectPlacements: jest.fn()
+            } as unknown as IRoktLauncher;
+            
+            roktManager.selectPlacements({} as IRoktSelectPlacementsOptions);
+            roktManager.selectPlacements({} as IRoktSelectPlacementsOptions);
```
rmi22186, line 33: delete

test/jest/roktManager.spec.ts
alexs-mparticle: Good catch. I miseed that one.

rmi22186 (approved):

PR #998, opened 2025-03-17
test: Remove set timeouts from tests and other cleanup

rmi22186:
## Instructions

 ## Summary
This PR pulls 2 commits [from here](https://github.com/mParticle/mparticle-web-sdk/pull/967/commits), and then also updates the identity tests to use async/await - this file seemed to be the cause of some other tests breaking.  

 ## Testing Plan
Unit tests run successfullyl

 - Closes https://go.mparticle.com/work/SQDSDKS-7143

### DIFF
--- a/test/src/tests-kit-blocking.ts
+++ b/test/src/tests-kit-blocking.ts
@@ -1406,7 +1406,7 @@ describe('kit blocking', () => {
                 fetchMock.restore();
             });
 
-            it('should create a proper kitblocker on a self hosted set up', function(done) {
+            it('should create a proper kitblocker on a self hosted set up', async () => {
                 fetchMock.get(`${urls.config}&plan_id=robs_plan&plan_version=1`, {
                     status: 200,
                     body: JSON.stringify({ dataPlanResult: dataPlan }),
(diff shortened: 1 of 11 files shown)


test/src/tests-kit-blocking.ts
```diff
@@ -1427,21 +1427,17 @@ describe('kit blocking', () => {
                 window.mParticle.config.kitConfigs.push(forwarderDefaultConfiguration('MockForwarder'));
                 window.mParticle.init(apiKey, window.mParticle.config);
 
-                window.mParticle.logEvent('Blocked event');
+                await waitForCondition(hasIdentifyReturned);
```
alexs-mparticle, line 1430: Should these have changed positions?

test/src/tests-mparticle-instance-manager.ts
```diff
@@ -450,74 +440,59 @@ describe('mParticle instance manager', () => {
                 5
             );
 
-            // setTimeout to allow config to come back from the beforeEach initialization
+            await waitForCondition(hasConfigurationReturned);
```
alexs-mparticle, line 443: Same here. Why is this going before the logging function?

alexs-mparticle (commented):
LGTM. Just two questions.

test/src/tests-kit-blocking.ts
rmi22186: technically it doesn't matter. i swapped the two just to show the tests still pass, but the reason it doesn't matter is the setTimeout previously just ensured the logged event wasn't in the queue and actually went to the forwarders.  so whether it is prior to the event being logged (so that it doesn't reach the queue), or after the event is logged (in which case the event goes into the queue, and the await flushes it out), it doesn't matter

test/src/tests-mparticle-instance-manager.ts
rmi22186: same answer

test/src/tests-kit-blocking.ts
alexs-mparticle: Works for me.

alexs-mparticle (approved):

PR #999, opened 2025-03-17
refactor: Create Distinct Kit Interfaces

alexs-mparticle:
## Instructions

 ## Summary
 - Defines specific interfaces for Kits:
 - UnregisteredKit - The state of a kit when imported via a kit config via NPM/CDN
 - RegisteredKit - A kit that has a constructor and is assigned to the forwarderConstructors data structure and can be configured
 - ConfiguredKit - the "final" state of a kit where it can be acted upon by the SDK

 ## Testing Plan
 - N/A - this just changes kit interfaces within the SDK

 - Closes https://go.mparticle.com/work/SQDSDKS-6035

### DIFF
--- a/src/forwarders.interfaces.ts
+++ b/src/forwarders.interfaces.ts
@@ -26,7 +26,7 @@ export interface RegisteredKit {
     constructor: () => void;
 
     // Applies to sideloaded kits only
-    filters: IKitFilterSettings;
+    filters?: IKitFilterSettings;
 }
 
 // This is the subset of the SDKConfig.kits object that is used to register kits.
@@ -38,6 +38,7 @@ export interface KitRegistrationConfig {
 export interface ConfiguredKit
     extends Omit<IKitConfigs, 'isDebugString' | 'hasDebugString'> {
     common: Dictionary<unknown>;
+    id: number;
     init(
         settings: Dictionary<unknown>,
         service: forwardingStatsCallback,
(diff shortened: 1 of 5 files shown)


src/forwarders.interfaces.ts
```diff
@@ -1,29 +1,37 @@
...
 // The state of the kit after being added to forwarderConstructors in the CDN
 // or after registered to SDKConfig.kits via NPM
 export interface RegisteredKit {
-    constructor(): void;
-    name: string;
-    getId(): number;
+    constructor: () => void;
+
+    // Applies to sideloaded kits only
+    filters: IKitFilterSettings;
```
rmi22186, line 29: if it applies to sideloaded kits only, perhaps this should be optional?

test/jest/utils.ts
```diff
@@ -37,11 +37,11 @@ export class MockForwarder {
         this.moduleId = id || 1;
     }
 
-    public register = (config: SDKInitConfig): void => {
+    public register = (config: KitRegistrationConfig): void => {
         if (config.kits) {
             config.kits[this.name] = {
-                constructor: this.constructor,
-            };
+                constructor: this.constructor as () => RegisteredKit,
```
rmi22186, line 43: is this line necessary?  i would have expected just lin 44 to be the `RegisteredKit`

src/forwarders.interfaces.ts
alexs-mparticle: Perhaps. The issue is that in the code for `processSideloadedKits`, it treats `filters` as required. IMO we may want to change the behavior of `processSideloadedKits` to internally use a SideloadedKit interface rather than RegisteredKit. I can look into that as part of https://go.mparticle.com/work/SQDSDKS-4475

test/jest/utils.ts
alexs-mparticle: It is if we make `filters` required. By making it optional, I can get rid of line 44.

rmi22186 (approved):

PR #1007, opened 2025-03-24
feat: Add User Filtering for Rokt Wrapper

alexs-mparticle:
## Instructions

 ## Summary
 - Exposes User Filtering Logic to Rokt Web Kit

 ## Testing Plan
 - Using a sample a with the latest version of the Rokt Kit, verify that filtered user attribute are not passed into the `selectPlacements` call

 - Closes https://go.mparticle.com/work/SQDSDKS-7162

### DIFF
--- a/src/forwarders.interfaces.ts
+++ b/src/forwarders.interfaces.ts
@@ -1,7 +1,7 @@
 import { SDKEvent, SDKEventCustomFlags } from './sdkRuntimeModels';
 import { Dictionary } from './utils';
 import { IKitConfigs, IKitFilterSettings } from './configAPIClient';
-import { IdentityApiData } from '@mparticle/web-sdk';
+import { IdentityApiData, IdentityType } from '@mparticle/web-sdk';
 import {
     IMParticleUser,
     ISDKUserIdentity,
@@ -12,16 +12,27 @@ import {
 export type MPForwarder = Dictionary;
 
 // The state of the kit when accessed via window.KitName via CDN
-// or imported as an NPM package
+// or imported as an NPM package but before it goes through the registration process
+// This also applies to sideloaded kits which have not yet been registered
 export interface UnregisteredKit {
     constructor: () => void;
     register(config: KitRegistrationConfig): void;
     name: string;
+
+
+    // Optional Attributes that are not used for sideloaded kits
+
+    // Module ID is used for kits that are provided via CDN or NPM
+    moduleId?: number;
+
+    // Suffix is used for kits that are provided via CDN or NPM
+    // and is used for version pinning of the SDK
     suffix?: string;
 }
 
 // The state of the kit after being added to forwarderConstructors in the CDN
 // or after registered to SDKConfig.kits via NPM
+// Sideloaded Kits would also be considered registere
 export interface RegisteredKit {
     constructor: () => void;
 
@@ -81,6 +92,8 @@ export interface ConfiguredKit
 
 export type UserIdentityId = string;
 export type UserIdentityType = number;
+export type UserAttributeFilters = number[];
+export type UserIdentityFilters = typeof IdentityType[];
 
 export type forwardingStatsCallback = (
     forwarder: ConfiguredKit,
--- a/src/forwarders.interfaces.ts
+++ b/src/forwarders.interfaces.ts
@@ -19,7 +19,6 @@ export interface UnregisteredKit {
     register(config: KitRegistrationConfig): void;
     name: string;
 
-
     // Optional Attributes that are not used for sideloaded kits
 
     // Module ID is used for kits that are provided via CDN or NPM
@@ -32,7 +31,7 @@ export interface UnregisteredKit {
 
 // The state of the kit after being added to forwarderConstructors in the CDN
 // or after registered to SDKConfig.kits via NPM
-// Sideloaded Kits would also be considered registere
+// Sideloaded Kits would also be considered registered
 export interface RegisteredKit {
     constructor: () => void;
 
--- a/src/mp-instance.ts
+++ b/src/mp-instance.ts
@@ -36,7 +36,7 @@ import Consent, { IConsent } from './consent';
 import KitBlocker from './kitBlocking';
 import ConfigAPIClient, { IKitConfigs } from './configAPIClient';
 import IdentityAPIClient from './identityApiClient';
-import { isFunction } from './utils';
+import { isFunction, parseConfig } from './utils';
 import { LocalStorageVault } from './vault';
 import { removeExpiredIdentityCacheDates } from './identity-utils';
 import IntegrationCapture from './integrationCapture';
@@ -1395,7 +1395,7 @@ function completeSDKInitialization(apiKey, config, mpInstance) {
         }
 
         // Configure Rokt Manager with filtered user
-        const roktConfig = parseConfig(config, 'Rokt', 181);
+        const roktConfig: IKitConfigs = parseConfig(config, 'Rokt', 181);
         if (roktConfig) {
             const { userAttributeFilters } = roktConfig;
             const roktFilteredUser = filteredMparticleUser(
@@ -1586,12 +1586,5 @@ function queueIfNotInitialized(func, self) {
         return true;
     }
     return false;
-}
-
-function parseConfig(config: SDKInitConfig, moduleName: string, moduleId: number): IKitConfigs {
-    return config.kitConfigs?.find((kitConfig: IKitConfigs) => 
-        kitConfig.name === moduleName && 
-        kitConfig.moduleId === moduleId
-    ) || null;
 }
 
--- a/src/mp-instance.ts
+++ b/src/mp-instance.ts
@@ -34,14 +34,14 @@ import ForwardingStatsUploader from './forwardingStatsUploader';
 import Identity from './identity';
 import Consent, { IConsent } from './consent';
 import KitBlocker from './kitBlocking';
-import ConfigAPIClient from './configAPIClient';
+import ConfigAPIClient, { IKitConfigs } from './configAPIClient';
 import IdentityAPIClient from './identityApiClient';
-import { isFunction, valueof } from './utils';
+import { isFunction } from './utils';
 import { LocalStorageVault } from './vault';
 import { removeExpiredIdentityCacheDates } from './identity-utils';
 import IntegrationCapture from './integrationCapture';
 import { IPreInit, processReadyQueue } from './pre-init-utils';
-import { BaseEvent, MParticleWebSDK, SDKHelpersApi } from './sdkRuntimeModels';
+import { BaseEvent, MParticleWebSDK, SDKHelpersApi, SDKInitConfig } from './sdkRuntimeModels';
 import { Dictionary, SDKEventAttrs } from '@mparticle/web-sdk';
 import { IIdentity } from './identity.interfaces';
 import { IEvents } from './events.interfaces';
@@ -1395,7 +1395,7 @@ function completeSDKInitialization(apiKey, config, mpInstance) {
         }
 
         // Configure Rokt Manager with filtered user
-        const roktConfig = mpInstance._RoktManager.parseConfig(config);
+        const roktConfig = parseConfig(config, 'Rokt', 181);
         if (roktConfig) {
             const { userAttributeFilters } = roktConfig;
             const roktFilteredUser = filteredMparticleUser(
@@ -1403,7 +1403,7 @@ function completeSDKInitialization(apiKey, config, mpInstance) {
                 { userAttributeFilters },
                 mpInstance
             );
-            mpInstance._RoktManager.init(config, roktFilteredUser);
+            mpInstance._RoktManager.init(roktConfig, roktFilteredUser);
         }
 
         mpInstance._Forwarders.processForwarders(
@@ -1586,5 +1586,12 @@ function queueIfNotInitialized(func, self) {
         return true;
     }
     return false;
+}
+
+function parseConfig(config: SDKInitConfig, moduleName: string, moduleId: number): IKitConfigs {
+    return config.kitConfigs?.find((kitConfig: IKitConfigs) => 
+        kitConfig.name === moduleName && 
+        kitConfig.moduleId === moduleId
+    ) || null;
 }
 
--- a/src/mp-instance.ts
+++ b/src/mp-instance.ts
@@ -41,7 +41,7 @@ import { LocalStorageVault } from './vault';
 import { removeExpiredIdentityCacheDates } from './identity-utils';
 import IntegrationCapture from './integrationCapture';
 import { IPreInit, processReadyQueue } from './pre-init-utils';
-import { BaseEvent, MParticleWebSDK, SDKHelpersApi, SDKInitConfig } from './sdkRuntimeModels';
+import { BaseEvent, MParticleWebSDK, SDKHelpersApi } from './sdkRuntimeModels';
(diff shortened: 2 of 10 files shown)


src/mp-instance.ts
```diff
@@ -1393,6 +1394,14 @@ function completeSDKInitialization(apiKey, config, mpInstance) {
             mpInstance._IntegrationCapture.capture();
         }
 
+        // Configure Rokt Manager with filtered user
+        const roktConfig = mpInstance._RoktManager.parseConfig(config);
+        const roktUserFilters = roktConfig?.userAttributeFilters;
+        const roktFilteredUser = filteredMparticleUser(currentUserMPID, {
+            userAttributeFilters: roktUserFilters,
+        }, mpInstance);
+        mpInstance._RoktManager.init(config, roktFilteredUser);
```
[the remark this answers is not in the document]
rmi22186: OR we could just do something like 
```
if (roktConfig) {
   mpInstance._RoktManager.init(roktConfig);
}
```
then move the consts for `roktUserFilter` and `roktFilteredUser` into the constructor. This would be cleaner, and then you wouldn't be parsing the `config` twice for the `roktConfig`. 🚀

src/roktManager.ts
```diff
@@ -25,21 +32,67 @@ export interface IRoktMessage {
...
+            roktForwarder.name === 'Rokt' && 
+            roktForwarder.moduleId === 181
+        );
     }
 
     public attachLauncher(launcher: IRoktLauncher): void {
-        this.setLauncher(launcher);
+        this.launcher = launcher;
+        this.processMessageQueue();
+    }
```
rmi22186, line 87: let's remove since now attaching the kit calls processMessageQueue to say that the kit and launcher are ready inside the kit

test/jest/kitFilterHelper.spec.ts
```diff
@@ -386,4 +386,92 @@ describe('FilterHashingUtilities', () => {
...
+            });
+        });
+    });
+
+    describe('#filterUserIdentities', () => {
+        it('should return the original user identities if no filter list is provided', () => {
+            const userIdentities = {
+                'foo-key': 'foo-value',
+                'bar-key': 'bar-value',
+                'baz-key': 'baz-value',
```
rmi22186, line 429: these shoudl be userIdentities
```suggestion
                'other': 'foo-value',
                'customerid': 'bar-value',
                'facebook': 'baz-value',
```

test/jest/roktManager.spec.ts
```diff
@@ -17,17 +19,85 @@ describe('RoktManager', () => {
...
+            roktManager.init({} as SDKInitConfig);
+            expect(roktManager['launcher']).toBeNull();
+            expect(roktManager['filters']).toEqual({
+                userAttributeFilters: undefined,
+                filterUserAttributes: expect.any(Function),
+            });
+            expect(roktManager['filteredUser']).toBeUndefined();
+            expect(roktManager['kit']).toBeNull();
+        });
+        it('should initialize the manager with user attribute filters from a config', () => {
```
rmi22186, line 33:
```suggestion
        });

        it('should initialize the manager with user attribute filters from a config', () => {
```

test/jest/roktManager.spec.ts
```diff
@@ -17,17 +19,85 @@ describe('RoktManager', () => {
...
+            expect(roktManager['filters']).toEqual({
+                userAttributeFilters: [816506310, 1463937872, 36300687],
+                filterUserAttributes: expect.any(Function),
+            });
+        });
+    });
+
+    describe('#parseConfig', () => {
+        it('should ONLY return the kit config for Rokt', () => {
+            const kitConfig: Partial<IKitConfigs> = {
```
rmi22186, line 54: for clarity, rename this constant
```suggestion
            const roktKitConfig: Partial<IKitConfigs> = {
```

test/jest/roktManager.spec.ts
```diff
@@ -17,17 +19,85 @@ describe('RoktManager', () => {
...
+            expect(result).toEqual(kitConfig);
+        });
+
+        it('should return undefined if no kit config is found', () => {
+            const config: SDKInitConfig = {
+                kitConfigs: [],
+            };
+
+            const result = roktManager.parseConfig(config);
+            expect(result).toBeUndefined();
```
rmi22186, line 78: nit: i think the source code should be refactored so that this returns `null`, since `undefined` means a particular thing just isn't defined, and `null` is the absence of something

test/jest/roktManager.spec.ts
```diff
@@ -17,17 +19,85 @@ describe('RoktManager', () => {
...
+        });
+    });
+
     describe('#attachLauncher', () => {
```
rmi22186, line 82: as above, I don't think we need to attach the launcher to the core anymore, so this entire test block should be removed when the code is removed in the core sdk above

test/jest/roktManager.spec.ts
```diff
@@ -36,33 +106,100 @@ describe('RoktManager', () => {
...
+
+            roktManager.selectPlacements({} as IRoktSelectPlacementsOptions);
+            roktManager.selectPlacements({} as IRoktSelectPlacementsOptions);
+            roktManager.selectPlacements({} as IRoktSelectPlacementsOptions);
+            
+            expect(roktManager['messageQueue'].length).toBe(3);
+            expect(launcher.selectPlacements).toHaveBeenCalledTimes(0);
+
+            // Indirectly attach the launcher and kit to test the message queue
+            roktManager.launcher = launcher;
```
rmi22186, line 154: remove launcher

test/jest/roktManager.spec.ts
```diff
@@ -36,33 +106,100 @@ describe('RoktManager', () => {
...
                 attributes: {}
             } as IRoktSelectPlacementsOptions;
 
             roktManager.selectPlacements(options);
-            expect(launcher.selectPlacements).toHaveBeenCalledWith(options);
+            expect(kit.selectPlacements).toHaveBeenCalledWith(options);
         });
 
-        it('should call launcher.selectPlacements with passed in attributes', () => {
+        it.only('should call kit.selectPlacements with passed in attributes', () => {
```
rmi22186, line 188:
```suggestion
        it('should call kit.selectPlacements with passed in attributes', () => {
```

test/jest/roktManager.spec.ts
```diff
@@ -36,33 +106,100 @@ describe('RoktManager', () => {
...
 
+            const kit: IRoktKit = {
+                launcher: launcher,
+                selectPlacements: jest.fn(),
+                filters: undefined,
+                filteredUser: undefined,
+                userAttributes: undefined
+            };
+
             roktManager.attachLauncher(launcher);
```
rmi22186, line 201: do a global search in the core sdk and tests for `attachLauncher` and remove

test/jest/utils.spec.ts
```diff
@@ -218,4 +220,56 @@ describe('Utils', () => {
...
+        });
+
+        it('should support different types of arrays', () => {
+            expect(inArray([1, 2, 3], 2)).toBe(true);
+            expect(inArray([1, 2, 3], '2')).toBe(false);
+            expect(inArray([1, 2, 3], 4)).toBe(false);
+
+            expect(inArray([1, 2, 3], 2)).toBe(true);
+            expect(inArray([1, 2, 3], '2')).toBe(false);
+            expect(inArray([1, 2, 3], 4)).toBe(false);
```
rmi22186, line 252: these are repeats of the lines directly above

test/jest/utils.spec.ts
```diff
@@ -218,4 +220,56 @@ describe('Utils', () => {
...
+
+        it('returns false if the array is null', () => {
+            expect(inArray(null, 'foo')).toBe(false);
+        });
+
+        it('returns false if the array is undefined', () => {
+            expect(inArray(undefined, 'foo')).toBe(false);
+        });
+
+        it('should support different types of arrays', () => {
```
rmi22186, line 245: I don't think the test naming is accurately stating what the test is doing here
```suggestion
        it('should be strict about the value type that is in the array', () => {
```

test/jest/utils.spec.ts
```diff
@@ -218,4 +220,56 @@ describe('Utils', () => {
...
+        it('should filter a dictionary based on a hash function and filter list', () => {
+            const dictionary = {
+                'foo': 'bar',
+                'bar': 'baz',
+                'baz': 'qux',
+                'quux': 'corge',
+            };
+    
+            const filterList = [98, 102];
+            const hashFn = (key: string) => key.charCodeAt(0);
```
rmi22186, line 266: this isn't human readable, can you make the hash function more simple?

rmi22186 (changes requested):
lots of little nits, but looks good overall!

src/roktManager.ts
```diff
@@ -25,21 +32,67 @@ export interface IRoktMessage {
...
+        this.filters = {
+            userAttributeFilters,
+            filterUserAttributes: KitFilterHelper.filterUserAttributes,
+        };
+
+        this.filteredUser = filteredUser;
+    }
+
+    public parseConfig(config: SDKInitConfig): IKitConfigs {
+        return config.kitConfigs?.find((roktForwarder: IKitConfigs) =>
```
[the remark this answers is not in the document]
alexs-mparticle: I like the idea of renaming `IKitConfigs` to be singular. I'll do that as part of https://go.mparticle.com/work/SQDSDKS-4475

test/jest/roktManager.spec.ts
alexs-mparticle: I think that's a great idea actually and not a nit.

test/jest/utils.spec.ts
alexs-mparticle: That's not exactly true. There's no real strict typing here, but rather it accepts all types.

src/forwarders.interfaces.ts
```diff
@@ -12,16 +12,27 @@ import {
...
+    moduleId?: number;
+
+    // Suffix is used for kits that are provided via CDN or NPM
+    // and is used for version pinning of the SDK
     suffix?: string;
 }
 
 // The state of the kit after being added to forwarderConstructors in the CDN
 // or after registered to SDKConfig.kits via NPM
+// Sideloaded Kits would also be considered registere
```
rmi22186, line 35:
```suggestion
// Sideloaded Kits would also be considered registered
```

src/forwarders.interfaces.ts
```diff
@@ -12,16 +12,27 @@ import {
...
 // The state of the kit when accessed via window.KitName via CDN
-// or imported as an NPM package
+// or imported as an NPM package but before it goes through the registration process
+// This also applies to sideloaded kits which have not yet been registered
 export interface UnregisteredKit {
     constructor: () => void;
     register(config: KitRegistrationConfig): void;
     name: string;
+
+
```
rmi22186, line 22:
```suggestion
```

rmi22186 (changes requested):

test/jest/utils.spec.ts
rmi22186: hm.  the arrays here are all the same though.  what's different is the 2nd argument that you are looking for

src/roktManager.ts
```diff
@@ -25,21 +32,58 @@ export interface IRoktMessage {
...
+
+    private filteredUser: IMParticleUser | null = null;
     private messageQueue: IRoktMessage[] = [];
 
     constructor() {
-        this.launcher = null;
+        this.kit = null;
+        this.filters = {};
+        this.filteredUser = null;
     }
```
rmi22186, line 59: the constructor is redundant with the default values. I think it can be removed

src/roktManager.ts
```diff
@@ -48,29 +92,28 @@ export default class RoktManager {
...
-            return this.launcher.selectPlacements(options);
+            return this.kit.selectPlacements(options);
         } catch (error) {
             return Promise.reject(error instanceof Error ? error : new Error('Unknown error occurred'));
         }
     }
 
+    private isReady(): boolean {
+        return Boolean(this.kit);
+    }
```
rmi22186, line 103: i suggest either setting a specific `isReady` key (inside of attachKit), or I'd be very explicit and put a comment here that says that the Rokt Manager is in a ready state only once a kit is attached.

it might be helpful to have a comment right above the class and link to the rokt web kit to discus the order of the flow, so that future developers can grok it more easily

rmi22186 (changes requested):

src/mp-instance.ts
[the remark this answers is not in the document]
rmi22186: @alexs-mparticle - will you address this comment or make a ticket?

src/mp-instance.ts
```diff
@@ -1393,6 +1394,18 @@ function completeSDKInitialization(apiKey, config, mpInstance) {
             mpInstance._IntegrationCapture.capture();
         }
 
+        // Configure Rokt Manager with filtered user
+        const roktConfig = parseConfig(config, 'Rokt', 181);
```
rmi22186, line 1398:
```suggestion
        const roktConfig: IKitConfigs = parseConfig(config, 'Rokt', 181);
```

rmi22186 (changes requested):

rmi22186 (approved):

test/jest/roktManager.spec.ts
```diff
@@ -75,41 +137,55 @@ describe('RoktManager', () => {
...
+            const kit: IRoktKit = {
+                launcher: {
+                    selectPlacements: jest.fn()
+                },
+                filters: undefined,
+                filteredUser: undefined,
+                userAttributes: undefined,
+
+                // We are mocking the selectPlacements method to return the
+                // launcher's selectPlacements method and verify that the
```
rmi22186, line 166:
```suggestion
                // launcher's selectPlacements method and verify that
```

rmi22186 (changes requested):
this still has the dist and rollup.config.js files in them.  please remove.  otherwise LGTM

src/roktManager.ts
```diff
@@ -25,21 +32,53 @@ export interface IRoktMessage {
...
+// The Rokt Manager should load before the Web Kit and stubs out many of the
+// Rokt Web SDK functions with an internal message queue in case a Rokt function
+// is requested before the Rokt Web Kit or SDK is finished loaded.
+// Once the Rokt Kit is attached to the Rokt Manager, we can consider the
+// Rokt Manager in a "ready" state and it can begin sending data to the kit.
+//
+// https://github.com/mparticle-integrations/mparticle-javascript-integration-rokt
 export default class RoktManager {
-    public launcher: IRoktLauncher | null = null;
+    public config: SDKInitConfig | null = null;
```
rmi22186, line 59: no need for this
```suggestion
```

src/roktManager.ts
```diff
@@ -25,21 +32,53 @@ export interface IRoktMessage {
...
+    public kit: IRoktKit = null;
+    public filters: RoktKitFilterSettings = {};
+
+    private filteredUser: IMParticleUser | null = null;
     private messageQueue: IRoktMessage[] = [];
 
-    constructor() {
-        this.launcher = null;
+    public init(config: IKitConfigs, filteredUser?: IMParticleUser): void {
```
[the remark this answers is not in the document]
rmi22186: see this again

src/mp-instance.ts
rmi22186: see this again

rmi22186 (changes requested):

rmi22186 (approved):

PR #1013, opened 2025-04-01
feat: Add sandbox attribute depending on isDevelopmentMode

crisryantan:
## Instructions

 ## Summary
 - Explicitly add sandbox attribute when calling selectPlacements depending on SDKInitConfig.isDevelopmentMode

 ## Testing Plan
 - Unit tests

### DIFF
--- a/src/mp-instance.ts
+++ b/src/mp-instance.ts
@@ -36,7 +36,7 @@ import Consent, { IConsent } from './consent';
 import KitBlocker from './kitBlocking';
 import ConfigAPIClient, { IKitConfigs } from './configAPIClient';
 import IdentityAPIClient from './identityApiClient';
-import { isFunction } from './utils';
+import { isFunction, parseConfig } from './utils';
 import { LocalStorageVault } from './vault';
 import { removeExpiredIdentityCacheDates } from './identity-utils';
 import IntegrationCapture from './integrationCapture';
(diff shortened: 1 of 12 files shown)


crisryantan:
As discussed with @alexs-mparticle, we're reverting this implementation back to using attributes as originally requested. In a joint deployment scenario, it's important that we maintain consistency, either all systems are configured globally or per page; we can't mix both approaches.

src/mp-instance.ts
```diff
@@ -1392,7 +1392,7 @@ function completeSDKInitialization(apiKey, config, mpInstance) {
         if (mpInstance._Helpers.getFeatureFlag(CaptureIntegrationSpecificIds)) {
             mpInstance._IntegrationCapture.capture();
         }
-
+        mpInstance._RoktManager.init(config);
```
crisryantan, line 1395: Let me know if this is the right approach to go with. Alternatively, I can make RoktManager dependent on config when instantiating so we don't have to explicitly call init

rmi22186 (commented):
@chrisryantan @alexs-mparticle this looks great!  the only feedback i have is, can we add logic for if a partner passes in `sandbox` directly to the `mParticle.Rokt.selectPlacements` call, that we should prioritize that over what is in the `isDevelopmentMode`

src/roktManager.ts
```diff
@@ -88,7 +88,19 @@ export default class RoktManager {
         }
 
         try {
-            return this.kit.selectPlacements(options);
+            let enhancedOptions = options;
+
+            if (this.config) {
```
rmi22186, line 102: I just realized that changes @alexs-mparticle  will making will break this because we will be passing the Rokt config directly, and not the mp config anymore.  I flagged this with Alex, and he had some ideas about how to get this still, including passing `inDevelopmentMode` as an option as a 3rd parameter.

alexs-mparticle:
Closing in favor of #1015

PR #1014, opened 2025-04-01
fix: identifyRequest intake when persistent data exists

### LINKED ISSUE #1010: Tracking issues across subdomains
I'm running into problems when tracking users across subdomains with this SDK.

My user flow:

1. customer visits my.domain.com. `mParticle.init()` is called when the page loads.
3. customer picks a product to buy
4. customer logs in and `mParticle.login()` is called, resulting in the correct `user_identities` being sent alongside future mParticle events
5. customer is redirected to www.domain.com to complete checkout. `mParticle.init()` is called when the page loads.

**Expected result:** the `user_identities` object contains our user's `customerid`.

**Actual result:** the `user_identities` object is empty.

The initializing code looks like this on the www.domain.com subdomain:

```ts
import mParticle from '@mparticle/web-sdk';

const mParticleConfig = {
  deviceId: '...',
  workspaceToken: '...',
  identifyRequest: {
    userIdentities: {
      // The value below comes from a cookie. I verified that
      // it is valid before building this config object.
      customerid: "12345678",
    }
  },
  dataPlan: {
    planId: '...',
  },
};

mParticle.init("api id goes here", mParticleConfig);
```

If I reload the page at this point, the next `mParticle.init()` call will work as expected. Could you help me figure out whether this is a user error or a bug in the SDK? Does the SDK maintain some kind of persistence layer that carries over between subdomains and would prevent the `identifyRequest` field from taking effect?

Thank you in advance 🙏

deammer:
## Summary

This fixes an issue with identifying users in certain contexts, namely across subdomains. See [this comment](https://github.com/mParticle/mparticle-web-sdk/issues/1010#issuecomment-2765301880) for more information.

Resolves #1010. 

 ## Testing Plan

 - [x] Was this tested locally? ➡️ It was tested successfully on real live traffic.

The testing scenario:

1. customer lands on www.website.com
  a. mParticle initializes on the subdomain
  b. mParticle starts a session and stores it in local storage
2. customer is sent to login.website.com to log in
  a. mParticle initializes on the subdomain
  b. mParticle starts a session and stores it in local storage
3. customer logs in
  a. `mParticle.login()` is called
  b. a cookie containing the `customerid` is stored
  c. user is thus identified on login.website.com
4. customer is sent back to www.website.com
  a. mParticle initializes on the subdomain
  b. mParticle **does not start a new session** because there's already one in local storage

**Before this fix:** users would not be identified.

**With this fix:** users are consistently identified.


dist/mparticle.common.js
```diff
@@ -508,8 +508,12 @@ if(!isEmpty(k)&&(isEmpty(o)||(i=!0),!(i&&d)))// If MPID is new to cookies, we sh
...
+var Messages$6=Constants.Messages;function SessionManager(a){var b=this;this.initialize=function(){var c;if(a._Store.sessionId){var d=6e4*a._Store.SDKConfig.sessionTimeout;if(new Date>new Date(a._Store.dateLastEventSent.getTime()+d))b.endSession(),b.startNewSession();else {var e=a.Identity.getCurrentUser(),f=!1;if(e){// If userIdentities is an empty object but identifyRequest
```
alexs-mparticle, line 511: You can remove the dist files. We build them as part of our CI/CD process.

src/sessionManager.ts
```diff
@@ -45,9 +45,33 @@ export default function SessionManager(
                 self.endSession();
                 self.startNewSession();
             } else {
+                const user = mpInstance.Identity.getCurrentUser()
+                let needToIdentify = false
+
+                if (user) {
+                    // If userIdentities is an empty object but identifyRequest
```
alexs-mparticle, line 52: I think from an organizational standpoint, you should move this into our helpers.js or identity.js module, or make a new method within the SessionManager called `needsToIdentify`.

It would make testing easier and make things more compartmentalized.

src/sessionManager.ts
```diff
@@ -45,9 +45,33 @@ export default function SessionManager(
...
+                    // has at least a customerid, we will force the identify()
+                    // call, regardless of whether any user data is already
+                    // persisted.
+                    const storedUserIdentities = user
+                        .getUserIdentities()?.userIdentities
+                    const hasStoredCustomerId =
+                        storedUserIdentities != null &&
+                        typeof storedUserIdentities.customerid === "string"
```
alexs-mparticle, line 60: Is the purpose of checking if the customerid is as string to verify that it's present? I wonder if checking `typeof storedUserIdentities.customerid !== "undefined"` is more obvious?

We have a utility method called `isEmpty` that might work, but it was intended for objects/arrays. 

If you need to make sure it's a string, we have an `isString` utility method as well.

alexs-mparticle (changes requested):
This PR is a great start. Thanks for opening it. My main request is for you to write some tests around the new feature to make sure any dependent functionality isn't affected.

src/sessionManager.ts
deammer: I would prefer to keep this in TypeScript instead of JS so I'll go with your second suggestion 👍

deammer:
Hey @alexs-mparticle thank you for the review 🙏 I've made the changes you requested except for writing tests, which looks pretty involved based on a brief look I took. Is there any chance you or your team could take care of that?

PR #1015, opened 2025-04-02
feat: Add support for sandbox mode via selectPlacements

alexs-mparticle:
## Instructions

 ## Summary
 - Explicitly adds sandbox attribute when calling selectPlacements as an override while also supporting sandbox via RoktManager init
 - Replaces #1013

 ## Testing Plan
 - Attempt to send `selectPlacement` with a sandbox mode that is different than the default config.isDevelopmentMode setting.

 - Closes https://go.mparticle.com/work/SQDSDKS-7182

### DIFF
--- a/src/roktManager.ts
+++ b/src/roktManager.ts
@@ -45,6 +45,10 @@ export interface IRoktKit  {
     selectPlacements: (options: IRoktSelectPlacementsOptions) => Promise<IRoktSelection>;
 }
 
+export interface IRoktManagerOptions {
+    sandbox?: boolean;
+}
+
 // The purpose of this class is to create a link between the Core mParticle SDK and the
 // Rokt Web SDK via a Web Kit.
 // The Rokt Manager should load before the Web Kit and stubs out many of the
@@ -60,8 +64,9 @@ export default class RoktManager {
 
     private filteredUser: IMParticleUser | null = null;
     private messageQueue: IRoktMessage[] = [];
+    private sandbox: boolean | null = null;
 
-    public init(roktConfig: IKitConfigs, filteredUser?: IMParticleUser): void {
+    public init(roktConfig: IKitConfigs, filteredUser?: IMParticleUser, options?: IRoktManagerOptions): void {
         const { userAttributeFilters } = roktConfig || {};
 
         this.filters = {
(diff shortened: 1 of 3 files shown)


src/roktManager.ts
```diff
@@ -85,8 +92,23 @@ export default class RoktManager {
             return Promise.resolve({} as IRoktSelection);
         }
 
+        debugger;
+
```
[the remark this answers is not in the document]
crisryantan: Good bot 😄

src/roktManager.ts
```diff
@@ -85,8 +92,23 @@ export default class RoktManager {
...
         try {
-            return this.kit.selectPlacements(options);
+            let enrichedOptions = options;
+            const { attributes } = options;
+            const sandbox = attributes?.sandbox ?? this.sandbox;
```
rmi22186, line 100: `options` is typed as `IRoktManagerOptions`, which doesn't have an attributes key.  I think you just need to pull sandbox directly from options, as opposed to looking for it in attributes

crisryantan (approved):

test/jest/roktManager.spec.ts
```diff
@@ -215,5 +220,195 @@ describe('RoktManager', () => {
             expect(kit.selectPlacements).toHaveBeenCalledWith(options);
             expect(kit.launcher.selectPlacements).toHaveBeenCalledWith(options);
         });
+
+        it('should set sandbox to false in placement attributes when initialized as true', () => {
```
rmi22186, line 224: should this be `true`?
```suggestion
        it('should set sandbox to true in placement attributes when initialized as true', () => {
```

test/jest/roktManager.spec.ts
```diff
@@ -215,5 +220,195 @@ describe('RoktManager', () => {
...
+            };
+
+            roktManager.selectPlacements(options);
+
+            const expectedOptions = {
+                attributes: {
+                    customAttr: 'value',
+                    'sandbox': true
+                },
+                identifier: 'test-identifier'
```
rmi22186, line 408: nit - since expectedOptions is the same as options, you can just do 

```
const expectedOptions = options
```

rmi22186 (changes requested):

rmi22186 (approved):

PR #1018, opened 2025-04-07
feat: Add user attribute mapping to Rokt Manager

alexs-mparticle:
## Instructions

 ## Summary
 - Allows User Attribute Mapping via the Rokt Manager in the SDK.

 ## Testing Plan
 - Use the User Attribute Mapping via the Configuration Settings in the mParticle UI. Mapped User Attributes should appear with updated keys when making a `selectPlacements` call.

 - Closes https://go.mparticle.com/work/SQDSDKS-7153

### DIFF
--- a/src/roktManager.ts
+++ b/src/roktManager.ts
@@ -2,7 +2,7 @@ import { IKitConfigs } from "./configAPIClient";
 import { UserAttributeFilters  } from "./forwarders.interfaces";
 import { IMParticleUser } from "./identity-user-interfaces";
 import KitFilterHelper from "./kitFilterHelper";
-import { Dictionary, parseSettingsString } from "./utils";
+import { Dictionary, isEmpty, parseSettingsString } from "./utils";
 
 // https://docs.rokt.com/developers/integration-guides/web/library/attributes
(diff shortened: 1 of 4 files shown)


rmi22186 (changes requested):

src/roktManager.ts
```diff
@@ -97,6 +111,22 @@ export default class RoktManager {
         return Boolean(this.kit && this.kit.launcher);
     }
 
+    private mapUserAttributes(attributes: IRoktPartnerAttributes, userAttributeMapping: Dictionary<string>[]): IRoktPartnerAttributes {
```
BrandonStalnaker, line 114: So the logic seems fine I just had a harder time understanding it because IRoktPartnerAttributes. It's just an interface of a dictionary right? Does it do something useful right now or it more for supporting future changes?

alexs-mparticle: It just makes sure we conform to the expected interface from Rokt: https://docs.rokt.com/developers/integration-guides/web/library/attributes/

rmi22186 (approved):

rmi22186 (approved):

PR #1019, opened 2025-04-07
feat: Add select placement attributes to current user

alexs-mparticle:
## Instructions

 ## Summary
 - Adds selectPlacement Attributes to Current User Attributes

 ## Testing Plan
 - Verify that any select placement attributes appear within the current user's attributes
 - Should also trigger a user attribute change event if an attribute changes

 - Closes https://go.mparticle.com/work/SQDSDKS-7175
 - Depends on #1018 to be merged in first

### DIFF
--- a/src/mp-instance.ts
+++ b/src/mp-instance.ts
@@ -1394,7 +1394,7 @@ function completeSDKInitialization(apiKey, config, mpInstance) {
             mpInstance._IntegrationCapture.capture();
         }
 
-        // Configure Rokt Manager with filtered user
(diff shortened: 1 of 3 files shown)


rmi22186 (approved):
there exists an edge case where someone checks out, gets a placement, then logs out.  then a friend logs in and checks out (without refreshing).  the attributes would then still be on the previous current user since that one is never updated.  this feels like a very edge case to me, but something to consider if there is an easy solution or not.  my first instinct is it would require resetting the current user whenever login happens, but that is done kit side

PR #1021, opened 2025-04-17
feat: Add hashAttributes method to RoktManager

alexs-mparticle:
## Instructions

 ## Summary
 -  Add hashAttributes method to RoktManager 

 ## Testing Plan
- Verify direct passthrough of attributes when kit is fully initialized
- Verify null return when launcher exists but kit is not initialized
- Verify error logging when called before initialization
- Verify null return when kit is initialized but launcher is missing
- Verify return of hashed attributes from launcher

 - Closes https://go.mparticle.com/work/SQDSDKS-7220

### DIFF
--- a/src/roktManager.ts
+++ b/src/roktManager.ts
@@ -24,7 +24,7 @@ export interface IRoktSelection {
 
 export interface IRoktLauncher {
     selectPlacements: (options: IRoktSelectPlacementsOptions) => Promise<IRoktSelection>;
-    hashedAttributes: (attributes: IRoktPartnerAttributes) => Promise<Record<string, string>>;
+    hashAttributes: (attributes: IRoktPartnerAttributes) => Promise<Record<string, string>>;
 }
 
 export interface IRoktMessage {
@@ -44,7 +44,7 @@ export interface IRoktKit  {
     launcher: IRoktLauncher | null;
     userAttributes: Dictionary<string>;
     selectPlacements: (options: IRoktSelectPlacementsOptions) => Promise<IRoktSelection>;
-    hashedAttributes: (attributes: IRoktPartnerAttributes) => Promise<Record<string, string>>;
+    hashAttributes: (attributes: IRoktPartnerAttributes) => Promise<Record<string, string>>;
 }
 
 export interface IRoktManagerOptions {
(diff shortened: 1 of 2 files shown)


src/roktManager.ts
```diff
@@ -136,6 +138,22 @@ export default class RoktManager {
...
+        if (!this.isReady()) {
+            this.queueMessage({
+                methodName: 'hashedAttributes',
+                payload: attributes,
+            });
+            return {};
+        }
+
+        try {
+            return this.kit.hashedAttributes(attributes);
```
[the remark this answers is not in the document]
rmi22186: My understanding of this is that it's wrong because the partner will be calling the await, right?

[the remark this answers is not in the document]
alexs-mparticle: Correct. We are simply passing through the call and the promise it returns.

src/roktManager.ts
```diff
@@ -136,6 +138,22 @@ export default class RoktManager {
         }
     }
 
+    public hashAttributes(attributes: IRoktPartnerAttributes): Promise<Record<string, string>> {
+        if (!this.isReady()) {
+            this.queueMessage({
```
rmi22186, line 143: Since these aren't `void`s like `selectPlacements` is, I'm not sure that from a real life use case perspective this makes sense.  If you try calling `hashAttributes`, before `isReady()` is true, you get a resolved promise, and then the code continues on its way.  _Eventually_, the actual hashed attributes are resolved but can't be used since you're returned something already, right?

test/jest/roktManager.spec.ts
```diff
@@ -23,6 +23,129 @@ describe('RoktManager', () => {
         });
     });
 
+    describe('#hashAttributes', () => {
+        beforeEach(() => {
+            roktManager['currentUser'] = currentUser;
+        });
+
+        it('should call kit.hashAttributes with empty attributes', () => {
```
rmi22186, line 31: nit:

seems like an AI test that isn't relevant since the test directly below it is doing the same thing...just that in this test it's an empty object that's passed in and spit back out.

test/jest/roktManager.spec.ts
```diff
@@ -23,6 +23,129 @@ describe('RoktManager', () => {
...
+            const kit: Partial<IRoktKit> = {
+                launcher: {
+                    selectPlacements: jest.fn(),
+                    hashAttributes: jest.fn()
+                },
+
+                // We are mocking the hashAttributes method to return the
+                // launcher's hashAttributes method and verify that
+                // both the kit's and the launcher's methods
+                // are called with the correct attributes
```
rmi22186, line 129:
```suggestion
                // are called with the correct attributes.
```

rmi22186 (changes requested):

test/jest/roktManager.spec.ts
alexs-mparticle: My intention was to be explicit with the use cases of an empty object vs a populated object, especially if we decide in the future to return something other than an empty object, i.e. `null` or throw an error.

src/roktManager.ts
alexs-mparticle: I see your point.

I'm thinking we may want to revisit the `queueMessage` function, because I would assume that a user would call `hashAttributes` before `selectPlacements`.

rmi22186: yea - when the sdk loads on the home page and a user navigates through, it will 100% be available, but if a customer puts a placement on the homepage, then it might not and may necessitate something ilke an mparticle.Rokt.isReady function, which isn't great but may be necessary

rmi22186 (approved):

PR #1023, opened 2025-04-23
feat: Add method for logProductActionV2

rmi22186:
## Instructions

 ## Summary
`logProductAction` currently has a suboptimail signature because `transactionAttributes` comes after `customFlags` which can often be `null`.  This exists from 5 years ago when we deprecated [logCheckout/logPurchase/logRefund](https://github.com/mParticle/mparticle-web-sdk/commit/5ba57a3760667e875fc209806661db14b25df178) in favor of using logProductAction, but `logProductAction` didn't take TA at the time which the deprecated methods did, and so we had to add it to the tail end of the method signature.

This PR attempts to resolve this by allowing anything optional to be put in a single options object via `logProductActionV2`.  [Amplitude does something similar](https://amplitude.com/docs/sdks/analytics/browser/javascript-sdk#track-revenue) with `logRevenueV2`.  I have no issue changing this to a brand new method name.

 ## Testing Plan
 - [X] Was this tested locally? 
Tested in a test app locally - See screenshot below
image

 - Closes https://go.mparticle.com/work/SQDSDKS-7237

### DIFF
--- a/src/ecommerce.interfaces.ts
+++ b/src/ecommerce.interfaces.ts
@@ -84,6 +84,16 @@ export interface SDKECommerceAPI extends IECommerceShared {
         transactionAttributes?: TransactionAttributes,
         eventOptions?: SDKEventOptions
     ): void;
+    logProductActionV2(
+        productActionType: valueof<typeof ProductActionType>,
+        product: SDKProduct | SDKProduct[],
+        options?: {
+            attrs?: SDKEventAttrs;
+            customFlags?: SDKEventCustomFlags;
+            transactionAttributes?: TransactionAttributes;
+            eventOptions?: SDKEventOptions;
+        }
+    ): void;
     logPromotion(
         type: valueof<typeof PromotionActionType>,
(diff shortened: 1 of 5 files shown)


src/ecommerce.interfaces.ts
```diff
@@ -84,6 +84,16 @@ export interface SDKECommerceAPI extends IECommerceShared {
         transactionAttributes?: TransactionAttributes,
         eventOptions?: SDKEventOptions
     ): void;
+    logProductActionV2(
+        productActionType: valueof<typeof ProductActionType>,
+        product: SDKProduct | SDKProduct[],
+        options?: {
+            attrs?: SDKEventAttrs;
```
alexs-mparticle, line 91: I think that calling this `attrs` or `attributes` may be confusing since we have `attributes` as part of `selectPlacements`.

```suggestion
            eventAttributes?: SDKEventAttrs;
```

alexs-mparticle (changes requested):

PR #1025, opened 2025-04-23
feat: Add setExtensionData method to RoktManager

alexs-mparticle:
## Instructions

 ## Summary
 - Add setExtensionData method to RoktManager

 ## Testing Plan

 - Closes https://go.mparticle.com/work/SQDSDKS-7234

### DIFF
--- a/src/roktManager.ts
+++ b/src/roktManager.ts
@@ -9,6 +9,12 @@ export interface IRoktPartnerAttributes {
     [key: string]: string | number | boolean | undefined | null;
 }
 
+export interface IRoktExtensionData {
+    [extensionName: string]: {
+        [key: string]: string | number | boolean | undefined | null;
+    };
+}
+
 // https://docs.rokt.com/developers/integration-guides/web/library/select-placements-options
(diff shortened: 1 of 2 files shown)


rmi22186 (changes requested):
This makes sense so far.  Don't we need to have `mParticle.Rokt.setExtensionData` to be consumed by a developer?

rmi22186:
Nevermind I'm wrong.

src/roktManager.ts
```diff
@@ -9,6 +9,12 @@ export interface IRoktPartnerAttributes {
     [key: string]: string | number | boolean | undefined | null;
 }
 
+export interface IRoktExtensionData {
+    [extensionName: string]: {
+        [key: string]: string | number | boolean | undefined | null;
```
rmi22186, line 14: Where does the types from this come from?  Are you confident that it doesn't take arrays/dictionaries?

rmi22186 (approved):
MInor comment, but otherwise LGTM

rmi22186 (approved):

PR #1028, opened 2025-04-28
fix: Add Rokt hashAttributes and setExtensionData to snippet.js

alexs-mparticle:
## Instructions

 ## Summary
 - Add Rokt hashAttributes and setExtensionData to snippet.js

 ## Testing Plan
 - Verify methods are accessible before web SDK loads.

 - Closes https://go.mparticle.com/work/SQDSDKS-7245


rmi22186 (approved):

rmi22186 (approved):

rmi22186:
This is the right thing to do since the API is changing. We should add the extensions update here also since that is another public api change...although I don't see a public API change here...https://github.com/mParticle/mparticle-web-sdk/pull/1025

rmi22186 (approved):

PR #1031, opened 2025-05-12
fix: Remove case sesitivity from query params for integration captures

mmustafa-tse:
## Summary
 - The Snapchat team reported that their snapchat click id in url could come in as either `ScCid` or `sccid`, the following fix allows for both, or any other integration captures via url queries to ignore case sensitivity.

 ## Testing Plan
 - E2E tested, unit tested was not feasible since we cannot mock window.location.href

 - Closes https://go.mparticle.com/work/SQDSDKS-7286

### DIFF
--- a/src/integrationCapture.ts
+++ b/src/integrationCapture.ts
@@ -134,7 +134,7 @@ const integrationMapping: IntegrationIdMapping = {
 
     // Snapchat
     // https://businesshelp.snapchat.com/s/article/troubleshooting-click-id?language=en_US
-    ScCid: {
+    sccid: {
         mappedKey: 'SnapchatConversions.ClickId',
         output: IntegrationOutputs.CUSTOM_FLAGS,
     },
--- a/src/integrationCapture.ts
+++ b/src/integrationCapture.ts
@@ -134,7 +134,7 @@ const integrationMapping: IntegrationIdMapping = {
 
     // Snapchat
     // https://businesshelp.snapchat.com/s/article/troubleshooting-click-id?language=en_US
-    sccid: {
+    ScCid: {
         mappedKey: 'SnapchatConversions.ClickId',
         output: IntegrationOutputs.CUSTOM_FLAGS,
     },
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -274,7 +274,7 @@ const queryStringParser = (
 ): Dictionary<string> => {
     let urlParams: URLSearchParams | URLSearchParamsFallback;
     let results: Dictionary<string> = {};
-    let lowCaseUrlParams: Dictionary<string> = {};
+    let lowerCaseUrlParams: Dictionary<string> = {};
 
     if (!url) return results;
 
@@ -286,14 +286,14 @@ const queryStringParser = (
     }
 
     urlParams.forEach((value, key) => {
-        lowCaseUrlParams[key.toLowerCase()] = value;
+        lowerCaseUrlParams[key.toLowerCase()] = value;
     });
 
     if (isEmpty(keys)) {
-        return lowCaseUrlParams;
+        return lowerCaseUrlParams;
     } else {
         keys.forEach(key => {
-            const value = lowCaseUrlParams[key];
+            const value = lowerCaseUrlParams[key];
             if (value) {
                 results[key] = value;
             }
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -293,7 +293,7 @@ const queryStringParser = (
         return lowerCaseUrlParams;
     } else {
         keys.forEach(key => {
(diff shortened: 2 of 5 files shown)


src/integrationCapture.ts
```diff
@@ -134,7 +134,7 @@ const integrationMapping: IntegrationIdMapping = {
 
     // Snapchat
     // https://businesshelp.snapchat.com/s/article/troubleshooting-click-id?language=en_US
-    ScCid: {
+    sccid: {
```
rmi22186, line 137: If this generically allows any query parameter click id to be case insensitive, is there a reason for changing the casing here?  I'd expect the keys for the click ids to be the same.

src/utils.ts
```diff
@@ -274,6 +274,7 @@ const queryStringParser = (
 ): Dictionary<string> => {
     let urlParams: URLSearchParams | URLSearchParamsFallback;
     let results: Dictionary<string> = {};
+    let lowCaseUrlParams: Dictionary<string> = {};
```
rmi22186, line 277: nit: Spell the full `lowerCaseUrlParams` out instead of abbreviating

alexs-mparticle: +1

src/integrationCapture.ts
mmustafa-tse: this is where we compare what to expect coming in the url after lowercasing them, if I leave this as ScCid and we lowercase every click id, it will always fail since it will never match it

test/src/tests-integration-capture.ts
```diff
@@ -52,7 +52,7 @@ describe('Integration Capture', () => {
             rtid: '45670808',
             rclid: '7183717',
             wbraid: '1234111',
-            ScCid: '1234',
+            sccid: '1234',
```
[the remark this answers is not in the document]
mmustafa-tse: so this is the issue where I mentioned in standup that I can't accurately unit test this, this function mocks getQueryParams which after adding the lowerCaseUrlParams, should always be returned as lowercase, I'm not sure if there is a way where I can pass a dummy url with different cases of ScCid to accurately test this but our getHref function only looks into document.location.href, which I don't think I can mock

src/utils.ts
```diff
@@ -284,13 +285,15 @@ const queryStringParser = (
...
+
     if (isEmpty(keys)) {
-        urlParams.forEach((value, key) => {
-            results[key] = value;
-        });
+        return lowCaseUrlParams;
     } else {
         keys.forEach(key => {
-            const value = urlParams.get(key);
+            const value = lowCaseUrlParams[key];
```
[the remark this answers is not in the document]
mmustafa-tse: if I do it in the result, it will never be caught in that if logic since urlParams.get(ScCid) will fail since it will be looking for sccid instead

src/utils.ts
mmustafa-tse: done

src/utils.ts
```diff
@@ -284,13 +285,15 @@ const queryStringParser = (
...
 
+    urlParams.forEach((value, key) => {
+        lowerCaseUrlParams[key.toLowerCase()] = value;
+    });
+
     if (isEmpty(keys)) {
-        urlParams.forEach((value, key) => {
-            results[key] = value;
-        });
+        return lowerCaseUrlParams;
```
alexs-mparticle, line 293: This is going to have downstream errors. Technically speaking, `urlParams` isn't a dictionary that you can just swap out for `lowerCaseUrlParams`.

mmustafa-tse: before this line, we would assign results[key] = values from urlParams and return results later, I'm already doing that same logic in lines 288-291 expect I lower case them this time and return it when `keys` is empty, let me know if it still needs to be changed

test/jest/utils.spec.ts
```diff
@@ -173,6 +173,23 @@ describe('Utils', () => {
 
                 expect(queryStringParser(malformedUrl, keys)).toEqual(expectedResult);
             });
+
+            it('should handle different params case sensitivity and returns them as lowercased params', () => {
```
alexs-mparticle, line 177:
```suggestion
            it('should handle different params case sensitivity and return them as lowercased params', () => {
```

src/utils.ts
```diff
@@ -284,13 +285,15 @@ const queryStringParser = (
...
     } else {
         keys.forEach(key => {
-            const value = urlParams.get(key);
+            const value = lowerCaseUrlParams[key];
```
rmi22186, line 296: add .toLowerCase() to `key`, then revert the casing of sccid above and in the tests

rmi22186 (approved):

alexs-mparticle (approved):

PR #1032, opened 2025-05-20
feat: Fire identify in RoktManager if user email is different from selectPlacements email

rmi22186:
## Instructions

 ## Summary
If the email on the current user doesn't exist or is different from what is set on the `selectPlacements` call, we should call identify with all known user identities, including the email passed to `selectPlacements`

 ## Testing Plan
 - Locally tested in an app.

 - Closes https://go.mparticle.com/work/SQDSDKS-7249


alexs-mparticle (changes requested):

src/roktManager.ts
```diff
@@ -125,12 +140,38 @@ export default class RoktManager {
...
+            const currentUser = this.identity.getCurrentUser();
+            const currentUserIdentities = currentUser?.getUserIdentities()?.userIdentities || {};
+            const currentEmail = currentUserIdentities.email;
+            const newEmail = mappedAttributes.email as string;
+
+            // Check if email exists and differs
+            if (newEmail && (!currentEmail || currentEmail !== newEmail)) {
+
+                if (currentEmail && currentEmail !== newEmail) {
+                    this.logger.warning(
```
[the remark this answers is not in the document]
rmi22186: this sounds vaguely familiar...i can't seem to find it.  can you point me in the right direction?

src/mp-instance.ts
```diff
@@ -1394,7 +1394,18 @@ function completeSDKInitialization(apiKey, config, mpInstance) {
...
+            mpInstance._APIClient.prepareForwardingStats
+        );
+        mpInstance._Forwarders.processPixelConfigs(config);
+
+        // Checks if session is created, resumed, or needs to be ended
+        // Logs a session start or session end event accordingly
+        mpInstance._SessionManager.initialize();
+        mpInstance._Events.logAST();
+
+                // Configure Rokt Manager with user and filtered user
```
alexs-mparticle, line 1408:
```suggestion
    // Configure Rokt Manager with user and filtered user
```

src/roktManager.ts
```diff
@@ -111,7 +121,23 @@ export default class RoktManager {
         this.processMessageQueue();
     }
 
-    public selectPlacements(options: IRoktSelectPlacementsOptions): Promise<IRoktSelection> {
+    /**
+     * Selects placements based on the provided options.
+     * This method handles user identity updates and attribute mapping before selecting placements.
```
alexs-mparticle, line 126: I think this comment sounds vague but also misleading. Yes it does handle user identity updates and attribute mappings, but only as a side effect.

src/roktManager.ts
```diff
@@ -111,7 +121,23 @@ export default class RoktManager {
...
+     * 
+     * @param {IRoktSelectPlacementsOptions} options - The options for selecting placements, including attributes and optional identifier
+     * @returns {Promise<IRoktSelection>} A promise that resolves to the selected placements
```
alexs-mparticle, line 129: 🤖 🔫

src/roktManager.ts
```diff
@@ -111,7 +121,23 @@ export default class RoktManager {
...
+     * 
+     * @example
+     * // Correct usage with await
+     * await roktManager.selectPlacements({
```
alexs-mparticle, line 133: I think this might be the more accurate usage.
```suggestion
     * await window.mParticle.Rokt.selectPlacements({
```

src/roktManager.ts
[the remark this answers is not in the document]
alexs-mparticle: Now that I look at it, it's not a specific util for this case, but it's something we can extend to make a message generator: https://github.com/mParticle/mparticle-web-sdk/blob/64b272e985342e2e67642b47affa1e86e25523a5/src/utils.ts#L56

alexs-mparticle (changes requested):

src/roktManager.ts
```diff
@@ -68,24 +70,31 @@ export interface IRoktManagerOptions {
 export default class RoktManager {
     public kit: IRoktKit = null;
     public filters: RoktKitFilterSettings = {};
-
     private currentUser: IMParticleUser | null = null;
-    private filteredUser: IMParticleUser | null = null;
```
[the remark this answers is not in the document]
rmi22186: It's being set on filters.filteredUser, as opposed to the class. So we can remove it

BrandonStalnaker (commented):
identify logic looks correct to me. LGTM with Alex's approval

src/roktManager.ts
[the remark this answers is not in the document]
rmi22186: Since there are 2 variables being passed through, and it's a 1 off, i'm going to elect to not create a new utility method, and just use string interpolation instead.

src/mp-instance.ts
```diff
@@ -1406,20 +1417,15 @@ function completeSDKInitialization(apiKey, config, mpInstance) {
             const roktOptions: IRoktManagerOptions = {
                 sandbox: config.isDevelopmentMode,
             };
-            mpInstance._RoktManager.init(roktConfig, roktFilteredUser, currentUser, roktOptions);
+            mpInstance._RoktManager.init(
+                roktConfig,
+                roktFilteredUser,
+                mpInstance.Identity,
+                mpInstance.Logger,
```
alexs-mparticle, line 1424: This may be a topic for future discussion but I was thinking we could potentially pass in services as an object, similar to the other config options.

So this would theoretically become:

```
const services = {
    identityService: mpInstance.Identity,
    loggerService: mpInstance.Logger,
};

mpInstance._RoktManager.init(
    roktConfig,
    roktFilteredUser,
    services,
    roktOptions,
)
```

src/roktManager.ts
```diff
@@ -111,7 +121,22 @@ export default class RoktManager {
         this.processMessageQueue();
     }
 
-    public selectPlacements(options: IRoktSelectPlacementsOptions): Promise<IRoktSelection> {
+    /**
+     * Selects placements based on the provided options.
```
alexs-mparticle, line 125:
```suggestion
     * Renders a placement ad based on the options provided.
```

src/roktManager.ts
```diff
@@ -111,7 +121,22 @@ export default class RoktManager {
...
+     * 
+     * @param {IRoktSelectPlacementsOptions} options - The options for selecting placements, including attributes and optional identifier
+     * @returns {Promise<IRoktSelection>} A promise that resolves to the selected placements
```
alexs-mparticle, line 128:
```suggestion
     * @returns {Promise<IRoktSelection>} A promise that resolves to the selection
```

alexs-mparticle (changes requested):

src/mp-instance.ts
rmi22186: I like this idea.  I'm making a tech debt ticket - https://mparticle-eng.atlassian.net/browse/SQDSDKS-7339

test/jest/roktManager.spec.ts
```diff
@@ -769,6 +811,187 @@ describe('RoktManager', () => {
...
+                    hashAttributes: jest.fn()
+                },
+                selectPlacements: jest.fn().mockResolvedValue({}),
+                hashAttributes: jest.fn(),
+                setExtensionData: jest.fn(),
+            };
+
+            roktManager.kit = kit as IRoktKit;
+
+            const mockError = { error: 'Identity service error', code: 500 };
```
alexs-mparticle, line 967: I think it might be more likely to encounter 401 or 403, so we should also test for those.

alexs-mparticle (approved):

alexs-mparticle (approved):

PR #1033, opened 2025-05-20
feat: Expose Rokt launcher config options

alexs-mparticle:
## Instructions

 ## Summary
 - Exposes [Rokt integration launcher config options](https://docs.rokt.com/developers/integration-guides/web/library/integration-launcher-options). This will act as a pass-through for the Kit SDK to pass configuration options to the Rokt Web SDK.

 ## Testing Plan
 - Using a sample app, pass in various rokt configuration options to `window.mParticle.config` and verify that they appear in the Rokt Web SDK `selectPlacements` call.

 - Closes https://go.mparticle.com/work/SQDSDKS-7318

### DIFF
--- a/src/identity.js
+++ b/src/identity.js
@@ -1677,6 +1677,10 @@ export default function Identity(mpInstance) {
         }
         mpInstance._Store.isInitialized = true;
 
+        if (mpInstance._RoktManager.isReady()) {
+            mpInstance._RoktManager.currentUser = mpInstance.Identity.getCurrentUser();
+        }
+
         mpInstance._preInit.readyQueue = processReadyQueue(
             mpInstance._preInit.readyQueue
         );
--- a/src/roktManager.ts
+++ b/src/roktManager.ts
@@ -82,9 +82,8 @@ export default class RoktManager {
     private sandbox: boolean | null = null;
     private placementAttributesMapping: Dictionary<string>[] = [];
     private identityService: SDKIdentityApi;
-    private launcherOptions: IRoktLauncherOptions = {};
+    private launcherOptions?: IRoktLauncherOptions;
     private logger: SDKLoggerApi;
-
     /**
      * Initializes the RoktManager with configuration settings and user data.
      * 
--- a/src/roktManager.ts
+++ b/src/roktManager.ts
@@ -162,7 +162,7 @@ export default class RoktManager {
 
         try {
             const { attributes } = options;
-            const sandboxValue = attributes?.sandbox ?? this.sandbox;
+            const sandboxValue = attributes?.sandbox || null;
             const mappedAttributes = this.mapPlacementAttributes(attributes, this.placementAttributesMapping);
 
             // Get current user identities
--- a/src/roktManager.ts
+++ b/src/roktManager.ts
@@ -121,8 +121,9 @@ export default class RoktManager {
         };
 
         // This is the global setting for sandbox mode
-        // It is set here and passed in to the selectPlacements method
-        // but can be overridden if sandbox is manually set on the placement attributes by the partner
+        // It is set here and passed in to the createLauncher method in the Rokt Kit
+        // This is not to be confused for the `sandbox` flag in the selectPlacements attributes
+        // as that is independent of this setting, though they share the same name.
         this.sandbox = options?.managerOptions?.sandbox;
 
         // Launcher options are set here for the kit to pick up and pass through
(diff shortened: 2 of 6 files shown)


rmi22186 (commented):
This looks good, just minor comments

src/roktManager.ts
```diff
@@ -42,20 +42,27 @@ export interface RoktKitFilterSettings {
...
 }
 
-export interface IRoktManagerOptions {
-    sandbox?: boolean;
+export interface IRoktOptions {
+    managerOptions?: IRoktManagerOptions;
+    launcherOptions?: IRoktLauncherOptions;
 }
 
+export type IRoktLauncherOptions = Dictionary<any>;
```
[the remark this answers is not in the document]
alexs-mparticle: My thought was the `ManagerOptions` should be the ones we want to define and be strict about, especially in cases like `sandbox` which might have specific rules we need to consider. The `LauncherOptions` should just be a pass through to be maintainable.

test/jest/roktManager.spec.ts
```diff
@@ -224,7 +224,9 @@ describe('RoktManager', () => {
                 undefined,
                 mockMPInstance.Identity,
                 undefined,
-                { sandbox: true }
+                {
+                    managerOptions: { sandbox: true },
+                }
```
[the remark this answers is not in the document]
alexs-mparticle: Not really. We treat both `createLauncher` and `selectPlacements` sandbox flags as pass throughs.

src/identity.js
```diff
@@ -1677,6 +1677,10 @@ export default function Identity(mpInstance) {
         }
         mpInstance._Store.isInitialized = true;
 
+        if (mpInstance._RoktManager.isReady()) {
```
rmi22186, line 1680: So if it's not ready yet because identity returns faster than rokt manager gets initialized, this means there still won't be a current user right/. Is there a way to queue setting the current user for when it is ready?

src/roktManager.ts
```diff
@@ -113,7 +120,16 @@ export default class RoktManager {
             filteredUser: filteredUser,
         };
 
-        this.sandbox = options?.sandbox;
+        // This is the global setting for sandbox mode
+        // It is set here and passed in to the selectPlacements method
+        // but can be overridden if sandbox is set by the placement attributes
```
rmi22186, line 125:
```suggestion
        // but can be overridden if sandbox is manually set on the placement attributes by the partner
```

src/roktManager.ts
```diff
@@ -113,7 +120,16 @@ export default class RoktManager {
...
+        // It was initially set as part of the rokt launcher in the kit, but
+        // we are now only using it within the context of selectPlacements.
```
rmi22186, line 127:
```suggestion
```

rmi22186: can be removed as this is an outdate comment, right?

test/src/tests-identity.ts
```diff
@@ -4355,6 +4362,145 @@ describe('identity', function() {
...
+            fetchMockSuccess(urls.logout, {
+                context: null,
+                matched_identities: {
+                    device_application_stamp: 'my-das',
+                },
+                is_ephemeral: true,
+                mpid: testMPID,
+                is_logged_in: false,
+            });
+
```
rmi22186, line 4460: is there a reason to mock `urls.logout` here if you are not calling `logout` below?

test/src/tests-identity.ts
```diff
@@ -4355,6 +4362,145 @@ describe('identity', function() {
...
+            mParticle.init(apiKey, { ...window.mParticle.config, kitConfigs: [roktConfig] });
+
+            await waitForCondition(() => {
+                console.log('hasIdentifyReturned', mParticle.Identity.getCurrentUser()?.getMPID());
```
rmi22186, line 4464:
```suggestion
```

rmi22186 (changes requested):

rmi22186:
Food for thought - If we are passing sandbox to createLauncher now, is there a reason to include it in the `selectPlacements` call?

test/src/tests-identity.ts
alexs-mparticle: Nope. That's a bad copypasta

src/identity.js
alexs-mparticle: Actually is more of a check to see if the Rokt Kit is being used at all. We could potentially create a new method called `isConfigured` or `isActive`, but I felt like that might be unnecessary.

rmi22186: ah got it.  yea seems like it's more readable/grokable if you go with an approach like isActive. I think you can make it a tech debt ticket.

rmi22186 (changes requested):
~~My previous comment was actually that I didn't think we needed RoktManager.sandbox at all because we aren't using it anywhere, right?  What is its current purpose?~~

Ignore me!

rmi22186 (approved):

PR #1037, opened 2025-06-03
feat: Expose config.domain for easier CNAME configuration

rmi22186:
## Instructions

 ## Summary
A new config.domain will be added to the snippet.  When this happens, we will create the CNAME urls automatically as opposed to requiring a customer pass specific URLs to us, which is cumbersome.

 ## Testing Plan
 - Tested on Rokt customer site to use new snippet and confirmed URLs are mapped correctly
 - Tested locally pulling down app.js using the default `https://apps.rokt-api.com/` URL.  works correctly
 - Tested on other customer sites using our CURRENT snippet to ensure that data flows properly.  Confirmed on 1 site that has their own CNAME, 1 site that does not have a CNAME, 

 - Closes https://go.mparticle.com/work/SQDSDKS-7366

### DIFF
--- a/test/src/tests-store.ts
+++ b/test/src/tests-store.ts
@@ -1568,6 +1568,8 @@ describe('Store', () => {
             it('should prioritize domain over custom baseUrls when both are set', () => {
                 const config = {
                     domain: 'custom.domain.com',
+                    v1SecureServiceUrl: 'foo.customer.mp.com/v1/JS/',
+                    v2SecureServiceUrl: 'foo.customer.mp.com/v2/JS/',
                     v3SecureServiceUrl: 'foo.customer.mp.com/v3/JS/',
                     configUrl: 'foo-configUrl.customer.mp.com/v2/JS/',
                     identityUrl: 'foo-identity.customer.mp.com/',
--- a/test/src/tests-store.ts
+++ b/test/src/tests-store.ts
@@ -1530,13 +1530,13 @@ describe('Store', () => {
                 );
 
                 const expectedResult = {
-                    v3SecureServiceUrl: 'foo.customer.mp.com/v3/JS/',
                     configUrl: 'foo-configUrl.customer.mp.com/v2/JS/',
                     identityUrl: 'foo-identity.customer.mp.com/',
                     userAudienceUrl: 'foo-user-segment.customer.mp.com/',
                     aliasUrl: 'foo-alias.customer.mp.com/',
                     v1SecureServiceUrl: 'jssdks.mparticle.com/v1/JS/',
                     v2SecureServiceUrl: 'jssdks.mparticle.com/v2/JS/',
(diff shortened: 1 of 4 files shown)


test/src/tests-store.ts
```diff
@@ -1541,6 +1541,56 @@ describe('Store', () => {
...
+                };
+
+                const result = processBaseUrls(
+                    (config as unknown) as SDKInitConfig,
+                    (featureFlags as unknown) as IFeatureFlags,
+                    'apikey'
+                );
+
+                const expectedResult = {
+                    v3SecureServiceUrl: 'custom.domain.com/webevents/v3/JS/',
```
alexs-mparticle, line 1557: Nitpick: Can you group this with the other SecureServiceUrls?

test/src/tests-store.ts
```diff
@@ -1541,6 +1541,56 @@ describe('Store', () => {
...
+                    configUrl: 'custom.domain.com/tags/JS/v2/',
+                    identityUrl: 'custom.domain.com/identity/v1/',
+                    aliasUrl: 'custom.domain.com/webevents/v1/identity/',
+                    v1SecureServiceUrl: 'custom.domain.com/webevents/v1/JS/',
+                    v2SecureServiceUrl: 'custom.domain.com/webevents/v2/JS/',
```
alexs-mparticle, line 1589: Nitpick: Can you group this with the other SecureServiceUrls?

test/src/tests-store.ts
```diff
@@ -1541,6 +1541,56 @@ describe('Store', () => {
...
+                };
+
+                expect(result).to.deep.equal(expectedResult);
+            });
+
+            it('should prioritize domain over custom baseUrls when both are set', () => {
+                const config = {
+                    domain: 'custom.domain.com',
+                    v3SecureServiceUrl: 'foo.customer.mp.com/v3/JS/',
```
alexs-mparticle, line 1571:
```suggestion
                    v1SecureServiceUrl: 'foo.customer.mp.com/v1/JS/',
                    v2SecureServiceUrl: 'foo.customer.mp.com/v2/JS/',
                    v3SecureServiceUrl: 'foo.customer.mp.com/v3/JS/',
```

alexs-mparticle (changes requested):

test/src/tests-store.ts
```diff
@@ -1541,6 +1541,56 @@ describe('Store', () => {
 
                 expect(result).to.deep.equal(expectedResult);
             });
+
+            it('should append url paths to domain when config.domain is set', () => {
```
alexs-mparticle, line 1545: Can you add a test that will handle cnames that don't conform to these new routes, to verify they'll still work?

rmi22186: They will always conform to these routes (if what you mean is the portion after `.com/this/is/the/route/all/cname/customers/will/use`

alexs-mparticle (approved):

PR #1039, opened 2025-06-09
feat: Add domain to RoktManager to support CNAME for Rokt

rmi22186:
## Instructions

 ## Summary
This PR supports CNAME-ing for Rokt by adding this to the RoktManager.

 ## Testing Plan
Locally tested with a CNAME and building the kit  - PR for kit here - https://github.com/mparticle-integrations/mparticle-javascript-integration-rokt/pull/34

 - Closes https://go.mparticle.com/work/SQDSDKS-7387

### DIFF
--- a/src/roktManager.ts
+++ b/src/roktManager.ts
@@ -58,6 +58,7 @@ export interface IRoktKit {
 export interface IRoktOptions {
     sandbox?: boolean;
     launcherOptions?: IRoktLauncherOptions;
+    domain?: string;
 }
 
 export type IRoktLauncherOptions = Dictionary<any>;
@@ -81,6 +82,7 @@ export default class RoktManager {
     private identityService: SDKIdentityApi;
     private launcherOptions?: IRoktLauncherOptions;
     private logger: SDKLoggerApi;
+    private domain?: string;
     /**
      * Initializes the RoktManager with configuration settings and user data.
      * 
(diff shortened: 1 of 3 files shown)


src/roktManager.ts
```diff
@@ -126,6 +128,8 @@ export default class RoktManager {
         // Launcher options are set here for the kit to pick up and pass through
         // to the Rokt Launcher.
         this.launcherOptions = { sandbox, ...options?.launcherOptions };
+
+        this.domain = options?.domain;
```
alexs-mparticle, line 132: Our instructions usually reference domain to include the protocol, for example `const ROKT_DOMAIN = "https://apps.rokt-api.com";`. I think we should expect that a user might pass in a domain with and without a protocol, we should wrap some smarts around this to strip that out.

alexs-mparticle, line 132: Nit: Not sure if we should attach `domain` if we don't have to.

```suggestion
        if (options?.domain) {
            this.domain = options?.domain;
        }    
```

alexs-mparticle (changes requested):

src/roktManager.ts
rmi22186: In [the snippet](https://docs.rokt.com/developers/integration-guides/getting-started/ecommerce/ecommerce-sdk-integration), it actually says

```
mParticle.config.domain = ROKT_DOMAIN.split('//')[1]
```

to account for this

rmi22186: updating slightly, but will include

alexs-mparticle (approved):

PR #1041, opened 2025-06-13
fix: Revise RoktManager promise queuing logic

alexs-mparticle:
## Instructions

 ## Summary
 - When we initially implemented the Rokt Manager, we added logic to queue Rokt methods that were called before the mParticle SDK was ready.
 - In some cases, there was a race condition that would occur where the SDK was ready to queue methods (via the ready queue) but the Rokt Manager was not ready because the launcher was not attached. In this scenario, the promise would get dropped and queued asynchronous methods would not fire.
 - In some rare cases, the promise would execute the associated Rokt method, but the promise would not be returned and event listeners would not be properly invoked.
 - This new change will store queued promises as promises (rather than a simple array of method calls) and replay them when the kit is attached.
 - This change will also set a timeout value to force a lost promise to reject which should help us with troubleshooting.
 - We've also added better logging and edge case handling.

 ## Testing Plan
 - Using a sample app, set the SDK script tag to be deferred or introduce a setTimeout to delay the loading of the Web SDK.
 - Attempt to fire a `selectPlacement` or `hashAttributes` call.
 - The method should be queued properly and execute once the kit is attached.
 - *Note*: Using a Tag Manager like GTM or Adobe Extensions will allow easier testing by triggering a selectPlacement at the same time as the SDK initialization.

 - Closes https://go.mparticle.com/work/SQDSDKS-7404

### DIFF
--- a/src/roktManager.ts
+++ b/src/roktManager.ts
@@ -34,6 +34,7 @@ export interface IRoktLauncher {
 }
 
 export interface IRoktMessage {
+    messageId?: string;
     methodName: string;
     payload: any;
 }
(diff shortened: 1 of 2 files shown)


alexs-mparticle:
closed in favor of #1051

PR #1046, opened 2025-07-17
fix: fire identify call if config identities differ from current user

rmi22186:
## Instructions

 ## Summary
Current user flow for an MPA:
* Developer initializes MP with no identities.  identify request is called
* Developer then adds a known identity to the mp initialization script.  When the page reloads, an identify request is not called because the SDK only looks to see if there is an existing session
New behavior:
* When a new identity is added to the initialization script, we check to see if there is any difference between the current user's identities.  If so, make an identify call.

 ## Testing Plan
 - Added integration tests and local tests

Videos are below of before and after behavior

1. [Previous behavior](https://share.zight.com/4guEDjxq) (identify call only made on the initial page load, not after a page refresh with new identities in the initialization script0
2. [New behavior](https://share.zight.com/4guEDjyq) (identify calls are made on the initial page load, then after identities are changed in the initialization script0

 - Closes https://go.mparticle.com/work/SQDSDKS-7504

### DIFF
--- a/src/identity-utils.ts
+++ b/src/identity-utils.ts
@@ -13,7 +13,6 @@ import {
     IIdentityResponse,
     IMParticleUser,
 } from './identity-user-interfaces';
-import { IMParticleWebSDKInstance } from './mp-instance';
 
 const { Identify, Modify, Login, Logout } = Constants.IdentityMethods;
 export const CACHE_HEADER = 'x-mp-max-age' as const;
--- a/src/identity-utils.ts
+++ b/src/identity-utils.ts
@@ -7,10 +7,11 @@ import {
     UserIdentities,
     IdentityCallback,
 } from '@mparticle/web-sdk';
-import { IdentityAPIMethod } from './identity.interfaces';
+import { IdentityAPIMethod, IIdentityRequest } from './identity.interfaces';
 import {
     IdentityResultBody,
     IIdentityResponse,
+    IMParticleUser,
 } from './identity-user-interfaces';
 import { IMParticleWebSDKInstance } from './mp-instance';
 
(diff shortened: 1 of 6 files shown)


alexs-mparticle (changes requested):

src/sessionManager.ts
```diff
@@ -62,7 +63,47 @@ export default function SessionManager(
         }
     };
 
-    this.getSession = function(): string {
+    this.hasIdentityRequestChanged = (): boolean => {
```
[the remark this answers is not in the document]
rmi22186: agreed

src/identity-utils.ts
```diff
@@ -279,3 +280,28 @@ const getExpireTimestamp = (maxAge: number = ONE_DAY_IN_SECONDS): number =>
...
+
+    const newIdentityRequest = mpInstance._Store.SDKConfig.identifyRequest;
+
+    if (
+        !currentUser ||
+        !newIdentityRequest ||
+        !newIdentityRequest.userIdentities
+    ) {
+        return false;
+    }
```
alexs-mparticle, line 297:
```suggestion
    if (!currentUser || !newIdentityRequest?.userIdentities) {
        return false;
    }
```

or if you really wanna make it a one liner:

`return !!(currentUser && newIdentityRequest?.userIdentities);`

src/identity-utils.ts
```diff
@@ -279,3 +280,28 @@ const getExpireTimestamp = (maxAge: number = ONE_DAY_IN_SECONDS): number =>
 
 const parseIdentityResponse = (responseText: string): IdentityResultBody =>
     responseText ? JSON.parse(responseText) : ({} as IdentityResultBody);
+
+export const hasIdentityRequestChanged = (
+    mpInstance: IMParticleWebSDKInstance
+): boolean => {
+    const currentUser = mpInstance.Identity.getCurrentUser();
+
+    const newIdentityRequest = mpInstance._Store.SDKConfig.identifyRequest;
```
alexs-mparticle, line 289: Let's avoid bringing in `mpInstance` if we can.

```suggestion
export const hasIdentityRequestChanged = (
    currentUser: MPUser, newIdentityRequest: IdentityRequest
): boolean => {
```

Then we can just set these outside the call:
```
    hasIdentityRequestChanged(
        mpInstance.Identity.getCurrentUser,
        mpInstance._Store.SDKConfig.identifyRequest;
    );
```

The benefit is that we now don't need to inject the whole SDK and testing should be easier.

alexs-mparticle (changes requested):

src/identity-utils.ts
```diff
@@ -7,11 +7,13 @@ import {
...
     IdentityCallback,
 } from '@mparticle/web-sdk';
-import { IdentityAPIMethod } from './identity.interfaces';
+import { IdentityAPIMethod, IIdentityRequest } from './identity.interfaces';
 import {
     IdentityResultBody,
     IIdentityResponse,
+    IMParticleUser,
 } from './identity-user-interfaces';
+import { IMParticleWebSDKInstance } from './mp-instance';
```
alexs-mparticle, line 16: Remove

```suggestion
```

alexs-mparticle (changes requested):

alexs-mparticle (approved):

PR #1051, opened 2025-08-05
fix: Refactor RoktManager with deferred call handling and message queue processing

alexs-mparticle:
## Instructions

 ## Summary
 - Refactors the queueMessage logic of the Rokt Manager to defer promises for Rokt functions so that consumers can properly await the results.

 ## Testing Plan
 - Using a sample web app, make a call to `selectPlacements` that triggers before the mParticle web sdk finishes loading the Rokt Web Kit. The call should return placements correctly after queuing.

As an example:

```javascript
mParticle.ready( async function () {
  try {
    const selection = await window.mParticle.Rokt.selectPlacements({
        identifier: `mparticle-layout-test`,
        attributes: {
            email: 'test@email.com',
            firstname: 'Kain',
            lastname: 'Marko',
        }
    });
    selection.on('OFFER_ENGAGEMENT').subscribe(function (data) {
        console.log('offer engagement fired', data);
    });
  } catch (error) {
    console.error('Error selectPlacements:', error);
  }
});
```

 - Closes https://go.mparticle.com/work/SQDSDKS-7404

### DIFF
--- a/test/jest/roktManager.spec.ts
+++ b/test/jest/roktManager.spec.ts
@@ -125,9 +125,10 @@ describe('RoktManager', () => {
             roktManager.hashAttributes(attributes);
 
             expect(roktManager['kit']).toBeNull();
-            expect(roktManager['messageQueue'].length).toBe(1);
-            expect(roktManager['messageQueue'][0].methodName).toBe('hashAttributes');
-            expect(roktManager['messageQueue'][0].payload).toBe(attributes);
+            expect(roktManager['messageQueue'].size).toBe(1);
+            const queuedMessage = roktManager['messageQueue'].values().next().value;
+            expect(queuedMessage.methodName).toBe('hashAttributes');
+            expect(queuedMessage.payload).toBe(attributes);
         });
 
         it('should process queued hashAttributes calls once the launcher and kit are attached', () => {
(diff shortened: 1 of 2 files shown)


test/jest/roktManager.spec.ts
```diff
@@ -1027,4 +1039,120 @@ describe('RoktManager', () => {
...
+            roktManager['completePendingPromise'](messageId, Promise.reject(error));
+
+            // Promise should reject with the error
+            await expect(promise).rejects.toThrow('test error');
+
+            // Should clean up the pending promise
+            expect(roktManager['pendingPromises'].has(messageId)).toBe(false);
+        });
+
+        it('should handle async results correctly', async () => {
```
rmi22186, line 1129: Can this be combined with the first test?  Seems like it's mostly doing the same thing...but adding a few extra checks.

test/jest/roktManager.spec.ts
```diff
@@ -444,9 +446,10 @@ describe('RoktManager', () => {
             roktManager.selectPlacements(options);
 
             expect(roktManager['kit']).toBeNull();
-            expect(roktManager['messageQueue'].length).toBe(1);
-            expect(roktManager['messageQueue'][0].methodName).toBe('selectPlacements');
-            expect(roktManager['messageQueue'][0].payload).toBe(options);
+            expect(roktManager['messageQueue'].size).toBe(1);
+            const queuedMessage = roktManager['messageQueue'].values().next().value;
```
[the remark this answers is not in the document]
rmi22186: here and elsewhere where you have `.next`

test/jest/roktManager.spec.ts
rmi22186: i was referring to not the test above, but the first test `should resolve pending promise with success result`

test/jest/roktManager.spec.ts
```diff
@@ -1027,4 +1049,125 @@ describe('RoktManager', () => {
             }).toThrow('Error setting extension data: ' + mockError.message);
         });
     });
+
+    describe('#deferredCall', () => {
+        it.only('should create a deferred promise with unique messageId', () => {
```
rmi22186, line 1054:
```suggestion
        it('should create a deferred promise with unique messageId', () => {
```

rmi22186 (approved):
tested locally and looks good.  great work!

this will require a docs update.  we need to let customers know to wrap selectplacement in `mParticle.ready` here https://docs.rokt.com/developers/integration-guides/getting-started/ecommerce/ecommerce-sdk-integration#5-show-a-placement

PR #1053, opened 2025-08-12
feat: added only rokt, all, none click ids when collectClickIdV2Enabled is enabled

jaissica12:
## Instructions

 ## Summary
 - Added only rokt, all and none as options when collectClickIdV2Enabled is enabled. 
 - default mode is none

 ## Testing Plan

 - Closes https://mparticle-eng.atlassian.net/browse/SQDSDKS-7536


src/constants.ts
```diff
@@ -184,6 +184,8 @@ const Constants = {
         CacheIdentity: 'cacheIdentity',
         AudienceAPI: 'audienceAPI',
         CaptureIntegrationSpecificIds: 'captureIntegrationSpecificIds',
+        CollectClickIdV2Enabled: 'collectClickIdV2Enabled',
+        IntegrationCaptureMode: 'integrationCaptureMode',
```
rmi22186, line 188: No need to have this in the feature flag constants since that should only be keys coming back from the server.  Move this to where you are using it.

src/integrationCapture.ts
```diff
@@ -339,4 +350,26 @@ export default class IntegrationCapture {
         }
         return filteredMappings;
     }
+
+    /**
+     * Filters a set of key/value IDs to only include Rokt-related IDs
+     */
+    private filterToRokt(map: Dictionary<string>): Dictionary<string> {
```
rmi22186, line 357: This function is defined but not being used anywhere

src/constants.ts
rmi22186: I also realized that although these are classified as FeatureFlags, they aren't necessarily ALL feature flags. It is starting to become a catchall for both feature flags AND web input specific settings that come down from the server.

Not a concern for your purposes right now, but I also wanted to note that `collectClickIdV2Enabled` is not what comes from the server. It's '`captureIntegrationSpecificIds.V2'` that you should be looking for, which has values of 'all', 'none', or 'roktonly'
```suggestion
        CaptureIntegrationSpecificIdsV2: 'captureIntegrationSpecificIds.V2',
```

src/store.ts
```diff
@@ -736,6 +741,13 @@ export function processFlags(config: SDKInitConfig): IFeatureFlags {
     flags[AudienceAPI] = config.flags[AudienceAPI] === 'True';
     flags[CaptureIntegrationSpecificIds] = config.flags[CaptureIntegrationSpecificIds] === 'True';
     flags[AstBackgroundEvents] = config.flags[AstBackgroundEvents] === 'True';
+    // default CollectClickIdV2Enabled to false when undefined
+    const collectV2 = config.flags[CollectClickIdV2Enabled] === 'True';
+    flags[CollectClickIdV2Enabled] = collectV2;
+    // Normalize mode to lower-case with default 'none'
+    const rawMode = (config.flags[IntegrationCaptureMode] || 'none').toString().trim().toLowerCase();
+    // Only meaningful when v2 is enabled; otherwise legacy boolean is used elsewhere
+    flags[IntegrationCaptureMode] = rawMode;
```
rmi22186, line 750: Based on all the above comments and the fact that we are only using `captureIntegrationSpecificIds.V2`, I think we can get rid of a lot of this code and just have:

```suggestion
    flags['captureIntegrationSpecificIds.V2'] = config.flags['captureIntegrationSpecificIds.V2'] || null
```

src/integrationCapture.ts
```diff
@@ -160,9 +161,15 @@ export default class IntegrationCapture {
...
      */
     public capture(): void {
-        const queryParams = this.captureQueryParams() || {};
-        const cookies = this.captureCookies() || {};
-        const localStorage = this.captureLocalStorage() || {};
+        let queryParams = this.captureQueryParams() || {};
+        let cookies = this.captureCookies() || {};
+        let localStorage = this.captureLocalStorage() || {};
+
+        if (this.captureMode === 'None' || !this.captureMode) {
```
rmi22186, line 168: I would add the capture modes as exported constants to the src/constants.ts file under a new object called `IntegrationCaptureModes` or something like that.  This way they can be referenced from `mp-instance.ts` and `integrationCapture.ts`

src/serverModel.ts
```diff
@@ -277,16 +277,23 @@ export default function ServerModel(
...
+                    const transformedClickIDs = mpInstance._IntegrationCapture.getClickIdsAsCustomFlags();
+                    customFlags = { ...transformedClickIDs, ...customFlags };
+                    const transformedIntegrationAttributes = mpInstance._IntegrationCapture.getClickIdsAsIntegrationAttributes();
+                    integrationAttributes = { ...transformedIntegrationAttributes, ...integrationAttributes };
+                } else if (mpInstance._Helpers.getFeatureFlag(FeatureFlags.CaptureIntegrationSpecificIds)) {
+                    mpInstance._IntegrationCapture.capture();
+                    const transformedClickIDs = mpInstance._IntegrationCapture.getClickIdsAsCustomFlags();
+                    customFlags = { ...transformedClickIDs, ...customFlags };
+                    const transformedIntegrationAttributes = mpInstance._IntegrationCapture.getClickIdsAsIntegrationAttributes();
+                    integrationAttributes = { ...transformedIntegrationAttributes, ...integrationAttributes };
```
rmi22186, line 295: I think this can be a single if statement and simplified.  lines 285-289 are idential to 291-295, unless i'm mistaken. 

We don't need to check the normalized values, then check the other feature flag, and run the same logic.  You can check it all once with an or statement, and then run through lines 285-289 once.

rmi22186 (changes requested):

src/mp-instance.ts
```diff
@@ -1390,7 +1390,18 @@ function completeSDKInitialization(apiKey, config, mpInstance) {
             mpInstance._ForwardingStatsUploader.startForwardingStatsTimer();
         }
 
-        if (mpInstance._Helpers.getFeatureFlag(CaptureIntegrationSpecificIds)) {
+        const isCaptureIntegrationSpecificIdsV2 = mpInstance._Helpers.getFeatureFlag(CaptureIntegrationSpecificIdsV2) as string;
+        if (typeof isCaptureIntegrationSpecificIdsV2 === 'string' && isCaptureIntegrationSpecificIdsV2.length > 0) {
+            const normalizedValue = isCaptureIntegrationSpecificIdsV2.toLowerCase();
+            if (normalizedValue === 'roktonly') {
+                mpInstance._IntegrationCapture.captureMode = 'RoktOnly';
+                mpInstance._IntegrationCapture.capture();
```
alexs-mparticle, line 1398: You don't need to include `mpInstance._IntegrationCapture.capture(); in each conditional. It may just be cleaner if you call it once at the end of the conditionals.

src/constants.ts
```diff
@@ -184,6 +184,7 @@ const Constants = {
         CacheIdentity: 'cacheIdentity',
         AudienceAPI: 'audienceAPI',
         CaptureIntegrationSpecificIds: 'captureIntegrationSpecificIds',
+        CaptureIntegrationSpecificIdsV2: 'captureIntegrationSpecificIds.V2',
```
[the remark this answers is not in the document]
jaissica12: Do you recommend adding it over any specific function, where it might be more helpful in `integrationCapture`

src/mp-instance.ts
```diff
@@ -1389,9 +1389,23 @@ function completeSDKInitialization(apiKey, config, mpInstance) {
         if (mpInstance._Helpers.getFeatureFlag(ReportBatching)) {
             mpInstance._ForwardingStatsUploader.startForwardingStatsTimer();
         }
+        const integrationSpecificIds = mpInstance._Helpers.getFeatureFlag(CaptureIntegrationSpecificIds) as boolean;
+        const integrationSpecificIdsV2 = mpInstance._Helpers.getFeatureFlag(CaptureIntegrationSpecificIdsV2) as string | null;
```
alexs-mparticle, line 1393:
```suggestion
        const { getFeatureFlag } = mpInstance._Helpers;
        const integrationSpecificIds = getFeatureFlag(CaptureIntegrationSpecificIds) as boolean;
        const integrationSpecificIdsV2 = getFeatureFlag(CaptureIntegrationSpecificIdsV2) as string | null;
```

src/mp-instance.ts
```diff
@@ -1389,9 +1389,23 @@ function completeSDKInitialization(apiKey, config, mpInstance) {
...
+        
+        const isIntegrationCaptureEnabled = (integrationSpecificIdsV2 ? integrationSpecificIdsV2 !== CaptureIntegrationSpecificIdsV2Modes.None : false) || (integrationSpecificIds === true);
+
+        if (isIntegrationCaptureEnabled) {
+            let captureMode: 'All' | 'RoktOnly' | undefined;
+            if (integrationSpecificIdsV2 === CaptureIntegrationSpecificIdsV2Modes.RoktOnly) {
+                captureMode = 'RoktOnly';
+            } else if (integrationSpecificIdsV2 === CaptureIntegrationSpecificIdsV2Modes.All || integrationSpecificIds === true) {
+                captureMode = 'All';
+            }
```
alexs-mparticle, line 1403: If the feature flag values are coming in as strings, then do we really need to define a new capture mode string? Can't we just use the original value for the selection?

I think that would simplify things.

alexs-mparticle (changes requested):

src/mp-instance.ts
jaissica12: Thanks for the feedback! We are also also using `integrationSpecificIds` which is boolean type utilized to set the captureMode as `'all'` if its `True`

dist/mparticle.js
```diff
@@ -203,7 +203,7 @@ var mParticle = (function () {
       Base64: Base64$1
     };
 
-    var version = "2.43.1";
+    var version = "2.43.3";
```
alexs-mparticle, line 206: Please don't commit the dist folder

package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
     "name": "@mparticle/web-sdk",
-    "version": "2.43.1",
+    "version": "2.43.3",
```
alexs-mparticle, line 3: I think you may need to rebase your source branch to make sure have the latest changes. These should not be in your PR.

jaissica12: Thanks for the feedback, I have rebased my source branch

PR #1055, opened 2025-08-19
feat: Migrate Logger to TypeScript

mmustafa-tse:
## Summary
 - Migrate Logger from JS to TS

 ## Testing Plan
 - E2E and unit tested

 - Closes https://go.mparticle.com/work/SQDSDKS-6773


src/logger.ts
```diff
@@ -1,21 +1,32 @@
-function Logger(config) {
+import { SDKInitConfig } from "./sdkRuntimeModels";
+
+interface Logger {
+    verbose: (msg: string) => void;
+    warning: (msg: string) => void;
+    error: (msg: string) => void;
+    setLogLevel: (msg: string) => void;
```
alexs-mparticle, line 7:
```suggestion
    setLogLevel: (logLevel: string) => void;
```

src/logger.ts
```diff
@@ -1,21 +1,32 @@
...
+    logLevel: string;
+    logger: this;
+}
+
+function Logger(this: Logger, config: SDKInitConfig): void {
     var self = this;
-    var logLevel = config.logLevel || 'warning';
+    var logLevel: string = config.logLevel || 'warning';
```
alexs-mparticle, line 14: You should use `let` and `const` in this file instead of `var`.

src/logger.ts
```diff
@@ -26,36 +37,37 @@ function Logger(config) {
...
             if (self.logger.error) {
                 self.logger.error(msg);
             }
         }
     };
 
-    this.setLogLevel = function(newLogLevel) {
+    this.setLogLevel = function(newLogLevel: string) {
         logLevel = newLogLevel;
+        this.logLevel = logLevel;
```
alexs-mparticle, line 50: I think you should be able to get rid of the redundant assignment and simplify this:

```
this.setLogLevel = function(_logLevel: string) {
        this.logLevel = _logLevel;
```

src/logger.ts
alexs-mparticle, line 7: Also, you should create a type definition for log level. For example:

```
type LogLevelType = 'none' | 'verbose' | 'warning' | 'error';
```

alexs-mparticle (changes requested):

rmi22186:
Since https://github.com/mParticle/mparticle-web-sdk/pull/1120 was merged, this can be closed, right @gtamanaha ?

PR #1059, opened 2025-08-26
feat: added noFunctional and noTargeting flags

jaissica12:
## Instructions

 ## Summary
 - Added two privacy flags - `noTargeting` `noFunctional` which will be extracted from `config` and used further to evaluate local storage / cookie. 

 ## Testing Plan
 - [X] Was this tested locally? If not, explain why. - Tested locally through testing app

 - Closes https://rokt.atlassian.net/jira/software/c/projects/SDKE/boards/1539?assignee=712020%3Aa3f11149-8d5c-4d52-8be3-57a9dbbeaea0&selectedIssue=SDKE-69


test/jest/store.flags.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+            },
+        } as any;
+
+        store = {} as any;
+        // Construct Store with our mock 'this'
+        (Store as any).call(store, {} as SDKInitConfig, mockMPInstance, 'apikey');
+    });
+
+    // Default case
+    it('defaults to false when not provided', () => {
```
rmi22186, line 41: i'd be more specific


```suggestion
    it('defaults noFunctional and noTargeting to false when not provided', () => {
```

rmi22186 (commented):
minor nits

PR #1060, opened 2025-08-27
feature: Map events to an attribute flag for selectPlacements

alexs-mparticle:
## Instructions

 ## Summary

 ## Testing Plan

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -153,7 +153,7 @@ export interface IStore {
     isEnabled: boolean;
     isInitialized: boolean;
 
-    // Session Attributes are presistant attributes that are tied to the current session and
+    // Session Attributes are persistent attributes that are tied to the current session and
     // are uploaded then cleared when the session ends.
     sessionAttributes: SessionAttributes;
 
--- a/test/jest/roktManager.spec.ts
+++ b/test/jest/roktManager.spec.ts
@@ -281,7 +281,8 @@ describe('RoktManager', () => {
                 kitConfig as IKitConfigs,
                 {} as IMParticleUser,
                 mockMPInstance.Identity,
-                mockMPInstance._Store
+                mockMPInstance._Store,
+                mockMPInstance.Logger,
             );
             expect(roktManager['placementAttributesMapping']).toEqual([
                 { 


src/apiClient.ts
```diff
@@ -148,6 +148,9 @@ export default function APIClient(
             // We need to check event again, because kitblocking
             // can nullify the event
             if (event) {
+                if (mpInstance._RoktManager.isReady()) {
+                    mpInstance._RoktManager.processEvent(event);
+                }
```
rmi22186, line 153: Was there a reason to put this inside of the logic for kit blocker? There is a tiny edge case where a customer may want to fire the event for the purposes of suppressing the ad, but also block the event from being sent to a kit.  I'd suggest moving this out of this if block and towards the top of this function

src/kitFilterHelper.ts
```diff
@@ -11,6 +11,10 @@ export default class KitFilterHelper {
         return generateHash(eventType + eventName);
     };
 
+    static hashEventMessage(eventName: string, eventType: valueof<typeof EventType>, messageType: valueof<typeof MessageType>): number {
+        return generateHash(String(messageType) + String(eventType) + eventName);
```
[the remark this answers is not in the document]
rmi22186: Tested and here is proof that it works as intended:

Image

src/store.ts
```diff
@@ -122,6 +122,7 @@ function createSDKConfig(config: SDKInitConfig): SDKConfig {
 //       to TypeScript
 export type ServerSettings = Dictionary;
 export type SessionAttributes = Dictionary;
+export type SessionSelectionAttributes = Dictionary;
```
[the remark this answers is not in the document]
rmi22186: Ties a bit more nicely with what you've defined [here](https://github.com/mParticle/mparticle-web-sdk/pull/1060/files#diff-7cb37e3efb8e0c8ba283a333407dd58bed345121a89f3b82014aabf2048cc61cR94-R95) as well

src/utils.ts
```diff
@@ -160,7 +160,14 @@ const parseNumber = (value: string | number): number => {
     return isNaN(floatValue) ? 0 : floatValue;
 };
 
-const parseSettingsString = (settingsString: string): Dictionary<string>[] => {
+interface SettingMappingElement {
```
rmi22186, line 163: Element is a very vague term and doesn't add any value to defining what it is

Maybe the following?

```suggestion
interface EventToPlacementKeyMapping {
```

src/roktManager.ts
```diff
@@ -101,19 +113,30 @@ export default class RoktManager {
...
             this.placementAttributesMapping = parseSettingsString(placementAttributesMapping);
         } catch (error) {
             console.error('Error parsing placement attributes mapping from config', error);
         }
 
+        try {
+            const placementEventMappingArray: SettingMappingElement[] = parseSettingsString(placementEventMapping);
+            this.placementEventMappingLookup = this.generateMappedEventLookup(placementEventMappingArray);
+        } catch (error) {
+            console.error('Error parsing placement event mapping from config', error);
```
rmi22186, line 135: Use the MP Logger:


```suggestion
            logger.error('Error parsing placement event mapping from config', error);
```

src/kitFilterHelper.ts
```diff
@@ -11,6 +11,10 @@ export default class KitFilterHelper {
         return generateHash(eventType + eventName);
     };
 
+    static hashEventMessage(eventName: string, eventType: valueof<typeof EventType>, messageType: valueof<typeof MessageType>): number {
```
rmi22186, line 14: Technically the kitfilterhelpers are all related to side loaded kits. (This can probably be documented better in this file). You'll see them only ever used in the `sideloadedKit.ts` file.

I can see why you put hashEventMessage here, because there's a lot of other hashing going on here.  But I think it belongs in `utils.ts` instead.

rmi22186: Docs about usage for the public api that ends up using the [KitFilterHelper](https://docs.mparticle.com/developers/client-sdks/web/kits/#filtering-data) as an FYI.

src/roktManager.ts
```diff
@@ -142,6 +163,21 @@ export default class RoktManager {
...
+    public processEvent(event: SDKEvent): void {
+        const hashedEvent = KitFilterHelper.hashEventMessage(
+            event.EventName,
+            event.EventCategory as valueof<typeof EventType>,
+            event.EventDataType as valueof<typeof MessageType>
+        ).toString();
+
+        if (Object.keys(this.placementEventMappingLookup).includes(hashedEvent)) {
+            const mappedValue = this.placementEventMappingLookup[hashedEvent];
+            this.store.setSessionSelectionAttributes({
```
rmi22186, line 175: If `this.placementEventMappingLookup` is an object with keys of `hashedEvent`, why turn it into an array to look up the key of `hashedEvent`?


```suggestion
        if (this.placementEventMappingLookup[hashedEvent]) {
```

src/store.ts
```diff
@@ -615,6 +620,17 @@ export default function Store(
...
 
+    this.getSessionSelectionAttributes = (): SessionSelectionAttributes =>
+        this.sessionSelectionAttributes || {};
+
+    this.setSessionSelectionAttributes = (attributes: SessionSelectionAttributes) => {
+        this.sessionSelectionAttributes = attributes;
+        this.persistenceData.gs.ssa = attributes;
+        console.warn('Setting session selection attributes', this.persistenceData);
+        // mpInstance._Persistence.update();
+        mpInstance._Persistence.savePersistence(this.persistenceData);
```
rmi22186, line 631: clarify the difference here for me.  Should .update be removed?

test/jest/kitFilterHelper.spec.ts
```diff
@@ -168,6 +168,105 @@ describe('FilterHashingUtilities', () => {
         });
     });
 
+    describe('#hashEventMessage', () => {
+        it('should hash event message for SessionStart', () => {
+            const eventName = 'Session Start';
+            const eventType = EventType.Other;
+            const messageType = MessageType.SessionStart;
+            const resultHash = KitFilterHelper.hashEventMessage(eventName, eventType, messageType);
+            expect(resultHash).toBe(1006044049);
```
rmi22186, line 177: session start and session end are not the best examples for this, because their event name is technically not correct. 

They are numbers.  Here's session start, as an example.  

Image

I don't think the examples _have_ to be session start and session end, and you were doing it just to change the message type, is that right?

test/jest/roktManager.spec.ts
```diff
@@ -347,6 +416,122 @@ describe('RoktManager', () => {
...
+            expect(lookup).toEqual({
+                '-1484452948': 'foo-mapped-flag',
+                '1838502119': 'ad_viewed_test'
+            });
+        });
+        
+        it('should return an empty object if the placement event mapping is empty', () => {
+            const lookup = roktManager['generateMappedEventLookup']([]);
+            expect(lookup).toEqual({});
+        });
```
rmi22186, line 446: maybe add a test that puts an invalid mapping in here to see if the error handling is robust?

test/src/tests-persistence.ts
```diff
@@ -1,6 +1,6 @@
 import Utils from './config/utils';
 import sinon from 'sinon';
-import fetchMock from 'fetch-mock/esm/client';
+import fetchMock, { done } from 'fetch-mock/esm/client';
```
rmi22186, line 3: seems like you are removing all the `done`s here in favor of `await`-ing.  Why is `done` being imported?

rmi22186 (commented):
One thing I didn't see here was the session selection attribute being removed when a session ends.

Also - I may have missed it, but is there an integration test that mimicks a page refresh with the same session?  Something like:
Initialize mParticle
Log a mapped event
check for session selection attributes
initialize mParticle again
check for session selection attributes to ensure they are pulled into the store

test/jest/roktManager.spec.ts
```diff
@@ -254,7 +277,12 @@ describe('RoktManager', () => {
                     ])
                 },
             };
-            roktManager.init(kitConfig as IKitConfigs, {} as IMParticleUser, mockMPInstance.Identity);
+            roktManager.init(
+                kitConfig as IKitConfigs,
+                {} as IMParticleUser,
+                mockMPInstance.Identity,
+                mockMPInstance._Store
+            );
```
rmi22186, line 285: Assume this was just missed here

```suggestion
                mockMPInstance._Store,
                mockMPInstance.Logger,
            );
```

rmi22186 (approved):

PR #1062, opened 2025-08-27
feat: added only rokt, all, none click ids when collectClickIdV2Enabl…

jaissica12:
## Instructions

 ## Summary
 - Added only rokt, all and none as options when collectClickIdV2Enabled is enabled. 
 - default mode is none

 ## Testing Plan

 - Closes https://mparticle-eng.atlassian.net/browse/SQDSDKS-7536

### DIFF
--- a/src/batchUploader.ts
+++ b/src/batchUploader.ts
@@ -175,7 +175,7 @@ export class BatchUploader {
         let integrationAttributes: IntegrationAttributes = _Store.integrationAttributes;
         const integrationSpecificIds = getFeatureFlag(Constants.FeatureFlags.CaptureIntegrationSpecificIds) as boolean;
         const integrationSpecificIdsV2 = getFeatureFlag(Constants.FeatureFlags.CaptureIntegrationSpecificIdsV2) as string || '';
-        const isIntegrationCaptureEnabled = (integrationSpecificIdsV2 ? integrationSpecificIdsV2 !== Constants.CaptureIntegrationSpecificIdsV2Modes.None : false) || (integrationSpecificIds === true);
+        const isIntegrationCaptureEnabled = (integrationSpecificIdsV2 && integrationSpecificIdsV2 !== Constants.CaptureIntegrationSpecificIdsV2Modes.None) || integrationSpecificIds === true;
 
         // https://go.mparticle.com/work/SQDSDKS-5053
         if (isIntegrationCaptureEnabled) {
--- a/src/mp-instance.ts
+++ b/src/mp-instance.ts
@@ -1393,7 +1393,7 @@ function completeSDKInitialization(apiKey, config, mpInstance) {
         const integrationSpecificIds = getFeatureFlag(CaptureIntegrationSpecificIds) as boolean;
         const integrationSpecificIdsV2 = getFeatureFlag(CaptureIntegrationSpecificIdsV2) as string;
         
-        const isIntegrationCaptureEnabled = (integrationSpecificIdsV2 ? integrationSpecificIdsV2 !== CaptureIntegrationSpecificIdsV2Modes.None : false) || (integrationSpecificIds === true);
+        const isIntegrationCaptureEnabled = (integrationSpecificIdsV2 && integrationSpecificIdsV2 !== CaptureIntegrationSpecificIdsV2Modes.None) || integrationSpecificIds === true;
         if (isIntegrationCaptureEnabled) {
             let captureMode: valueof<typeof CaptureIntegrationSpecificIdsV2Modes> | undefined;
             if (integrationSpecificIds || integrationSpecificIdsV2 === CaptureIntegrationSpecificIdsV2Modes.All) {
--- a/src/serverModel.ts
+++ b/src/serverModel.ts
@@ -281,7 +281,7 @@ export default function ServerModel(
             // https://go.mparticle.com/work/SQDSDKS-7639
             const integrationSpecificIds = getFeatureFlag && (getFeatureFlag(Constants.FeatureFlags.CaptureIntegrationSpecificIds) as boolean);
             const integrationSpecificIdsV2 = getFeatureFlag && ((getFeatureFlag(Constants.FeatureFlags.CaptureIntegrationSpecificIdsV2) as string) || '');
-            const isIntegrationCaptureEnabled = (integrationSpecificIdsV2 ? integrationSpecificIdsV2 !== Constants.CaptureIntegrationSpecificIdsV2Modes.None : false) || (integrationSpecificIds === true);
+            const isIntegrationCaptureEnabled = (integrationSpecificIdsV2 && integrationSpecificIdsV2 !== Constants.CaptureIntegrationSpecificIdsV2Modes.None) || integrationSpecificIds === true;     
             if (isIntegrationCaptureEnabled) {
                 // Attempt to recapture click IDs in case a third party integration
                 // has added or updated  new click IDs since the last event was sent.
--- a/test/src/tests-batchUploader.ts
+++ b/test/src/tests-batchUploader.ts
@@ -413,7 +413,6 @@ describe('batch uploader', () => {
             const integrationCapture = window.mParticle.getInstance()._IntegrationCapture;
             // Mock the query params capture function because we cannot mock window.location.href
             sinon.stub(integrationCapture, 'getQueryParams').returns({ rtid: 'test-click-id' });
-            integrationCapture.capture();
             window.mParticle.Identity.getCurrentUser().setUserAttribute('foo', 'value');
             const user = window.mParticle.Identity.getCurrentUser();
             user.setConsentState(consentState);
--- a/src/sdkToEventsApiConverter.ts
+++ b/src/sdkToEventsApiConverter.ts
@@ -130,7 +130,7 @@ export function convertEvents(
     const integrationSpecificIds = getFeatureFlag && Boolean(getFeatureFlag(CaptureIntegrationSpecificIds));
     const integrationSpecificIdsV2 = getFeatureFlag && (getFeatureFlag(CaptureIntegrationSpecificIdsV2) as string);
         
-    const isIntegrationCaptureEnabled = (integrationSpecificIdsV2 ? integrationSpecificIdsV2 !== Constants.CaptureIntegrationSpecificIdsV2Modes.None : false) || (integrationSpecificIds === true);
+    const isIntegrationCaptureEnabled = (integrationSpecificIdsV2 && integrationSpecificIdsV2 !== Constants.CaptureIntegrationSpecificIdsV2Modes.None) || integrationSpecificIds === true;
 
     if (isIntegrationCaptureEnabled) {
         const capturedPartnerIdentities: PartnerIdentities = _IntegrationCapture?.getClickIdsAsPartnerIdentities();


src/integrationCapture.ts
```diff
@@ -160,9 +165,9 @@ export default class IntegrationCapture {
      * Captures Integration Ids from cookies and query params and stores them in clickIds object
      */
     public capture(): void {
-        const queryParams = this.captureQueryParams() || {};
-        const cookies = this.captureCookies() || {};
-        const localStorage = this.captureLocalStorage() || {};
+        let queryParams = this.captureQueryParams() || {};
+        let cookies = this.captureCookies() || {};
+        let localStorage = this.captureLocalStorage() || {};
```
alexs-mparticle, line 170: Why is this being changed to `let`? I don't see them being reassigned anywhere.

test/src/tests-integration-capture.ts
```diff
@@ -42,20 +43,21 @@ describe('Integration Capture', () => {
...
-            gclid: '234',
-            gbraid: '6574',
-            rtid: '45670808',
-            rclid: '7183717',
-            wbraid: '1234111',
-            ScCid: '1234',
-        });
-
         mParticle.init(apiKey, window.mParticle.config);
+        return waitForCondition(hasIdentifyReturned).then(function() {
```
alexs-mparticle, line 47: Why is this being moved after the init?

alexs-mparticle (commented):
Just a few questions. Otherwise this looks good.

test/src/tests-integration-capture.ts
jaissica12: Moved return `waitForCondition(hasIdentifyReturned).then()` after `init` because the condition becomes true once `init` triggers the identify flow and the mocked identity response returns. `_IntegrationCapture` is also created during `init` (when flags enable it), so stubbing and calling `capture()` happens after init to avoid `undefined` error. Returning that Promise from `beforeEach` tells Mocha to wait for identity completion and the stubbed capture to finish before running each test, which might remove flakiness. Previously, tests failed because the stub ran before `_IntegrationCapture` existed (and before identity completed), leading to `Trying to stub property 'getQueryParams' of undefined` error. Let me know if it sounds right, or we should update it?

src/integrationCapture.ts
jaissica12: I agree, they are being reassigned anywhere else. I have updated it to `const`

src/integrationCapture.ts
```diff
@@ -221,8 +227,9 @@ export default class IntegrationCapture {
      * Captures local storage based on the integration ID mapping.
      */
     public captureLocalStorage(): Dictionary<string> {
+        const keys = this.getAllowedKeysForMode();
```
rmi22186, line 230: Maybe we be more descriptive here.  Is this more accurate?
```suggestion
        const integrationKeys = this.getAllowedKeysForMode();
```

rmi22186: I see `activeKeys` below.  I'd just like to avoid using keys because it's too generic.  This is a non-blocking comment btw

src/mp-instance.ts
```diff
@@ -1386,11 +1386,22 @@ function completeSDKInitialization(apiKey, config, mpInstance) {
             ? { userIdentities: currentUserIdentities }
             : mpInstance._Store.SDKConfig.identifyRequest;
 
-        if (mpInstance._Helpers.getFeatureFlag(ReportBatching)) {
+        if (getFeatureFlag(ReportBatching)) {
             mpInstance._ForwardingStatsUploader.startForwardingStatsTimer();
         }
-
-        if (mpInstance._Helpers.getFeatureFlag(CaptureIntegrationSpecificIds)) {
+        // https://go.mparticle.com/work/SQDSDKS-7639
```
rmi22186, line 1392: I'd prioritize this ticket during our cool down week assuming there isn't anything else right now.

src/mp-instance.ts
```diff
@@ -1386,11 +1386,22 @@ function completeSDKInitialization(apiKey, config, mpInstance) {
...
+        const integrationSpecificIds = getFeatureFlag(CaptureIntegrationSpecificIds) as boolean;
+        const integrationSpecificIdsV2 = getFeatureFlag(CaptureIntegrationSpecificIdsV2) as string;
+        
+        const isIntegrationCaptureEnabled = (integrationSpecificIdsV2 ? integrationSpecificIdsV2 !== CaptureIntegrationSpecificIdsV2Modes.None : false) || (integrationSpecificIds === true);
```
rmi22186, line 1396: Just triple checking, because our server is a little weird and I haven't been in this code for a while.  Can you confirm
`integrationSpecificIds` is definitely a boolean `true` and not a string `"true"`?

rmi22186: Nevermind.  I see below in the store that in fact it will return a boolean

test/jest/integration-capture.spec.ts
```diff
@@ -32,14 +32,103 @@ describe('Integration Capture', () => {
...
+            const clickIds = integrationCapture.captureQueryParams();
+            const clickIdCookies = integrationCapture.captureCookies();
+            const clickIdLocalStorage = integrationCapture.captureLocalStorage();
+
+            expect(clickIds).toEqual({ rtid: 'rt1', rclid: 'rc1' });
+            expect(clickIdCookies).toEqual({ RoktTransactionId: 'xyz' });
+            expect(clickIdLocalStorage).toEqual({ RoktTransactionId: 'ls-rok' });
+        });
+
+        it('should return all mapped keys from helpers when captureMode is all (lowercase)', () => {
```
rmi22186, line 106: i see tests for `roktonly` and `all`, how about `none`?

src/integrationCapture.ts
jaissica12: Replaced `keys` and `activeKeys` with `integrationKeys`

test/jest/integration-capture.spec.ts
jaissica12: added test for `none` case

test/jest/integration-capture.spec.ts
```diff
@@ -503,7 +590,7 @@ describe('Integration Capture', () => {
             window.document.cookie = '_cookie2=39895811.9165333198';
             window.document.cookie = 'baz=qux';
 
-            const integrationCapture = new IntegrationCapture();
+            const integrationCapture = new IntegrationCapture('none');
```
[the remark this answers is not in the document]
jaissica12: I used `none` in tests that assert no values should be captured to make the intent explicit and `all` in tests that assert positive capture. For the mapping/name‑validation cases, the capture mode is effectively independent, the outcome is driven by whether the inputs are mapped — so using `none` or `all` doesn’t change results. I can standardize to `all` across tests

test/src/tests-batchUploader.ts
```diff
@@ -386,7 +386,8 @@ describe('batch uploader', () => {
                 }
             };
 
-            window.mParticle.config.flags.captureIntegrationSpecificIds = "True";
+            window.mParticle.config.flags.captureIntegrationSpecificIds = "False";
+            window.mParticle.config.flags.captureIntegrationSpecificIdsV2 = "roktonly";
```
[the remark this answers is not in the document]
jaissica12: Since the valid cases which the web-sdk receives is either (`captureIntegrationSpecificIds` as `True` and `captureIntegrationSpecificIdsV2` = `all`) or (`captureIntegrationSpecificIds` = `False` and `captureIntegrationSpecificIdsV2` = `roktlonly`). Since in this test we are just testing for `rtid` I add tests like this to keep in sync with the flags sdk would be receiving. The output would remain same with (`True` and `all`) and (`False` and `roktonly`)

src/serverModel.ts
```diff
@@ -277,16 +277,20 @@ export default function ServerModel(
...
-                mpInstance._IntegrationCapture.capture();
-                const transformedClickIDs = mpInstance._IntegrationCapture.getClickIdsAsCustomFlags();
-                customFlags = {...transformedClickIDs, ...customFlags};
-
-                const transformedIntegrationAttributes = mpInstance._IntegrationCapture.getClickIdsAsIntegrationAttributes();
-                integrationAttributes = {...transformedIntegrationAttributes, ...integrationAttributes};
+            if (mpInstance._Helpers.getFeatureFlag) {
+                // https://go.mparticle.com/work/SQDSDKS-7639
+                const integrationSpecificIds = mpInstance._Helpers.getFeatureFlag(Constants.FeatureFlags.CaptureIntegrationSpecificIds) as boolean;
+                const integrationSpecificIdsV2 = ((mpInstance._Helpers.getFeatureFlag(Constants.FeatureFlags.CaptureIntegrationSpecificIdsV2) as string) || '');
```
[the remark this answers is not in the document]
jaissica12: Tests are passing without the `getFeatureFlag` so I'll remove the redundant `if` block

src/batchUploader.ts
```diff
@@ -173,9 +173,12 @@ export class BatchUploader {
 
         let customFlags: SDKEventCustomFlags = {...event.CustomFlags};
         let integrationAttributes: IntegrationAttributes = _Store.integrationAttributes;
+        const integrationSpecificIds = getFeatureFlag(Constants.FeatureFlags.CaptureIntegrationSpecificIds) as boolean;
+        const integrationSpecificIdsV2 = getFeatureFlag(Constants.FeatureFlags.CaptureIntegrationSpecificIdsV2) as string || '';
+        const isIntegrationCaptureEnabled = (integrationSpecificIdsV2 ? integrationSpecificIdsV2 !== Constants.CaptureIntegrationSpecificIdsV2Modes.None : false) || (integrationSpecificIds === true);
```
rmi22186, line 178: this ternary operator is a bit confusing.  

This is a tiny bit shorter, but feels congnitively less complex to me.  

```suggestion
        const isIntegrationCaptureEnabled = (integrationSpecificIdsV2 && integrationSpecificIdsV2 !== Constants.CaptureIntegrationSpecificIdsV2Modes.None) || integrationSpecificIds === true;;
```

What do you think?

src/serverModel.ts
```diff
@@ -276,18 +276,22 @@ export default function ServerModel(
...
             let integrationAttributes: IntegrationAttributes = mpInstance._Store.integrationAttributes;
 
+            const { getFeatureFlag } = mpInstance._Helpers;
             // https://go.mparticle.com/work/SQDSDKS-5053
-            if (mpInstance._Helpers.getFeatureFlag && mpInstance._Helpers.getFeatureFlag(Constants.FeatureFlags.CaptureIntegrationSpecificIds)) {
-
+            // https://go.mparticle.com/work/SQDSDKS-7639
+            const integrationSpecificIds = getFeatureFlag && (getFeatureFlag(Constants.FeatureFlags.CaptureIntegrationSpecificIds) as boolean);
+            const integrationSpecificIdsV2 = getFeatureFlag && ((getFeatureFlag(Constants.FeatureFlags.CaptureIntegrationSpecificIdsV2) as string) || '');
+            const isIntegrationCaptureEnabled = (integrationSpecificIdsV2 ? integrationSpecificIdsV2 !== Constants.CaptureIntegrationSpecificIdsV2Modes.None : false) || (integrationSpecificIds === true);
```
rmi22186, line 284: repeat of previous ternary refactor

test/src/tests-batchUploader.ts
```diff
@@ -407,13 +408,12 @@ describe('batch uploader', () => {
...
-            sinon.stub(window.mParticle.getInstance()._IntegrationCapture, 'getQueryParams').returns({
-                rtid: 'test-click-id',
-            });
-
             window.mParticle.init(apiKey, window.mParticle.config);
             await waitForCondition(hasIdentifyReturned);
+            const integrationCapture = window.mParticle.getInstance()._IntegrationCapture;
+            // Mock the query params capture function because we cannot mock window.location.href
+            sinon.stub(integrationCapture, 'getQueryParams').returns({ rtid: 'test-click-id' });
+            integrationCapture.capture();
```
rmi22186, line 416: doesn't seem like you need this line here.  the capture will happen during the visibility change, which runs through the batchuploader code

(I pulled this down locally and commented it out. but you can try as well)

test/src/tests-integration-capture.ts
rmi22186: The above may be true (it's late, so maybe I'm repeating the same thing). 

the `_IntegraitonCapture` creation was moved from when [mpInstance is made](https://github.com/mParticle/mparticle-web-sdk/pull/1062/files#diff-b5cbf8c804983cdb3506dd74cb6aecac559f7d270a9b9567ebfd8c2698437c64L131) to [inside of completeSDKInitialization](https://github.com/mParticle/mparticle-web-sdk/pull/1062/files#diff-b5cbf8c804983cdb3506dd74cb6aecac559f7d270a9b9567ebfd8c2698437c64R1397-R1404) in order to  have the feature flags available to pass into `_IntegrationCapture`.

rmi22186 (changes requested):
minor comments before we merge

src/batchUploader.ts
jaissica12: I agree, makes it more readable. I have updated the same for all the occurrences of `isIntegrationCaptureEnabled`

alexs-mparticle (approved):

PR #1063, opened 2025-09-02
feat: added noFunctional in batchUploader and updated tests

jaissica12:
## Instructions

 ## Summary
 - Added `noFunctional` to the `batchUploader `such that `offlineStorageEnabled` is true when is `noFunctional` false

 ## Testing Plan

 - Closes https://rokt.atlassian.net/browse/SDKE-68

### DIFF
--- a/src/batchUploader.ts
+++ b/src/batchUploader.ts
@@ -72,7 +72,7 @@ export class BatchUploader {
 
         // Cache Offline Storage Availability boolean
         // so that we don't have to check it every time
-        this.offlineStorageEnabled = this.isOfflineStorageAvailable();
+        this.offlineStorageEnabled = this.isOfflineStorageAvailable() && !mpInstance._Store.getNoFunctional();
 
         if (this.offlineStorageEnabled) {
             this.eventVault = new SessionStorageVault<SDKEvent[]>(
--- a/test/jest/batchUploader.spec.ts
+++ b/test/jest/batchUploader.spec.ts
@@ -15,18 +15,27 @@ describe('BatchUploader', () => {
         // Create a mock mParticle instance with mocked methods for instantiating a BatchUploader
         mockMPInstance = {
             _Store: {
+                storageName: 'mprtcl-v4_abcdef',
+                noFunctional: false,
+                getNoFunctional: function(this: any) { return this.noFunctional; },
+                deviceId: 'device-1',
                 SDKConfig: {
                     flags: {}
                 }
             },
             _Helpers: {
-                getFeatureFlag: jest.fn().mockReturnValue(false),
+                getFeatureFlag: jest.fn().mockReturnValue('100'),
                 createServiceUrl: jest.fn().mockReturnValue('https://mock-url.com')
             },
             Identity: {
                 getCurrentUser: jest.fn().mockReturnValue({
                     getMPID: () => 'test-mpid'
                 })
+            },
+            Logger: {
+                verbose: jest.fn(),
+                error: jest.fn(),
+                warning: jest.fn(),
             }
         } as unknown as IMParticleWebSDKInstance;
 
@@ -102,6 +111,45 @@ describe('BatchUploader', () => {
             const secondCallTime = batchUploader['lastASTEventTime'];
             
             expect(secondCallTime).toBe(firstCallTime);
+        });
+    });
+
+    describe('noFunctional', () => {
+        beforeEach(() => {
+            localStorage.clear();
+            sessionStorage.clear();
+        });
+
+        it('should disable offline storage when noFunctional is true', () => {
+            mockMPInstance._Store.noFunctional = true;
+
+            const uploader = new BatchUploader(mockMPInstance, 1000);
+            // @ts-expect-error private
+            expect(uploader.offlineStorageEnabled).toBe(false);
+
+            uploader.queueEvent({ EventDataType: 4 } as any);
+            expect(sessionStorage.getItem('mprtcl-v4_abcdef-events')).toBeNull();
+            expect(localStorage.getItem('mprtcl-v4_abcdef-batches')).toBeNull();
+        });
+
+        it('should enable offline storage when noFunctional is default (false)', () => {
+            // mockMPInstance._Store.noFunctional = false;
+            const uploader = new BatchUploader(mockMPInstance, 1000);
+
+            // @ts-expect-error private
+            expect(uploader.offlineStorageEnabled).toBe(true);
+
+            uploader.queueEvent({ EventDataType: 4 } as any);
+            expect(sessionStorage.getItem('mprtcl-v4_abcdef-events')).not.toBeNull();
+        });
+
+        it('should enable offline storage when noFunctional is false', () => {
+            mockMPInstance._Store.noFunctional = false;
+
+            const uploader = new BatchUploader(mockMPInstance, 1000);
+
+            uploader.queueEvent({ EventDataType: 4 } as any);
+            expect(sessionStorage.getItem('mprtcl-v4_abcdef-events')).not.toBeNull();
         });
     });
 }); 
--- a/test/src/tests-batchUploader.ts
+++ b/test/src/tests-batchUploader.ts
@@ -1203,6 +1203,7 @@ describe('batch uploader', () => {
             });
 
             window.sessionStorage.clear();
+            window.localStorage.clear();
         });
 
         afterEach(() => {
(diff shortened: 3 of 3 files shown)


test/jest/batchUploader.spec.ts
```diff
@@ -104,4 +113,43 @@ describe('BatchUploader', () => {
...
+            expect(sessionStorage.getItem('mprtcl-v4_abcdef-events')).toBeNull();
+            expect(localStorage.getItem('mprtcl-v4_abcdef-batches')).toBeNull();
+        });
+
+        it('should enable offline storage when noFunctional is default (false)', () => {
+            // mockMPInstance._Store.noFunctional = false;
+            const uploader = new BatchUploader(mockMPInstance, 1000);
+
+            // @ts-expect-error private
+            expect(uploader.offlineStorageEnabled).toBe(true);
```
alexs-mparticle, line 140: You may be able to get rid of this override by accessing the private attribute like this:

```suggestion
            expect(uploader['offlineStorageEnabled']).toBe(true);
```

test/src/tests-batchUploader.ts
```diff
@@ -1716,5 +1717,113 @@ describe('batch uploader', () => {
...
+                waitForCondition(hasIdentifyReturned)
+                .then(() => {
+                    const mpInstance = window.mParticle.getInstance();
+                    const uploader = mpInstance._APIClient.uploader;
+                    uploader.queueEvent(event0);
+                    expect(!!window.sessionStorage.getItem(eventStorageKey)).to.equal(true);
+                    done();
+                })
+                .catch(done);
+            });
```
alexs-mparticle, line 1741: Try to use `async/await` instead of `then` for these new tests.

```suggestion
            it('should write session storage when noFunctional is default (false)', async () => {
                window.mParticle._resetForTests(MPConfig);
                window.mParticle.config.flags = {
                    offlineStorage: '100',
                    ...enableBatchingConfigFlags,
                };
                window.mParticle.init(apiKey, window.mParticle.config);
                await waitForCondition(hasIdentifyReturned);

                const mpInstance = window.mParticle.getInstance();
                const uploader = mpInstance._APIClient.uploader;
                uploader.queueEvent(event0);
                expect(!!window.sessionStorage.getItem(eventStorageKey)).to.equal(true);
            });
```

alexs-mparticle (changes requested):

src/store.ts
```diff
@@ -590,6 +591,17 @@ export default function Store(
         this.noTargeting = noTargeting;
     };
 
+    this.getPrivacyFlagForStorage = (storageType: typeof StorageTypes[number]): boolean => {
+        const privacyControl = StoragePrivacyMap[storageType];
+        if (privacyControl === 'functional') {
+            return !!this.getNoFunctional();
+        }
+        if (privacyControl === 'targeting') {
+            return !!this.getNoTargeting();
```
rmi22186, line 600: is there a reason for !!?  getNoFuntional and getNoTargeting already return booleans

src/constants.ts
```diff
@@ -215,3 +215,17 @@ export const HTTP_UNAUTHORIZED = 401 as const;
...
+
+export const StorageTypes = ['UserData', 'Products', 'Events', 'Batches', 'IdCache', 'TimeOnSite'];
+
+export const StoragePrivacyMap: Record<typeof StorageTypes[number], PrivacyControl> = {
+    'UserData' : 'functional',
+    'Products': 'targeting',
+    'Events': 'functional',
+    'Batches': 'functional',
+    'IdCache': 'functional',
+    'TimeOnSite': 'targeting',
```
rmi22186, line 230: camelCase is fine here.  without '' since keys are strings by default.

test/jest/batchUploader.spec.ts
```diff
@@ -104,4 +124,41 @@ describe('BatchUploader', () => {
...
+            const uploader = new BatchUploader(mockMPInstance, 1000);
+            expect(uploader['offlineStorageEnabled']).toBe(false);
+
+            uploader.queueEvent({ EventDataType: 4 } as any);
+            expect(sessionStorage.getItem('mprtcl-v4_abcdef-events')).toBeNull();
+            expect(localStorage.getItem('mprtcl-v4_abcdef-batches')).toBeNull();
+        });
+
+        it('should enable offline storage when noFunctional is default (false)', () => {
+            // mockMPInstance._Store.noFunctional = false;
```
rmi22186, line 146: remove comment?
```suggestion
```

src/store.ts
jaissica12: I agree that `getNoFunctional()` and `getNoTargeting()` already return booleans, so the `!!` is technically redundant. I added it as a safety measure to ensure the result is explicitly a boolean in case the internal implementation of those accessors changes in the future.
We can safely remove the `!!` without affecting correctness. If the return type ever changes, we can revisit and reintroduce normalization at the call site as needed.

test/src/tests-batchUploader.ts
```diff
@@ -1716,5 +1717,110 @@ describe('batch uploader', () => {
...
+
+            it('should NOT store events in session storage when noFunctional is true', async () => {
+                window.mParticle.config.flags = {
+                    offlineStorage: '100',
+                    ...enableBatchingConfigFlags,
+                };
+                window.mParticle.config.launcherOptions = {
+                    ...(window.mParticle.config.launcherOptions || {}),
+                    noFunctional: true,
+                };
```
rmi22186, line 1749: is there a reason to spread what was previously existing on launcherOptions?  i'd expect just setting the noFunctional:true or false in each of these would be more appropriate so that we know exactly what is going in the `launcherOptions` for each test.  same for all the tests below

```suggestion
                window.mParticle.config.launcherOptions = {
                    noFunctional: true,
                };
```

test/jest/batchUploader.spec.ts
```diff
@@ -135,29 +139,39 @@ describe('BatchUploader', () => {
...
 
-        it('should enable offline storage when noFunctional is default (false)', () => {
+        it('should enable offline storage when noFunctional is default (false)', async () => {
             const uploader = new BatchUploader(mockMPInstance, 1000);
 
-            expect(uploader['offlineStorageEnabled']).toBe(true);
+            expect(uploader['OfflineStorage']).toBe(true);
 
-            uploader.queueEvent({ EventDataType: 4 } as any);
+            uploader.queueEvent({ EventDataType: 4, SessionId: 's1' } as any);
```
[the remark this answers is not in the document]
jaissica12: It’s not strictly required. The code will still bucket by `MPID` if `SessionId` is missing, it will group under an `undefined` session and can still produce a batch. I thought leaving `SessionId` out can make the test flaky, which is why I added it. Please let me know if I am missing something here.

src/constants.ts
```diff
@@ -215,3 +215,17 @@ export const HTTP_UNAUTHORIZED = 401 as const;
...
+// Privacy dependency map for storage keys
+export type PrivacyControl = 'functional' | 'targeting';
+
+export type StorageTypes = 'SDKState' | 'Products' | 'Events' | 'Batches' | 'IdCache' | 'TimeOnSite';
+
+export const StoragePrivacyMap: Record<StorageTypes, PrivacyControl> = {
+    SDKState: 'functional',
+    Products: 'targeting',
+    Events: 'functional',
+    Batches: 'functional',
```
alexs-mparticle, line 228:
```suggestion
    OfflineEvents: 'functional',
```

src/batchUploader.ts
```diff
@@ -43,7 +43,7 @@ export class BatchUploader {
     batchingEnabled: boolean;
     private eventVault: SessionStorageVault<SDKEvent[]>;
     private batchVault: LocalStorageVault<Batch[]>;
-    private offlineStorageEnabled: boolean = false;
+    private OfflineStorage: boolean = false;
```
alexs-mparticle, line 46: Change this back

```suggestion
    private offlineStorageEnabled: boolean = false;
```

test/src/tests-batchUploader.ts
```diff
@@ -1715,5 +1716,106 @@ describe('batch uploader', () => {
...
+                window.mParticle.config.launcherOptions = {
+                    noFunctional: true,
+                };
+                window.mParticle.init(apiKey, window.mParticle.config);
+                await waitForCondition(hasIdentifyReturned);
+                const mpInstance = window.mParticle.getInstance();
+                const uploader = mpInstance._APIClient.uploader;
+                uploader.queueEvent(event0);
+                await window.mParticle.getInstance()._APIClient.uploader.prepareAndUpload();
```
rmi22186, line 1797: ditto 
```suggestion
```

test/src/tests-batchUploader.ts
```diff
@@ -1715,5 +1716,106 @@ describe('batch uploader', () => {
...
+                    ...enableBatchingConfigFlags,
+                };
+                window.mParticle.config.launcherOptions = {
+                    noFunctional: false,
+                };
+                window.mParticle.init(apiKey, window.mParticle.config);
+                await waitForCondition(hasIdentifyReturned);
+                const mpInstance = window.mParticle.getInstance();
+                const uploader = mpInstance._APIClient.uploader;
+                fetchMock.post(urls.events, 500, { overwriteRoutes: true });
```
rmi22186, line 1813: not needed if you don't call prepareAndUpload anymore
```suggestion
```

test/src/tests-batchUploader.ts
```diff
@@ -1715,5 +1716,106 @@ describe('batch uploader', () => {
...
+                uploader.queueEvent(event0);
+                await window.mParticle.getInstance()._APIClient.uploader.prepareAndUpload();
```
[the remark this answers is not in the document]
jaissica12: @rmi22186 As per my understanding we would require `await window.mParticle.getInstance()._APIClient.uploader.prepareAndUpload();` when we want to test for local storage because this is where batches are updated.

test/src/tests-batchUploader.ts
```diff
@@ -1715,5 +1716,106 @@ describe('batch uploader', () => {
...
+                uploader.queueEvent(event0);
+                expect(!!window.sessionStorage.getItem(eventStorageKey)).to.equal(true);
+            });
+
+            it('should store batches in local storage when noFunctional is false by default', async () => {
+                window.mParticle.init(apiKey, window.mParticle.config);
+                await waitForCondition(hasIdentifyReturned);
+                const mpInstance = window.mParticle.getInstance();
+                const uploader = mpInstance._APIClient.uploader;
+                fetchMock.post(urls.events, 500, { overwriteRoutes: true });
```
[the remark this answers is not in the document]
rmi22186: As discussed in person, I was wrong, resolving

test/src/tests-batchUploader.ts
rmi22186: As discussed in person, I was wrong, resolving

test/src/tests-batchUploader.ts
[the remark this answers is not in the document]
rmi22186: As discussed in person, I was wrong, resolving

rmi22186 (approved):

alexs-mparticle (approved):

PR #1064, opened 2025-09-02
feat: disable time tracking when noTargeting is set to true

jaissica12:
## Instructions

 ## Summary
 - Updated `store.ts` such that `_timeOnSiteTimer` is initialized only if `noTargeting` is `false`.

 ## Testing Plan

 - Closes https://rokt.atlassian.net/browse/SDKE-90

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -700,9 +700,20 @@ export default function Store(
             this.SDKConfig[baseUrlKeys] = baseUrls[baseUrlKeys];
         }
 
+        // Extract privacy flags directly from config into Store BEFORE initializing timers
+        if (config.hasOwnProperty('noFunctional')) {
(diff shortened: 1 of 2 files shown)


alexs-mparticle (approved):

rmi22186 (changes requested):
looks fine but needs to be rebased against https://github.com/mParticle/mparticle-web-sdk/pull/1063 and logic need to be updated using mpInstance._Store.getPrivacyFlag

rmi22186 (changes requested):
pretty sure it should be config.launcher.noFunc/noTargeting per https://docs.rokt.com/developers/integration-guides/getting-started/ecommerce/ecommerce-sdk-integration#overview

PR #1070, opened 2025-09-15
feat: disable session cookie/localStorage when noFunctional is set

jaissica12:
## Instructions

 ## Summary
 - In persistence, updated to disable localStorage/cookies when noFunctional is set.
 - added if block that localStorage is enabled when noFunctional is not set.
 - Added jest and integration tests

 ## Testing Plan

 - Closes https://rokt.atlassian.net/browse/SDKE-91


src/persistence.js
```diff
@@ -39,51 +39,52 @@ export default function _Persistence(mpInstance) {
...
-                        storage.removeItem(mpInstance._Store.storageName);
-                    } else if (cookies) {
-                        allData = cookies;
-                    }
-                    self.storeDataInMemory(allData);
-                } else {
-                    // For migrating from cookie to localStorage -- If an instance is newly switching from cookies to localStorage, then
-                    // no mParticle localStorage exists yet and there are cookies. Get the cookies, set them to localStorage, then delete the cookies.
-                    if (cookies) {
+            if (!mpInstance._Store.getNoFunctional()) {
```
rmi22186, line 42: You can reduce the number of changes here by doing the inverse and returning early:


```suggestion
            if (mpInstance._Store.getNoFunctional()) {
                    return;
            }
            // then unindent everything you indented below
```

src/persistence.js
```diff
@@ -114,14 +115,15 @@ export default function _Persistence(mpInstance) {
...
             }
-
-            // https://go.mparticle.com/work/SQDSDKS-6046
-            // Stores all non-current user MPID information into the store
-            for (var key in allData) {
-                if (allData.hasOwnProperty(key)) {
-                    if (!SDKv2NonMPIDCookieKeys[key]) {
-                        mpInstance._Store.nonCurrentUserMPIDs[key] =
-                            allData[key];
+            if (!mpInstance._Store.getNoFunctional()) {
```
rmi22186, line 118: Not sure why this check is necessary. it wouldn't get into here if it didn't pass the check on the line 42 that you are making, right?

src/persistence.js
```diff
@@ -326,15 +332,12 @@ export default function _Persistence(mpInstance) {
         return parsedDecodedProducts;
     };
 
-    // https://go.mparticle.com/work/SQDSDKS-6021
-    this.setLocalStorage = function() {
+    this.setProductStorage = function() {
```
rmi22186, line 335: What is the reason for doing product changes in this PR?  Seems out of scope, but maybe I'm missing something.

src/persistence.js
```diff
@@ -356,6 +359,21 @@ export default function _Persistence(mpInstance) {
...
+        if (!mpInstance._Store.isLocalStorageAvailable) {
+            return;
+        }
+
+        var key = mpInstance._Store.storageName,
+            localStorageData = self.getLocalStorage() || {},
+            currentUser = mpInstance.Identity.getCurrentUser(),
+            mpid = currentUser ? currentUser.getMPID() : null;
+
+        // Always update product storage when setLocalStorage is invoked
```
rmi22186, line 375: Why?

test/jest/persistence.spec.ts
```diff
@@ -0,0 +1,155 @@
...
+    describe('#noFunctional privacy flag', () => {
+        describe('true', () => {
+            beforeEach(() => {
+                store.setNoFunctional(true);
+                store.webviewBridgeEnabled = false;
+            });
+
+            it('should NOT write to cookie when useCookieStorage is true', () => {
+                store.SDKConfig.useCookieStorage = true;
+                jest.clearAllMocks();
```
rmi22186, line 61: I think you can add an afterEach block and include jest.clearAllMocks() in that.

test/jest/persistence.spec.ts
```diff
@@ -0,0 +1,155 @@
...
+            });
+
+            it('should write to localStorage when useCookieStorage is false', () => {
+                store.SDKConfig.useCookieStorage = false;
+                jest.clearAllMocks();
+
+                mockMPInstance._Persistence.update();
+
+                expect(mockMPInstance._Persistence.setCookie).not.toHaveBeenCalled();
+                expect(mockMPInstance._Persistence.setLocalStorage).toHaveBeenCalled();
```
rmi22186, line 106: called..but can you try to pull the item to ensure it's not null here and in other tests?

test/src/tests-persistence.ts
```diff
@@ -1948,4 +1948,112 @@ describe('persistence', () => {
...
     });
+
+    describe('#noFunctional', () => {
+        beforeEach(() => {
+            mParticle._resetForTests(MPConfig);
+        });
+
+        describe('true', () => {
+            beforeEach(() => {
+                mParticle.config.noFunctional = true;
```
rmi22186, line 1959:
```suggestion
                mParticle.config.launcherOptions = {noFunctional: true};
```

test/src/tests-persistence.ts
```diff
@@ -1948,4 +1948,112 @@ describe('persistence', () => {
...
+            });
+            it('should NOT store cookie when useCookieStorage = true', async () => {
```
rmi22186, line 1961:
```suggestion
            });

            it('should NOT store cookie when useCookieStorage = true', async () => {
```

rmi22186 (changes requested):

src/persistence.js
jaissica12: @rmi22186  `setLocalStorage` currently writes both SDK state (`noFunctional`) and product storage (`noTargeting`). As current PR is for main local storage so if `noFunctional` is `true` then I independently call `setProductStorage` function to update the product storage and if its `false` the login remain as is. So the comment simply indicates irrespective of `noFunctional` privacy flag product storage will be updated. 
Let me know if this is causing confusion I can remove it, initially i added it when I was breaking down the functions for better understanding.

src/persistence.js
jaissica12: I didn't do this as `initializeStorage` function is also used to initialize product storage and if i return early here then product storage will be affected

src/persistence.js
jaissica12: Let me know your opinion on line 42 changes (I have added my comment above for not returning early)

src/persistence.js
rmi22186: @jaissica12 - I'd say since you, @alexs-mparticle, and I made the decision to remove functionality for product storage, that this could be reverted just to avoid too many code changes.  then the next PR will remove all the setproduct functionality anyhow.

jaissica12: @rmi22186 I've reverted `setLocalStorage` to how it initially was and removed separate function `setProductStorage`

src/persistence.js
```diff
@@ -86,6 +89,17 @@ export default function _Persistence(mpInstance) {
...
+            // https://go.mparticle.com/work/SQDSDKS-6046
+            // Stores all non-current user MPID information into the store
+            for (var key in allData) {
+                if (allData.hasOwnProperty(key)) {
+                    if (!SDKv2NonMPIDCookieKeys[key]) {
+                        mpInstance._Store.nonCurrentUserMPIDs[key] =
+                            allData[key];
+                    }
+                }
+            }
```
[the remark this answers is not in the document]
jaissica12: I kept it as is in this PR because I thought keeping the code related to main storage / `SDKState` together might be easier for better understanding. However, I understand your point about keeping changes minimal in this PR and eventually after products de-scoping it would be the same so I'll revert it now.

src/persistence.js
```diff
@@ -39,6 +39,9 @@ export default function _Persistence(mpInstance) {
             if (!mpInstance._Store.isLocalStorageAvailable) {
                 mpInstance._Store.SDKConfig.useCookieStorage = true;
             }
+            if (mpInstance._Store.getPrivacyFlag('SDKState')) {
```
alexs-mparticle, line 42: I would consider moving this to the top of the function so that we don't read from cookies/localStorage at all if privacy is turned on.

test/jest/persistence.spec.ts
```diff
@@ -0,0 +1,145 @@
+import Store, { IStore } from '../../src/store';
+import { IMParticleWebSDKInstance } from '../../src/mp-instance';
+import { SDKInitConfig } from '../../src/sdkRuntimeModels';
+
+describe('Persistence store', () => {
```
alexs-mparticle, line 5: Nit: `store` is redundant here.

```suggestion
describe('Persistence', () => {
```

test/jest/persistence.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+    let store: IStore;
+    let mockMPInstance: IMParticleWebSDKInstance;
+
+    beforeEach(() => {
+        mockMPInstance = {
```
alexs-mparticle, line 10: I think you may be able to remove some of these mocks after your recent refactors.

test/jest/persistence.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+            },
+        } as unknown as IMParticleWebSDKInstance;
+
+        store = {} as IStore;
+        Store.call(store, {} as SDKInitConfig, mockMPInstance, 'apikey');
+        window.localStorage.removeItem('mprtcl-v4');
+    });
+
+    describe('#noFunctional privacy flag', () => {
+        describe('true', () => {
```
alexs-mparticle, line 50:
```suggestion
        describe('noFunctional privacy flag set to true', () => {
```

test/jest/persistence.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+
+                mockMPInstance._Persistence.update();
+
+                expect(mockMPInstance._Persistence.setCookie).not.toHaveBeenCalled();
+                expect(mockMPInstance._Persistence.setLocalStorage).not.toHaveBeenCalled();
+                expect(window.localStorage.getItem('mprtcl-v4')).toBeNull();
+            });
+        });
+
+        describe('false', () => {
```
alexs-mparticle, line 77:
```suggestion
        describe('noFunctional privacy flag set to false', () => {
```

test/jest/persistence.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+                getCurrentUser: () => ({ getMPID: () => 'mpid' }),
+            },
+        } as unknown as IMParticleWebSDKInstance;
+
+        store = {} as IStore;
+        Store.call(store, {} as SDKInitConfig, mockMPInstance, 'apikey');
+        window.localStorage.removeItem('mprtcl-v4');
+    });
+
+    describe('#noFunctional privacy flag', () => {
```
alexs-mparticle, line 49:
```suggestion
    describe('#update', () => {
```

test/jest/persistence.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+
+                mockMPInstance._Persistence.update();
+
+                expect(mockMPInstance._Persistence.setCookie).not.toHaveBeenCalled();
+                expect(mockMPInstance._Persistence.setLocalStorage).toHaveBeenCalled();
+                expect(window.localStorage.getItem('mprtcl-v4')).not.toBeNull();
+            });
+        });
+
+        describe('is false by default', () => {
```
alexs-mparticle, line 104:
```suggestion
        describe('noFunctional privacy flag set to false by default', () => {
```

test/jest/persistence.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+            beforeEach(() => {
+                // default is false
+                store.webviewBridgeEnabled = false;
+            });
+
+            it('should write to cookie and localStorage by default when useCookieStorage is true', () => {
```
alexs-mparticle, line 110: Are you sure it writes to both localStorage AND cookie storage? It should only be one or the other?

test/src/tests-persistence.ts
```diff
@@ -1948,4 +1948,90 @@ describe('persistence', () => {
         user2.getAllUserAttributes()['ua-list'][1].should.equal('<b>');
         user2.getAllUserAttributes()['ua-1'].should.equal('a');
     });
+
+    describe('#noFunctional', () => {
```
alexs-mparticle, line 1952:
```suggestion
    describe('noFunctional privacy flag', () => {
```

test/src/tests-persistence.ts
```diff
@@ -1948,4 +1948,90 @@ describe('persistence', () => {
...
+        beforeEach(() => {
+            mParticle._resetForTests(MPConfig);
+        });
+
+        describe('true', () => {
```
alexs-mparticle, line 1957:
```suggestion
        describe('set to true', () => {
```

test/src/tests-persistence.ts
```diff
@@ -1948,4 +1948,90 @@ describe('persistence', () => {
...
+                mParticle.init(apiKey, mParticle.config);
+                await waitForCondition(hasIdentifyReturned);
+
+                mParticle.getInstance()._Persistence.update();
+
+                expect(getLocalStorage()).to.not.be.ok;
+            });
+        });
+
+        describe('false', () => {
```
alexs-mparticle, line 1985:
```suggestion
        describe('set to false', () => {
```

test/jest/persistence.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+                }),
+                update: function () {
+                    if (store.webviewBridgeEnabled || store.getNoFunctional()) {
+                        return;
+                    }
+                    if (store.SDKConfig.useCookieStorage) {
+                        mockMPInstance._Persistence.setCookie();
+                    }
+                    mockMPInstance._Persistence.setLocalStorage();
+                },
```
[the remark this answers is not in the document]
jaissica12: Got it, as discussed in person I have added `persistence` object. Let me know your feedback on updated tests.

src/persistence.js
```diff
@@ -20,6 +20,9 @@ export default function _Persistence(mpInstance) {
     };
 
     this.initializeStorage = function() {
+        if (mpInstance._Store.getPrivacyFlag('SDKState')) {
```
rmi22186, line 23: I think we still would want to know that `mpInstance._Store.isFirstRun = true` based on if the customer had been here in the past prior to setting the nolocation and notargeting.  We don't actively delete the cookies, (and while this information is likely not valuable). We'd send a boolean for isFirstRun to the backend rather than null which is what is happening now.

Image

rmi22186 (changes requested):
one minor nit

rmi22186 (approved):

PR #1072, opened 2025-09-17
feat: added DisabledVault to disable id-cache when noFunctional is set

jaissica12:
## Instructions

 ## Summary
 - Created a new DisabledVault which is used when noFunctional is set true.
 - Added jest and integration tests for it

 ## Testing Plan

 - Closes https://rokt.atlassian.net/browse/SDKE-132

### DIFF
--- a/test/src/tests-identity.ts
+++ b/test/src/tests-identity.ts
@@ -657,14 +657,14 @@ describe('identity', function() {
             });
 
             it('should NOT save id cache to local storage when noFunctional is true', async () => {
-                mParticle.config.noFunctional = true;
+                mParticle.config.launcherOptions = { noFunctional: true };
                 mParticle.init(apiKey, window.mParticle.config);
                 await waitForCondition(hasIdentifyReturned);
                 expect(localStorage.getItem('mprtcl-v4_abcdef-id-cache')).not.to.be.ok;
             });
 
             it('should save id cache to local storage when noFunctional is false', async () => {
-                mParticle.config.noFunctional = false;
+                mParticle.config.launcherOptions = { noFunctional: false };
                 mParticle.init(apiKey, window.mParticle.config);
                 await waitForCondition(hasIdentifyReturned);
                 expect(localStorage.getItem('mprtcl-v4_abcdef-id-cache')).to.be.ok;


src/mp-instance.ts
```diff
@@ -1534,9 +1534,12 @@ function createKitBlocker(config, mpInstance) {
 }
 
 function createIdentityCache(mpInstance) {
-    return new LocalStorageVault(`${mpInstance._Store.storageName}-id-cache`, {
-        logger: mpInstance.Logger,
-    });
+    const cacheKey = `${mpInstance._Store.storageName}-id-cache`;
+    if (mpInstance._Store.getNoFunctional()) {
+        console.log('createIdentityCache: noFunctional is true');
```
rmi22186, line 1539: remove
```suggestion
```

test/src/tests-identity.ts
```diff
@@ -642,6 +642,36 @@ describe('identity', function() {
...
 
+    describe('privacy flags', function() {
+        beforeEach(function() {
+            mParticle._resetForTests(MPConfig);
+            mParticle.config.flags.cacheIdentity = 'True';
+            localStorage.clear();
+        });
+
+        describe('#createIdentityCache', function() {
+            it('should use LocalStorageVault when noFunctional is false by default', async () => {
```
rmi22186, line 653: you aren't really confirming that the vault is or isn't localstorage vs disablevault.

test/src/tests-identity.ts
```diff
@@ -642,6 +642,36 @@ describe('identity', function() {
...
+                mParticle.init(apiKey, window.mParticle.config);
+                await waitForCondition(hasIdentifyReturned);
+                expect(localStorage.getItem('mprtcl-v4_abcdef-id-cache')).to.be.ok;
+            });
+
+            it('should use DisabledVault when noFunctional is true', async () => {
+                mParticle.config.noFunctional = true;
```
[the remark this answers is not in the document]
jaissica12: @rmi22186 Currently `noFunctional` / `noTargeting` is only read from top‑level `mParticle.config` during `init`, not from `config.launcherOptions` that is why `mParticle.config.noTargeting = true`; works. To use `mParticle.config.launcherOptions = {noTargeting: true};`  we need to update store to source privacy flags from `config.launcherOptions`.

PR #1073, opened 2025-09-29
refactor: SDKE 283 remove cart related function from persistence

jaissica12:
## Instructions

 ## Summary
 - Removed functions related to cart products across code since that is deprecated.

 ## Testing Plan

 - Closes https://rokt.atlassian.net/browse/SDKE-283


alexs-mparticle (approved):
Check with @rmi22186 for a second review since this touches a lot of older code, but it looks good to me.

src/identity.js
```diff
@@ -1363,56 +1314,16 @@ export default function Identity(mpInstance) {
              * @param {Boolean} [logEvent] a boolean to log adding of the cart object. If blank, no logging occurs.
              * @deprecated
              */
+            // eslint-disable-next-line no-unused-vars
```
rmi22186, line 1317: Instead of this comment, you can just remove the arguments below

src/identity.js
```diff
@@ -1286,6 +1286,7 @@ export default function Identity(mpInstance) {
      * @class mParticle.Identity.getCurrentUser().getCart()
      * @deprecated
      */
+    // eslint-disable-next-line no-unused-vars
```
rmi22186, line 1289: Instead of this, you can just remove the arguments below

this applies to all areas you added this lint check

src/utils.ts
```diff
@@ -57,18 +57,25 @@ const findKeyInObject = (obj: any, key: string): string => {
...
-    const messageArray: string[] = [
-        methodName,
-        Messages.DeprecationMessages.MethodIsDeprecatedPostfix,
-    ];
+    const messageArray: string[] = [methodName];
+
+    if (isDeprecated) {
+        messageArray.push(Messages.DeprecationMessages.MethodHasBeenDeprecated);
+    } else {
+        messageArray.push(Messages.DeprecationMessages.MethodMarkedForDeprecationPostfix);
```
rmi22186, line 69: `MethodMarkedForDeprecationPostfix` doesn't exist

src/utils.ts
```diff
@@ -57,18 +57,25 @@ const findKeyInObject = (obj: any, key: string): string => {
...
-        messageArray.push(alternateMethod);
-        messageArray.push(
-            Messages.DeprecationMessages.MethodIsDeprecatedPostfix
-        );
+        messageArray.push(Messages.DeprecationMessages.AlternativeMethodPrefix);
+        messageArray.push(alternateMethod + ".");
+    }
+
+    if (docsUrl) {
+        messageArray.push("See - " + docsUrl);
```
rmi22186, line 78:
```suggestion
        messageArray.push(" See - " + docsUrl);
```

src/utils.ts
rmi22186: Whoops, didn't have the latest commits locally and was looking at local code

test/src/tests-eCommerce.js
```diff
@@ -1690,7 +1521,7 @@ describe('eCommerce', function() {
 
             bond.called.should.eql(true);
             bond.getCalls()[0].args[0].should.eql(
-                'Deprecated function eCommerce.Cart.add() will be removed in future releases'
+                'Deprecated function eCommerce.Cart.add()'
```
rmi22186, line 1524: Do these deprecation messages need to be changed?

test/src/tests-identity.ts
```diff
@@ -2059,159 +2058,6 @@ describe('identity', function() {
         done();
     });
 
-    it("should find the related MPID's cookies when given a UI with fewer IDs when passed to login, logout, and identify, and then log events with updated cookies", async () => {
```
rmi22186, line 2062: This test can remain.  but you can remove the areas associated with products here.  There's still checks for other non-product items that are probably important

src/utils.ts
jaissica12: @Rob On line 81, we use `messageArray.join(' ')`, which inserts a space between each element. If we apply this after a period, it will result in two consecutive spaces. Current message: `Identity.getCurrentUser().getCart().getCartProducts() has been deprecated. Please use the alternate method: eCommerce.logProductAction(). See - https://docs.mparticle.com/developers/sdk/web/commerce-tracking`

test/src/tests-native-sdk.js
```diff
@@ -1323,71 +1105,6 @@ describe('native-sdk methods', function() {
                 done();
             });
 
-            it('should send an event with a product list when calling logPurchase', function(done) {
```
rmi22186, line 1326: I think this test should remain also.  logPurchase is technically deprecated too, but still works and removing it isn't part of this PR.  you can probably remove lines 1340-1349, but the rest of the test should remain. The crux of this test is for logPurchase and it happens to have some cart.add in it, but all the other non cart.add logic should remain for now

rmi22186 (changes requested):
mostly nits, with one comment that might change some of the work.  Let's way to hear back from Brandon/Mansi

src/utils.ts
rmi22186: Missed that.  Thanks for clarifying

src/identity.js
jaissica12: Got it, I wasn't sure if we want to update the method definitions. Removed the arguments, updated the interface and related calls.

rmi22186:
LGTM.  I'd say let's confirm 100% with mansi/brandon before merging, but I'm pretty confident about it given the docs.

rmi22186 (approved):

PR #1076, opened 2025-09-30
fix: SDKE-317 Call deferred methods on Rokt Manager instead of kit

rmi22186:
## Instructions

 ## Summary
Deferred calls happen if mParticle is ready but Rokt isn’t yet and a developer tries to call `mParticle.Rokt.METHOD_NAME`.  Eventually Rokt will become ready, and the deferred calls will be processed.  When that happens, the current code has the deferred method called on the kit and not the Rokt Manager.  This makes sense for `use` and `hashAttributes` because we simply pass the data directly to the kit method.  However, it doesn't work with `selectPlacements` because we do lots of logic between RoktManager’s selectPlacements and the kit’s selectPlacements.  We need to call the method on the Rokt Manager and not the kit.

 ## Testing Plan
Unit tests and local test app

 - Closes https://go/jSDKE-317

### DIFF
--- a/src/roktManager.ts
+++ b/src/roktManager.ts
@@ -319,15 +319,15 @@ export default class RoktManager {
         this.logger?.verbose(`RoktManager: Processing ${this.messageQueue.size} queued messages`);
 
         this.messageQueue.forEach((message) => {
-            if(!(message.methodName in this.kit) || !isFunction(this.kit[message.methodName])) {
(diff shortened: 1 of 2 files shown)


alexs-mparticle (changes requested):

alexs-mparticle (approved):

jaissica12 (approved):
LGTM

PR #1099, opened 2025-10-13
test: resolve failing tests after converting .then to async/await

jaissica12:
## Background
- We refactored all tests to use async/await instead of `.then`. Following that change, a few tests started failing. This PR addresses and resolves those failing tests.

## What Has Changed
- Updated tests
   - test/src/tests-batchUploader.ts
   - test/src/tests-batchUploader_3.ts
   - test/src/tests-event-logging.js
   - test/src/tests-legacy-alias-requests.ts
   - test/src/tests-session-manager.ts

## Screenshots/Video

## Checklist

## Additional Notes
- Will remove comment with ticket name `http://go/j-SDKE-301` once PR is reviewed.

- Closes https://rokt.atlassian.net/browse/SDKE-301

### DIFF
--- a/test/src/tests-session-manager.ts
+++ b/test/src/tests-session-manager.ts
@@ -325,6 +325,39 @@ describe('SessionManager', () => {
                 expect(persistenceSpy.called).to.equal(true);
             });
 
+            it('should NOT end a session if NOT enough time has passed in between sessions', () => {
+                // The default timeout limit is 30 minutes.
+                const twentyMinutesAgo = new Date();
+                twentyMinutesAgo.setMinutes(now.getMinutes() - 20);
+
+                mParticle.init(apiKey, window.mParticle.config);
+                const mpInstance = mParticle.getInstance();
+                const timerSpy = sinon.spy(
+                    mpInstance._SessionManager,
+                    'setSessionTimer'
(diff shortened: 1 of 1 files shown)


test/src/tests-session-manager.ts
```diff
@@ -493,8 +493,7 @@ describe('SessionManager', () => {
                 );
             });
 
-            it('should NOT end session if session has not timed out', () => {
-                const now = new Date();
+            it('should NOT end a session if enough time has passed in between sessions', () => {
```
rmi22186:
```suggestion
            it('should NOT end a session if NOT enough time has passed in between sessions', () => {
```

test/src/tests-session-manager.ts
```diff
@@ -493,8 +493,7 @@ describe('SessionManager', () => {
                 );
             });
 
-            it('should NOT end session if session has not timed out', () => {
-                const now = new Date();
+            it('should NOT end a session if NOT enough time has passed in between sessions', () => {
```
rmi22186: move this up to be right after  `should end a session if enough time has passed in between sessions`  for context consistency

rmi22186 (approved):

PR #1115, opened 2025-11-14
fix: Append domain query parameter to Trade Desk cookie syncs

rmi22186:
## Background
Our cookie sync builds a url by appending settings from the pixel configuration.  It was discovered that The Trade Desk (TTD) requires a `domain` parameter when a user is subject to GDPR (ie. user accesses the internet from the UK). 

The parameter is optional, but to keep things simple, we will be adding it to each time a TTD pixel needs to be fired. 

## What Has Changed
This PR adds updates to:
* Cookie sync manager - added constants for each potential cookie sync partner we have (you can reference the moduleIds and names from this [config endpoint](https://jssdkcdns.mparticle.com/js/v2/e207c24e36a7a8478ba0fcb3707a616b/config) which contains all of them - it is pretty printed below as well), adds a check for TTD and adds the domain if so
* Utils - update the cookie sync manager's URL creation by adding an optional domain parameter

## Screenshots/Video
Pretty printed config:
image

image

## Checklist

## Additional Notes

- Closes https://go/j/SDKE-335

### DIFF
--- a/src/cookieSyncManager.ts
+++ b/src/cookieSyncManager.ts
@@ -13,6 +13,16 @@ const { InformationMessages } = Messages;
 
 export const DAYS_IN_MILLISECONDS = 1000 * 60 * 60 * 24;
 
+// Partner module IDs for cookie sync configurations
+export const PARTNER_MODULE_IDS = {
+    AdobeEventForwarder: 11,
+    DoubleclickDFP: 41,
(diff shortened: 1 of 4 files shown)


jaissica12 (approved):
LGTM! Thanks for adding the screenshot of the potential cookie sync partner, it was helpful for the review.

PR #1117, opened 2025-11-17
fix: Ensure transitory DAS for launcherOptions settings

rmi22186:
## Background
A DAS is required for the identity framework to succeed for anonymous users. Previously no DAS was created due to `initializeStorage` returning early under a newly introduced code path for a `noFunctional` setting for `launcherOptions` . With this PR, a DAS is created for the purposes of the identity call succeeding but is never saved to persistence.

## What Has Changed
- Call `storeDataInMemory` with no arguments which reuses existing code to create a new DAS.

## Checklist

- Closes https://go/j/SDKE-602


jaissica12 (approved):
LGTM 🚀

samdozor (changes requested):
as we chatted - this mapping shouldn't exist. we can bring it back if we want but the guidance for clients should be that if they do, they need to be setting a persistent ID, otherwise we will blow up identity

PR #1123, opened 2025-11-25
feat: Create Logging Service SDKE-511

gtamanaha:
## Background
This PR introduces a new server-side logging service that enables the mParticle Web SDK to send error and warning logs to mParticle's backend for improved monitoring and debugging. The implementation includes creating a new `ReportingLogger` service that can send logs to the server when enabled.

`Important`: Next PR will include the reporting service call from Logger.ts

### Major Features
- New ReportingLogger service for server-side log submission
- API Client enhanced with sendLogToServer() method
- New logging infrastructure (logRequest, errorCodes, etc.)
- Configuration option isWebSdkLoggingEnabled added
- Test coverage added for all new functionality
### Technical Highlights
- Rate limiting: 10 logs per severity type
- Conditional enabling: Only active when Rokt domain is present + feature flag or debug mode

## Screenshots/Video

## Checklist
- [] I have made corresponding changes to the documentation.

## Additional Notes

- Closes https://go/j/SDKE-511

### DIFF
--- a/src/logging/reportingLogger.ts
+++ b/src/logging/reportingLogger.ts
@@ -16,6 +16,7 @@ export class ReportingLogger implements IReportingLogger {
     
     constructor(
         private baseUrl: string,
+        private readonly isFeatureFlagEnabled: boolean,
         private readonly sdkVersion: string,
         private readonly accountId: string,
         private readonly roktLauncherInstanceGuid: string,
@@ -64,20 +65,13 @@ export class ReportingLogger implements IReportingLogger {
     private isReportingEnabled(): boolean {
         return (
             this.isRoktDomainPresent() && 
-            (this.isFeatureFlagEnabled() ||
+            (this.isFeatureFlagEnabled ||
             this.isDebugModeEnabled())
         );
     }
 
     private isRoktDomainPresent(): boolean {
         return Boolean(window['ROKT_DOMAIN']);
-    }
-
-    private isFeatureFlagEnabled(): boolean {
-        return window.
-                mParticle?.
-                config?.
-                isWebSdkLoggingEnabled ?? false;
     }
 
     private isDebugModeEnabled(): boolean {
--- a/test/jest/reportingLogger.spec.ts
+++ b/test/jest/reportingLogger.spec.ts
@@ -25,7 +25,7 @@ describe('ReportingLogger', () => {
             ROKT_DOMAIN: 'set',
             fetch: mockFetch
         });
-        logger = new ReportingLogger(baseUrl, sdkVersion, accountId, roktLauncherInstanceGuid);
+        logger = new ReportingLogger(baseUrl, true, sdkVersion, accountId, roktLauncherInstanceGuid);
     });
 
     afterEach(() => {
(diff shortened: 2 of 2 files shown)


src/logging/errorCodes.ts
```diff
@@ -0,0 +1,5 @@
+export type ErrorCodes = (typeof ErrorCodes)[keyof typeof ErrorCodes];
+
+export const ErrorCodes = {
```
alexs-mparticle: I think this should be `ErrorMessages`. `ErrorCodes` sounds like they'd be something numerical like HTTP Error Codes (404, 429, etc).

Are you naming this `ErrorMessages` because that's something defined on the server side?

src/logging/reportingLogger.ts
```diff
@@ -0,0 +1,153 @@
...
+    private canSendLog(severity: LogRequestSeverity): boolean {
+        return this.isEnabled && !this.isRateLimited(severity);
+    }
+
+    private isRateLimited(severity: LogRequestSeverity): boolean {
+        return this.rateLimiter.incrementAndCheck(severity);
+    }
+
+    private getUrl(): string {
+        return window.location.href;
```
alexs-mparticle: In these cases where you're using `window`, you should first check if `window` exists, especially if this module is going to log errors that are independent of the SDK itself.

We've had problems in the past of customers and partners using the Web SDK server side and `window` will not exist.

test/jest/reportingLogger.spec.ts
```diff
@@ -0,0 +1,147 @@
...
+    let logger: ReportingLogger;
+    const sdkVersion = '1.2.3';
+    let mockFetch: jest.Mock;
+    const accountId = '1234567890';
+    beforeEach(() => {
+        mockFetch = jest.fn().mockResolvedValue({ ok: true });
+        global.fetch = mockFetch;
+        
+        mpInstance = {
+            _Helpers: {
```
alexs-mparticle: This may be future work but if it's easy enough, we should rewrite `createServiceUrl` so that it isn't as tightly coupled to the full SDK instance or Store, or make it a function within `Store`.

If it's future work, we should collect some of these refactors into a launch pad.

src/logging/logRequest.ts
```diff
@@ -0,0 +1,21 @@
+import { ErrorCodes } from "./errorCodes";
+
+export enum LogRequestSeverity {
```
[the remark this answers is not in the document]
gtamanaha: I would prefer to keep them separated because these ones depends on the API contract.

src/logging/reportingLogger.ts
```diff
@@ -0,0 +1,153 @@
...
+
+export interface IRateLimiter {
+    incrementAndCheck(severity: LogRequestSeverity): boolean;
+}
+
+export class RateLimiter implements IRateLimiter {
+    private readonly rateLimits: Map<LogRequestSeverity, number> = new Map([
+        [LogRequestSeverity.Error, 10],
+        [LogRequestSeverity.Warning, 10],
+        [LogRequestSeverity.Info, 10],
```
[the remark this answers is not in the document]
gtamanaha: I would like to discuss if these amounts are good amount first. @mattbodle @rmi22186 what do you think about these limits?

src/logging/reportingLogger.ts
gtamanaha: Added null conditional operator

src/logging/reportingLogger.ts
```diff
@@ -0,0 +1,148 @@
...
+        private readonly sdkVersion: string,
+        private readonly accountId: string,
+        rateLimiter?: IRateLimiter,
+    ) {
+        this.isEnabled = this.isReportingEnabled();
+        this.rateLimiter = rateLimiter ?? new RateLimiter();
+    }
+
+    public error(msg: string, code?: ErrorCodes, stackTrace?: string) {
+        this.sendLog(LogRequestSeverity.Error, msg, code ?? ErrorCodes.UNHANDLED_EXCEPTION, stackTrace);
```
mattbodle: Can we make code a mandatory parameter?
UNHANDLED_EXCEPTION is a code we specifically use when an error escapes any catch clause and is receeived by the window

PR #1124, opened 2025-12-02
[WIP] feat: Invoke Reporting Service from Logger

gtamanaha:
## Background
Adds server-side error reporting to the mParticle Web SDK. SDK errors and warnings are automatically sent to the /v1/log endpoint for centralized monitoring.

## What Has Changed
- Add Error codes to existing logger invocation. 

## Screenshots/Video

## Checklist

## Additional Notes

### DIFF
--- a/src/mp-instance.ts
+++ b/src/mp-instance.ts
@@ -1553,7 +1553,7 @@ function createIdentityCache(mpInstance) {
 }
 
 function runPreConfigFetchInitialization(mpInstance, apiKey, config) {
-    const reportingLogger = new ReportingLogger(Constants.sdkVersion, mpInstance);
+    const reportingLogger = new ReportingLogger(mpInstance, Constants.sdkVersion);
     mpInstance.Logger = new Logger(config, reportingLogger);
     mpInstance._Store = new Store(config, mpInstance, apiKey);
     window.mParticle.Store = mpInstance._Store;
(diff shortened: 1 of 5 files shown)


src/logger.ts
```diff
@@ -1,15 +1,21 @@
...
+import { IReportingLogger } from './logging/reportingLogger';
+import { ErrorCodes } from './logging/errorCodes';
 
 export type ILoggerConfig = Pick<SDKInitConfig, 'logLevel' | 'logger'>;
 export type IConsoleLogger = Partial<Pick<SDKLoggerApi, 'error' | 'warning' | 'verbose'>>;
 
 export class Logger {
     private logLevel: LogLevelType;
     private logger: IConsoleLogger;
+    private reportingLogger: IReportingLogger;
```
alexs-mparticle: From what I've seen, the error code is optional, so shouldn't this also be optional?

src/logging/errorCodes.ts
```diff
@@ -2,4 +2,22 @@ export type ErrorCodes = (typeof ErrorCodes)[keyof typeof ErrorCodes];
 
 export const ErrorCodes = {
     UNHANDLED_EXCEPTION: 'UNHANDLED_EXCEPTION',
+    EMAIL_MISMATCH: 'EMAIL_MISMATCH',
```
alexs-mparticle: Talked this over with @mattbodle and @rmi22186 and I think we may need to review the existing error codes for the WSDK and see how they overlap. There's a lot of error codes here that are specific to the mP SDK modules, but may not be relevant to our metrics and logs in the dashboard.

PR #1244, opened 2026-04-08
feat: support batch forwarding in kits

jaissica12:
## Background
- Kits currently only receive individual events via forwarder.process(event). Some kits need access to the full batch object (grouped by MPID and session, in Events API format) to support batch-level processing.

## What Has Changed
- Added optional `processBatch(batch)` method to the ConfiguredKit interface, kits that implement it receive batches, others are unaffected
- Added `sendBatchToForwarders(batch)` in `forwarders.js` to route batches to opted-in kits
- Wired batch forwarding into `createNewBatches` in `batchUploader.ts`, after `convertEvents` but before `onCreateBatch`, so kits get the unmodified batch

## Screenshots/Video

## Checklist

## Additional Notes

- Closes NA

### DIFF
--- a/src/forwarders.interfaces.ts
+++ b/src/forwarders.interfaces.ts
@@ -1,4 +1,3 @@
-import { Batch } from '@mparticle/event-models';
 import { SDKEvent, SDKEventCustomFlags } from './sdkRuntimeModels';
 import { Dictionary } from './utils';
 import { IKitConfigs, IKitFilterSettings } from './configAPIClient';
@@ -82,7 +81,6 @@ export interface ConfiguredKit
     ): string;
     onUserIdentified(user: IMParticleUser): string;
     process(event: SDKEvent): string;
-    processBatch?(batch: Batch): string;
     setOptOut(isOptingOut: boolean): string;
     removeUserAttribute(key: string): string;
     setUserAttribute(key: string, value: string): string;
--- a/src/forwarders.interfaces.ts
+++ b/src/forwarders.interfaces.ts
@@ -1,3 +1,4 @@
+import { Batch } from '@mparticle/event-models';
 import { SDKEvent, SDKEventCustomFlags } from './sdkRuntimeModels';
 import { Dictionary } from './utils';
 import { IKitConfigs, IKitFilterSettings } from './configAPIClient';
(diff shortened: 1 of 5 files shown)


src/forwarders.interfaces.ts
```diff
@@ -81,6 +82,7 @@ export interface ConfiguredKit
     ): string;
     onUserIdentified(user: IMParticleUser): string;
     process(event: SDKEvent): string;
+    processBatch?(batch: Batch): string;
```
rmi22186: I added a new `KitInterface` because it's different from `ConfiguredKit`.  Once you rebase against main, this file should have no updates anymore., because I [imported Batch](https://github.com/mParticle/mparticle-web-sdk/blob/master/src/forwarders.interfaces.ts#L5) here and also and have [processBatch here](https://github.com/mParticle/mparticle-web-sdk/blob/master/src/forwarders.interfaces.ts#L127)

test/jest/batchUploader.spec.ts
```diff
@@ -41,6 +57,227 @@ describe('BatchUploader', () => {
...
+                getAllUserAttributes: () => ({}),
+            };
+
+            BatchUploader['createNewBatches'](
+                mockEvents as any,
+                mockUser as any,
+                mockMPInstance
+            );
+
+            expect(forwardedEventCount).toBeGreaterThan(0);
```
rmi22186: Any reason this isn't `toBeGreaterThan(1)`?

test/src/tests-forwarders.ts
```diff
@@ -4016,4 +4018,176 @@ describe('forwarders', function() {
...
+
+            const fakeBatch = {
+                events: [{ event_type: 'custom_event' }],
+                mpid: testMPID,
+            };
+
+            // Should not throw
+            mParticle
+                .getInstance()
+                ._Forwarders.sendBatchToForwarders(fakeBatch);
```
rmi22186: Should you somehow incorporate
` }).to.not.throw();`

if the test assertion is to handle errors gracefully?

jaissica12: good call, added it.

test/jest/batchUploader.spec.ts
jaissica12: Added `toBeGreaterThan(0) ` to prove the batch wasn't empty before `onCreateBatch` wiped events and `toBeGreaterThan(1)` would fail since we pass 1 event. Changed to `toBe(1)` since thats more precise.

src/forwarders.interfaces.ts
jaissica12: Updated

test/jest/batchUploader.spec.ts
rmi22186: Oh, yea that's what I meant.  to equal 1, but I copied and pasted.  Thanks!

rmi22186 (approved):

PR #1245, opened 2026-04-09
feat: enable data filtering for batches

jaissica12:
## Background
- Kits that implement processBatch receive the full unfiltered batch from the SDK. Unlike individual events (which go through sendEventToForwarders with 5 types of data filtering), batches were forwarded without any filtering applied. This meant kits receiving batches could see events, user attributes, user identities, and event attributes that should have been blocked or stripped based on dashboard-configured filter rules.

## What Has Changed
- Applied the same 5 data filters to `sendBatchToForwarders` that already exist in `sendEventToForwarders`: attribute forwarding rules, event name/type filters, event attribute filters, user identity filters, and user attribute filters
- Extracted shared filter helper functions (`isBlockedByForwardingRule`, `isBlockedByEventFilter`, `filterEventAttributes`, `filterUserIdentities`) from inline code in `sendEventToForwarders` so both event and batch paths reuse the same logic
- Added reverse-mapping functions in `sdkToEventsApiConverter.ts` (`getMessageTypeFromEventType`, `getEventCategoryFromCustomEventType`, `getIdentityTypeFromBatchKey`, `getEventNameFromBatchEvent`) to translate batch snake_case fields back to SDK numeric types needed by the filter functions
- If all events in a batch are filtered out, processBatch is skipped entirely for that forwarder

## Checklist

## Additional Notes

- Closes NA


src/forwarders.js
```diff
@@ -414,21 +276,115 @@ export default function Forwarders(mpInstance, kitBlocker) {
...
+
+                        const messageType = getMessageTypeFromEventType(
+                            batchEvent.event_type
+                        );
+                        const eventName = getEventNameFromBatchEvent(
+                            batchEvent
+                        );
+                        const eventCategory = getEventCategoryFromCustomEventType(
+                            batchEvent.data.custom_event_type
+                        );
```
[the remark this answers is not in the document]
jaissica12: good catch, resolved

src/sdkToEventsApiConverter.ts
```diff
@@ -795,3 +795,191 @@ export function convertUserIdentityTypeToServerIdentityType(
             return EventsApi.IdentityTypeEnum.phoneNumber3;
     }
 }
+
+export function getMessageTypeFromEventType(
```
alexs-mparticle: Nit: I'd add this to the `types.ts` file since we have similar functions there.

src/forwarders.js
```diff
@@ -831,3 +713,214 @@ export default function Forwarders(mpInstance, kitBlocker) {
...
+            );
+
+            if (inArray(filterList, hash)) {
+                delete attributes[attrName];
+            }
+        }
+    }
+}
+
+function filterUserIdentities(event, filterList) {
```
alexs-mparticle: Nit: I would extract these functions out of this file into a `forwarder-utils.ts` file, similar to `identity-utils.ts` so we can apply Type Definitions to them.

dist/types/src/forwarders.d.ts
```diff
@@ -7,6 +7,7 @@ export default class Forwarders {
     isEnabledForUnknownUser: (excludeAnonymousUserBoolean: any, user: any) => boolean;
     applyToForwarders: (functionName: any, functionArgs: any) => void;
     sendEventToForwarders: (event: any) => void;
+    sendBatchToForwarders: (batch: any) => void;
```
alexs-mparticle: Should you be committing this? I would assume the build task would do this for you.

src/forwarders.js
```diff
@@ -177,64 +183,7 @@ export default function Forwarders(mpInstance, kitBlocker) {
...
-                }
-
-                return false;
-            },
-            forwardingRuleMessageTypes = [
-                Types.MessageType.PageEvent,
-                Types.MessageType.PageView,
-                Types.MessageType.Commerce,
-            ];
+        let clonedEvent, hashedEventName, hashedEventType;
```
alexs-mparticle: Nit: put these on their own lines

src/forwarders.js
```diff
@@ -831,3 +713,214 @@ export default function Forwarders(mpInstance, kitBlocker) {
...
+    for (const attrName in attributes) {
+        if (attributes.hasOwnProperty(attrName)) {
+            const hash = KitFilterHelper.hashEventAttributeKey(
+                eventCategory,
+                eventName,
+                attrName
+            );
+
+            if (inArray(filterList, hash)) {
+                delete attributes[attrName];
```
alexs-mparticle: I know this is just moving stuff around, but I would love it if we could fast follow this with a small refactor to not mutate the original object.

test/src/tests-forwarders.ts
```diff
@@ -4190,5 +4190,669 @@ describe('forwarders', function() {
...
+                                        price: 19.99,
+                                        quantity: 1,
+                                    },
+                                ],
+                            },
+                            custom_attributes: {},
+                        },
+                    },
+                    {
+                        event_type: 'commerce_event',
```
alexs-mparticle, line 4347: I see you're writing tests for a bunch of event types, which is great, but I don't see one for `screen_view`

alexs-mparticle (changes requested):

dist/types/src/forwarders.d.ts
jaissica12: Good catch, this was accidentally included. The build task handles dist/ generation on merge. I'll remove it!

src/forwarders.js
jaissica12: done

src/sdkToEventsApiConverter.ts
jaissica12: Added to `types.ts`

src/forwarders.js
jaissica12: updated

test/src/tests-forwarders.ts
jaissica12: added test

src/forwarders.js
jaissica12: Refactored in the latest push. Both `filterEventAttributes` and `filterBatchIdentities` now return new objects instead of mutating their inputs:
- `filterEventAttributes` returns a new Record<string, unknown> with only the allowed attributes, rather than deleting blocked keys in-place.
- `filterBatchIdentities` now takes userIdentities and filterList as separate arguments and returns a new filtered object, instead of reaching into the batch to delete keys.

src/sdkToEventsApiConverter.ts
```diff
@@ -795,3 +795,65 @@ export function convertUserIdentityTypeToServerIdentityType(
...
     }
 }
+
+import { getEventCategoryFromCustomEventType } from './types';
+
+export {
+    getMessageTypeFromEventType,
+    getEventCategoryFromCustomEventType,
+    getIdentityTypeFromBatchKey,
+} from './types';
```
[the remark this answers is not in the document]
alexs-mparticle: @jaissica12 let's clean this up.

src/forwarder-utils.ts
```diff
@@ -0,0 +1,263 @@
...
+        return false;
+    }
+
+    (batchEvent as any).data.custom_attributes = filterEventAttributes(
+        messageType,
+        eventCategory,
+        eventName,
+        (batchEvent as any).data.custom_attributes,
+        forwarder
+    );
```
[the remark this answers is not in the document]
alexs-mparticle: @jaissica12 confirm if the robot is correct. Should be a quick fix.

src/forwarder-utils.ts
```diff
@@ -0,0 +1,263 @@
+import Types, {
+    EventType,
+    getMessageTypeFromEventType,
+    getIdentityTypeFromBatchKey,
+} from './types';
+import { inArray, valueof } from './utils';
+import KitFilterHelper from './kitFilterHelper';
+import * as EventsApi from '@mparticle/event-models';
+import { IKitFilterSettings } from './configAPIClient';
+import { SDKEvent } from './sdkRuntimeModels';
```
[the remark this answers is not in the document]
jaissica12: fixed

alexs-mparticle (approved):

src/forwarder-utils.ts
```diff
@@ -0,0 +1,262 @@
...
+    Types.MessageType.Commerce,
+];
+
+function inFilteredList(
+    filterList: number[] | undefined,
+    hash: number
+): boolean {
+    if (filterList && filterList.length) {
+        if (inArray(filterList, hash)) {
+            return true;
```
rmi22186: could you do:
```suggestion
        return inArray(filterList, hash);
```

src/forwarder-utils.ts
```diff
@@ -0,0 +1,262 @@
...
+    return false;
+}
+
+export function isBlockedByForwardingRule(
+    messageType: number,
+    attributes: Record<string, unknown> | null,
+    forwarder: ForwarderFilterConfig
+): boolean {
+    if (
+        FORWARDING_RULE_MESSAGE_TYPES.indexOf(messageType) === -1 ||
```
rmi22186:
```suggestion
        !FORWARDING_RULE_MESSAGE_TYPES.includes(messageType) ||
```

src/forwarder-utils.ts
jaissica12: updated

src/forwarder-utils.ts
jaissica12: done

src/forwarders.js
```diff
@@ -414,21 +280,50 @@ export default function Forwarders(mpInstance, kitBlocker) {
...
+                    });
+                }
+
+                batchCopy.user_identities = filterBatchIdentities(
+                    batchCopy.user_identities,
+                    forwarder.userIdentityFilters
+                );
+
+                if (batchCopy.user_attributes) {
+                    batchCopy.user_attributes = mpInstance._Helpers.filterUserAttributes(
```
rmi22186: Nit - is it possible to pull `filterUserAttributes` and `extend` off of mpInstance._Helpers, like you did for `inArray`? to clean this up further?

src/forwarders.js
```diff
@@ -414,21 +280,50 @@ export default function Forwarders(mpInstance, kitBlocker) {
...
                     );
-                    const result = forwarder.processBatch(batchCopy);
+                }
 
-                    if (result) {
-                        mpInstance.Logger.verbose(result);
-                    }
-                } catch (e) {
-                    mpInstance.Logger.verbose(e);
+                if (hadEvents && batchCopy.events.length === 0) {
```
rmi22186: This is a little awkward to me.  If a batch had events before, and now is empty, we continue.  Is there a scenario where this would be false && true?  I'm wondering if you need hadEvents at all here.

src/types.ts
```diff
@@ -397,6 +397,140 @@ export const PerformanceMarkType = {
...
+        case 'refund':
+            return CommerceEventType.ProductRefund;
+        case 'promotion_view':
+            return CommerceEventType.PromotionView;
+        case 'promotion_click':
+            return CommerceEventType.PromotionClick;
+        case 'add_to_wishlist':
+            return CommerceEventType.ProductAddToWishlist;
+        case 'remove_from_wishlist':
+        case 'remove_from_wish_list':
```
rmi22186, line 474: What's the reason for these 2 being here?

rmi22186: and not say, for `add_to_wish_list`

rmi22186:
Overall this is such a great improvement vs the mess that forwarders.js was prior.  Great work not just migrating to TS, but cleaning and refactoring the logic to make it more understandable.

src/types.ts
jaissica12: Only `removeFromWishlist` needs the dual case. Checked the `@mparticle/event-models` package, `ProductActionActionEnum.removeFromWishlist` is `remove_from_wish_list`"while `CommerceEventDataCustomEventTypeEnum.removeFromWishlist` is `remove_from_wishlist`. Both strings can flow into this switch via different code paths (`product_action.action` vs `custom_event_type`). 
The other actions like `addToWishlist` use `add_to_wishlist` consistently across both enums, so they don't need a duplicate case.

src/forwarders.js
jaissica12: Good catch, removed `hadEvents`. The batch uploader wouldn't create a batch with zero events. Simplified it

src/forwarders.js
jaissica12: Done. `filterUserAttributes` now uses `KitFilterHelper` directly. `extend` extracted to `utils.ts` as a standalone export

test/src/tests-forwarders.ts
```diff
@@ -4190,5 +4190,723 @@ describe('forwarders', function() {
...
+            // --- Event path: logEvent should be blocked ---
+            mParticle.logEvent(
+                'blocked event',
+                mParticle.EventType.Navigation
+            );
+            const eventResult = forwarderInstance.receivedEvent;
+            expect(
+                !eventResult ||
+                    eventResult.EventName !== 'blocked event'
+            ).to.equal(true);
```
rmi22186: This seems a little flakey-prone to me.  It means it's holdign an event from before.  You should delete anything that is on forwarderInstance.receivedEvent prior to logging the event above so that it isn't a conditional assertion.

test/src/tests-forwarders.ts
```diff
@@ -4190,5 +4190,723 @@ describe('forwarders', function() {
...
+            if (
+                eventResult &&
+                eventResult.UserIdentities &&
+                eventResult.UserIdentities.length
+            ) {
+                const hasEmail = eventResult.UserIdentities.some(
+                    (id: any) => id.Type === IdentityType.Email
+                );
+                expect(hasEmail).to.equal(false);
+            }
```
rmi22186: ditto to the above commen.t. we should not have a conditional assertion.  delete the receivedEvent prior to logging the event to ensure logging the event will definitely result in no event.

test/src/tests-forwarders.ts
```diff
@@ -4190,5 +4190,723 @@ describe('forwarders', function() {
...
+            const eventResult = forwarderInstance.receivedEvent;
+            if (eventResult && eventResult.UserAttributes) {
+                expect(eventResult.UserAttributes).to.not.have.property(
+                    'secret'
+                );
+                expect(eventResult.UserAttributes).to.have.property(
+                    'color',
+                    'blue'
+                );
+            }
```
rmi22186: ditto again

test/src/tests-forwarders.ts
jaissica12: Good catch, removing conditional assertions from forwarder tests

rmi22186 (approved):

PR #1247, opened 2026-04-10
feat: Pass $NoTargeting as User Attribute value when passed to launch…

rmi22186:
## Background
This feature enables customers who elect noTargeting to be filtered out of audiences for retargeting.  Server expects this to now be set as a value for the user attribute as `$NoTargeting` with a boolean `true`.

## What Has Changed
Add setUserAttribute in `parseIdentityResponse`.

## Screenshots/Video
image

## Checklist


src/identity.js
```diff
@@ -1581,6 +1581,11 @@ export default function Identity(mpInstance) {
         mpInstance._preInit.readyQueue = processReadyQueue(
             mpInstance._preInit.readyQueue
         );
+
+        var noTargeting = mpInstance._CookieConsentManager?.getNoTargeting();
```
mattbodle: As below - use const

mattbodle: Also not sure why your agent generated var, do we need to update agent file or something here?

rmi22186: This was actually me.  Agent had a solution that I wasn't happy with so I updated it, and I saw the file had var (prob should also be updated...and bc it was a js file I assumed we were only using const in our TS files), and didn't see the `const`s throughout as well. Updating.

khushi1033 (approved):
LGTM!

src/constants.ts
```diff
@@ -225,6 +225,8 @@ export default Constants;
 export const ONE_DAY_IN_SECONDS = 60 * 60 * 24;
 export const MILLIS_IN_ONE_SEC = 1000;
 
+export const NO_TARGETING_ATTRIBUTE = '$NoTargeting';
```
jamesnrokt, line 231: Nit: Is it worth adding docs above this? Essentially covering what was in your PR background

rmi22186: Good call.  Added!

PR #1248, opened 2026-04-13
refactor: migrate identity module to TS

jaissica12:
## Background
- The identity.js module was one of the remaining JavaScript files in the SDK. As part of the ongoing TypeScript migration effort, it needed to be converted to TypeScript to improve type safety and maintainability. This is a direct migration with no behavior, API, or breaking changes.

## What Has Changed
- Renamed src/identity.js to src/identity.ts with type annotations added to all functions, parameters, and variables
- Corrected pre-existing type mismatches in identity.interfaces.ts, identityApiClient.ts, persistence.interfaces.ts, store.ts, and mp-instance.ts where interfaces didn't match actual runtime behavior
- Updated test files to align with corrected interface types
- No runtime logic, public API, or deprecated behavior was changed — compiled output is functionally identical

## Testing
- All existing Karma + Mocha integration tests pass
- All existing Jest unit tests pass

## Screenshots/Video
- 
Screenshot 2026-05-20 at 4 01 44 PM
Screenshot 2026-05-20 at 4 02 06 PM
Screenshot 2026-05-20 at 4 02 42 PM
Screenshot 2026-05-20 at 4 03 04 PM

## Checklist

- Closes https://go/j/SDKE-1103

### DIFF
--- a/src/identity-utils.ts
+++ b/src/identity-utils.ts
@@ -34,7 +34,7 @@ export type IParseCachedIdentityResponse = (
     callback: IdentityCallback,
     identityApiData: IdentityApiData,
     identityMethod: string,
-    knownIdentities: IKnownIdentities,
+    knownIdentities: IKnownIdentities | UserIdentities,
     fromCachedIdentity: boolean
 ) => void;
 
--- a/src/identity-utils.ts
+++ b/src/identity-utils.ts
@@ -5,8 +5,8 @@ import Types from './types';
 import {
     IdentityApiData,
     UserIdentities,
-    IdentityCallback,
 } from '@mparticle/web-sdk';
+import { IdentityCallback } from './identity-user-interfaces';
 import { IdentityAPIMethod, IIdentityRequest } from './identity.interfaces';
 import {
     IdentityResultBody,
--- a/src/identity-utils.ts
+++ b/src/identity-utils.ts
@@ -33,7 +33,7 @@ export type IParseCachedIdentityResponse = (
     mpid: string,
     callback: IdentityCallback,
     identityApiData: IdentityApiData,
-    identityMethod: string,
+    identityMethod: IdentityAPIMethod,
     knownIdentities: IKnownIdentities,
     fromCachedIdentity: boolean
 ) => void;
--- a/src/identity-utils.ts
+++ b/src/identity-utils.ts
@@ -6,9 +6,9 @@ import {
     IdentityApiData,
     UserIdentities,
 } from '@mparticle/web-sdk';
-import { IdentityCallback } from './identity-user-interfaces';
 import { IdentityAPIMethod, IIdentityRequest } from './identity.interfaces';
 import {
+    IdentityCallback,
     IdentityResultBody,
     IIdentityResponse,
     IMParticleUser,
@@ -34,7 +34,7 @@ export type IParseCachedIdentityResponse = (
     callback: IdentityCallback,
     identityApiData: IdentityApiData,
     identityMethod: string,
-    knownIdentities: IKnownIdentities | UserIdentities,
+    knownIdentities: IKnownIdentities,
     fromCachedIdentity: boolean
 ) => void;
 
--- a/src/identity.interfaces.ts
+++ b/src/identity.interfaces.ts
@@ -1,4 +1,5 @@
 import { IdentityApiData, MPID, UserIdentities } from '@mparticle/web-sdk';
+import { Context } from '@mparticle/event-models';
 import AudienceManager from './audienceManager';
 import { ICachedIdentityCall, IKnownIdentities } from './identity-utils';
 import { BaseVault } from './vault';
@@ -18,6 +19,7 @@ const { platform, sdkVendor, sdkVersion, HTTPCodes } = Constants;
 export type IdentityPreProcessResult = {
     valid: boolean;
     error?: string;
+    cleanedIdentities?: IdentityApiData;
 };
 
 export type IdentityAPIMethod = valueof<typeof Constants.IdentityMethods>;
@@ -49,14 +51,14 @@ export enum SDKIdentityTypeEnum {
 
 export interface IIdentityAPIRequestData {
     client_sdk: {
-        platform: typeof platform;
-        sdk_vendor: typeof sdkVendor;
-        sdk_version: typeof sdkVersion;
+        platform: string;
+        sdk_vendor: string;
+        sdk_version: string;
     };
-    context: string | null;
+    context: Context | null;
     environment: Environment;
     request_id: string;
-    request_timestamp_unixtime_ms: number;
+    request_timestamp_ms: number;
     previous_mpid: MPID | null;
     known_identities: IKnownIdentities;
 }
@@ -86,7 +88,7 @@ export interface IIdentityRequest {
         sdkVendor: string,
         sdkVersion: string,
         deviceId: string,
-        context: string | null,
+        context: Context | null,
         mpid: MPID
     ): IIdentityAPIRequestData;
     createModifyIdentityRequest(
@@ -95,17 +97,20 @@ export interface IIdentityRequest {
         platform: string,
         sdkVendor: string,
         sdkVersion: string,
-        context: string | null
+        context: Context | null
     ): IIdentityAPIModifyRequestData;
     createIdentityChanges(
         previousIdentities: UserIdentities,
-        newIdentitie: UserIdentities
-    ): IIdentityAPIIdentityChangeData;
+        newIdentities: UserIdentities
+    ): IIdentityAPIIdentityChangeData[];
     preProcessIdentityRequest(
         identityApiData: IdentityApiData,
         callback: IdentityCallback,
         method: IdentityAPIMethod
     ): IdentityPreProcessResult;
+    createAliasNetworkRequest(aliasRequest: IAliasRequest): object;
+    convertAliasToNative(aliasRequest: IAliasRequest): object;
+    convertToNative(identityApiData: IdentityApiData): object | void;
 }
 
 export type AliasRequestScope = 'device' | 'mpid';
@@ -150,7 +155,7 @@ export interface SDKIdentityApi {
     getUsers?(): IMParticleUser[];
     aliasUsers?(
         aliasRequest?: IAliasRequest,
-        callback?: IdentityCallback
+        callback?: IAliasCallback
     ): void;
     createAliasRequest?(
         sourceUser: IMParticleUser,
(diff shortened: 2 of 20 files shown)


src/identity.ts
```diff
@@ -722,9 +756,9 @@ export default function Identity(mpInstance) {
                             ' -> ' +
                             aliasRequest.destinationMpid
                     );
-                    var aliasRequestMessage = mpInstance._Identity.IdentityRequest.createAliasNetworkRequest(
+                    const aliasRequestMessage = mpInstance._Identity.IdentityRequest.createAliasNetworkRequest(
                         aliasRequest
-                    );
+                    ) as IAliasRequest;
```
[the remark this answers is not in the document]
jaissica12: Will address this in follow-up branch `fix/SDKE-1103-alias-network-request-type` by introducing `IAliasNetworkRequest` interface and updates `sendAliasRequest` + tests to use it.

src/identity-utils.ts
```diff
@@ -5,8 +5,8 @@ import Types from './types';
 import {
     IdentityApiData,
     UserIdentities,
-    IdentityCallback,
 } from '@mparticle/web-sdk';
+import { IdentityCallback } from './identity-user-interfaces';
```
alexs-mparticle, line 9: Clean up the imports

src/identity-utils.ts
```diff
@@ -34,7 +34,7 @@ export type IParseCachedIdentityResponse = (
     callback: IdentityCallback,
     identityApiData: IdentityApiData,
     identityMethod: string,
-    knownIdentities: IKnownIdentities,
+    knownIdentities: IKnownIdentities | UserIdentities,
```
alexs-mparticle: I'm not sure if these overlap cleanly

src/identity.interfaces.ts
```diff
@@ -49,14 +51,14 @@ export enum SDKIdentityTypeEnum {
...
+        platform: string;
+        sdk_vendor: string;
+        sdk_version: string;
     };
-    context: string | null;
+    context: Context | null;
     environment: Environment;
     request_id: string;
-    request_timestamp_unixtime_ms: number;
+    request_timestamp_ms: number;
```
alexs-mparticle, line 61: Why are we renaming this?

src/identity.interfaces.ts
```diff
@@ -95,17 +97,20 @@ export interface IIdentityRequest {
...
         sdkVendor: string,
         sdkVersion: string,
-        context: string | null
+        context: Context | null
     ): IIdentityAPIModifyRequestData;
     createIdentityChanges(
         previousIdentities: UserIdentities,
-        newIdentitie: UserIdentities
-    ): IIdentityAPIIdentityChangeData;
+        newIdentities: UserIdentities
```
alexs-mparticle, line 104: How long has this typo been here? 😓

src/identity.interfaces.ts
```diff
@@ -211,18 +216,18 @@ export interface IIdentity {
...
-        previousUserAttributeValue: string,
+        newValue: string | string[] | null,
+        previousUserAttributeValue: string | string[] | null,
         isNewAttribute: boolean,
         deleted: boolean,
         user: IMParticleUser
     ): IUserAttributeChangeEvent;
     createUserIdentityChange(
-        identityType: SDKIdentityTypeEnum,
+        identityType: SDKIdentityTypeEnum | string,
```
alexs-mparticle: I don't think we should use `string` but really limit this to known types

alexs-mparticle (changes requested):

src/identity.ts
```diff
@@ -722,9 +756,9 @@ export default function Identity(mpInstance) {
                             ' -> ' +
                             aliasRequest.destinationMpid
                     );
-                    var aliasRequestMessage = mpInstance._Identity.IdentityRequest.createAliasNetworkRequest(
+                    const aliasRequestMessage = mpInstance._Identity.IdentityRequest.createAliasNetworkRequest(
```
crisryantan, line 764: This cast hides a real mismatch and is worth fixing before merge.

`createAliasNetworkRequest` returns the network wire shape:
```ts
{ request_id, request_type: 'alias', environment, data: { destination_mpid, source_mpid, start_unixtime_ms, end_unixtime_ms, device_application_stamp } }
```
But `IAliasRequest` is:
```ts
{ destinationMpid, sourceMpid, startTime, endTime, scope? }
```

Those are completely different shapes. The `as IAliasRequest` cast tells the compiler "trust me, it's an IAliasRequest" and from there `sendAliasRequest(aliasRequest: IAliasRequest)` is happy. At runtime `sendAliasRequest` just does `JSON.stringify(aliasRequest)` (`identityApiClient.ts:113`) so the actual wire format is still right, but the types are lying.

src/identity.interfaces.ts
```diff
@@ -95,17 +97,20 @@ export interface IIdentityRequest {
...
-        newIdentitie: UserIdentities
-    ): IIdentityAPIIdentityChangeData;
+        newIdentities: UserIdentities
+    ): IIdentityAPIIdentityChangeData[];
     preProcessIdentityRequest(
         identityApiData: IdentityApiData,
         callback: IdentityCallback,
         method: IdentityAPIMethod
     ): IdentityPreProcessResult;
+    createAliasNetworkRequest(aliasRequest: IAliasRequest): object;
```
crisryantan, line 111: Tied to the [earlier comment](https://github.com/mParticle/mparticle-web-sdk/pull/1248/changes#r3283373738). These three methods all return `object`, which is barely better than `any`. The migration goal is type safety, so this is a missed opportunity.

If you do above for `createAliasNetworkRequest`, the same pattern applies here, declare proper interfaces for the native shapes and use them. `convertToNative` returning `object | void` is especially confusing because `void` is meaningless as a return type union with anything else (it means "ignore the return value"), and the actual code returns `undefined` explicitly. `IdentityNativeRequest | undefined` would be clearer.

Happy to defer to a follow-up if you want to keep this PR scoped, but please file a ticket so it does not slip.

src/identity.ts
```diff
@@ -395,8 +423,9 @@ export default function Identity(mpInstance) {
                             mpInstance._Store.activeForwarders.forEach(function(
                                 forwarder
                             ) {
-                                if (forwarder.logOut) {
-                                    forwarder.logOut(evt);
+                                const fwd = forwarder as unknown as Record<string, Function>;
```
crisryantan: Why the double-cast to `Record<string, Function>`? That tells the compiler "this thing is a string-keyed bag of arbitrary functions" which is the loosest possible cast and erases all type information from `forwarder`.

I checked `ConfiguredKit` in `forwarders.interfaces.ts` does **not** have a `logOut` method. So either:

1. The old JS code (`if (forwarder.logOut)`) was always-falsy defensive code that never actually fired. If that is the case, this whole block can be deleted.
2. There is a `logOut` method present at runtime on some kit implementations that the `ConfiguredKit` interface does not describe. If that is the case, add `logOut?: (evt: SDKEvent) => void` to `ConfiguredKit` and the cast disappears.

src/identity-utils.ts
jaissica12: done

src/identity-utils.ts
jaissica12: Good catch. `IKnownIdentities | UserIdentities` collapses to plain `UserIdentities` at the type level (since `IKnownIdentities` extends `UserIdentities`), so the union added zero safety and actually hid `device_application_stamp`

src/identity.interfaces.ts
jaissica12: Yeah, it's really a typo fix on the interface. The SDK has always been sending request_timestamp_ms, the interface was just declaring the different key name

src/identity.interfaces.ts
jaissica12: Tightened to `SDKIdentityTypeEnum` on the interface and narrowed the loop

src/identity.ts
jaissica12: Agreed, the cast is hiding a real type drift. Already addressed in fix/SDKE-1103-alias-network-request-type branch

src/identity.interfaces.ts
jaissica12: Agreed, object is barely better than any, and object | void is genuinely confusing semantics.
Picking up the two alias-related ones on fix/SDKE-1103-alias-network-request-type:
createAliasNetworkRequest: : object → : IAliasNetworkRequest
convertAliasToNative: : object → : INativeAliasRequest (new interface — { DestinationMpid, SourceMpid, StartUnixtimeMs, EndUnixtimeMs, Scope? })
Creating a separate ticket (https://go/j/SDKE-1204) for convertToNative since it's a different code path (identity native bridge, not alias), will switch to INativeIdentityRequest | undefined per your suggestion

src/identity.ts
jaissica12: Option 2, logOut is real at runtime (MockForwarder defines it and the CBT test asserts logOutCalled === true after logout). Added logOut?: (evt: SDKEvent) => void to ConfiguredKit and dropped the cast.

src/persistence.interfaces.ts
```diff
@@ -120,10 +120,10 @@ export interface IPersistence {
     saveUserCookieSyncDatesToPersistence(mpid: MPID, csd: CookieSyncDates): void;
     savePersistence(persistance: IPersistenceMinified): void;
```
mattbodle: nit: Existing typo here

src/mp-instance.ts
mattbodle, line 141: Why is there a type error here?
Two things:
- Does the type check block the build?
- If this is accurate, does Persistance need to be updated to implement IPersistence?

src/persistence.interfaces.ts
jaissica12: Fixed — param was a typo in IPersistence

src/mp-instance.ts
jaissica12: Yes, `IPersistence.getLocalStorage` is `IPersistenceMinified`, but `persistence.js` is inferred as {}. It doesn’t fail CI (tsc || true). Aligning Persistence with `IPersistence` is the persistence migrate

crisryantan (approved):

mattbodle (approved):

src/identity.ts
```diff
@@ -396,8 +423,11 @@ export default function Identity(mpInstance) {
                             mpInstance._Store.activeForwarders.forEach(function(
                                 forwarder
                             ) {
-                                if (forwarder.logOut) {
-                                    forwarder.logOut(evt);
+                                const logOut = (forwarder as {
+                                    logOut?: (event: SDKEvent) => void;
+                                }).logOut;
+                                if (logOut) {
+                                    logOut(evt);
```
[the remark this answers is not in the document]
jaissica12: fixed

src/identity.ts
```diff
@@ -328,7 +351,9 @@ export default function Identity(mpInstance) {
                     HTTPCodes.validationIssue,
                     preProcessResult.error
                 );
-                mpInstance.Logger.verbose(preProcessResult);
+                mpInstance.Logger.verbose(
+                    preProcessResult as unknown as string
+                );
```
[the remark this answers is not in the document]
jaissica12: Intentional, to match master behavior

PR #1252, opened 2026-04-21
feat: Add alias for sha256 identities

jamesnrokt:
## Background
- Currently clients have to use `other` for Sha256 identity types. This PR adds explicit aliases to other for clarity 

## What Has Changed
- Add aliases for sha256 identity types

## Screenshots/Video
- N/A

## Checklist

## Additional Notes
- N/A

- Closes N/A

### DIFF
--- a/src/identity-utils.ts
+++ b/src/identity-utils.ts
@@ -289,24 +289,31 @@ const getExpireTimestamp = (maxAge: number = ONE_DAY_IN_SECONDS): number =>
 const parseIdentityResponse = (responseText: string): IdentityResultBody =>
     responseText ? JSON.parse(responseText) : ({} as IdentityResultBody);
 
+type Sha256IdentityAlias = 'email_sha256' | 'mobile_sha256';
+
 const SHA256_IDENTITY_ALIASES: Readonly<
-    Partial<Record<keyof UserIdentities, keyof UserIdentities>>
+    Record<Sha256IdentityAlias, keyof UserIdentities>
 > = {
     email_sha256: 'other',
     mobile_sha256: 'other2',
 };
 
+type UserIdentitiesWithAliases = UserIdentities &
+    Partial<Record<Sha256IdentityAlias, string | null>>;
+
 export const normalizeUserIdentityKeys = (
-    userIdentities: UserIdentities
+    userIdentities: UserIdentitiesWithAliases
 ): UserIdentities => {
-    const normalized: UserIdentities = { ...userIdentities };
-    for (const alias of Object.keys(SHA256_IDENTITY_ALIASES)) {
-        if (alias in normalized) {
-            const value = normalized[alias];
-            delete normalized[alias];
-            normalized[SHA256_IDENTITY_ALIASES[alias]] = value;
+    const normalized: UserIdentitiesWithAliases = { ...userIdentities };
+    (Object.keys(SHA256_IDENTITY_ALIASES) as Sha256IdentityAlias[]).forEach(
+        (alias) => {
+            if (alias in normalized) {
+                const value = normalized[alias];
+                delete normalized[alias];
+                normalized[SHA256_IDENTITY_ALIASES[alias]] = value;
+            }
         }
-    }
+    );
     return normalized;
 };
 
--- a/src/identity-utils.ts
+++ b/src/identity-utils.ts
@@ -296,11 +296,9 @@ const getExpireTimestamp = (maxAge: number = ONE_DAY_IN_SECONDS): number =>
 const parseIdentityResponse = (responseText: string): IdentityResultBody =>
     responseText ? JSON.parse(responseText) : ({} as IdentityResultBody);
 
-// Maps convenience identity key names to their canonical server-side names.
-// Mirrors the emailSha256/mobileSha256 helpers added in the Apple SDK (PR #756).
 const SHA256_IDENTITY_ALIASES: Readonly<Record<string, string>> = {
     email_sha256: 'other',
-    mobile_sha256: 'other',
+    mobile_sha256: 'other2',
 };
 
 export const normalizeUserIdentityKeys = (
@@ -313,9 +311,7 @@ export const normalizeUserIdentityKeys = (
         if (alias in normalized) {
             const value = normalized[alias];
             delete normalized[alias];
-            if (value !== null) {
-                normalized[SHA256_IDENTITY_ALIASES[alias]] = value;
-            }
+            normalized[SHA256_IDENTITY_ALIASES[alias]] = value;
         }
     }
     return normalized as UserIdentities;
--- a/src/identity-utils.ts
+++ b/src/identity-utils.ts
@@ -317,6 +317,20 @@ export const normalizeUserIdentityKeys = (
     return normalized;
 };
 
+// Compare two identity objects by key and value, independent of
+// object key insertion order. The two sides come from different sources
+// (numeric IdentityType iteration vs. partner-provided / alias-normalized
+// order)
+const identitiesEqual = (
+    a?: UserIdentities,
+    b?: UserIdentities
+): boolean => {
+    const aKeys = Object.keys(a ?? {});
+    const bKeys = Object.keys(b ?? {});
+    if (aKeys.length !== bKeys.length) return false;
+    return aKeys.every((key) => a[key] === b[key]);
+};
+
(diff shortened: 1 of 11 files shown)


rmi22186:
Updated to include mobilesha --> Other2 mapping

An identify request of:
```
const identifyRequest = {
    userIdentities: { 
        email_sha256: 'emailsha',
        mobile_sha256: 'mobilenumber',
    } 
}
```
now maps to

image

src/identity-utils.ts
```diff
@@ -7,6 +7,13 @@ import {
     UserIdentities,
     IdentityCallback,
 } from '@mparticle/web-sdk';
+
+declare module '@mparticle/web-sdk' {
+    interface UserIdentities {
```
alexs-mparticle: Shouldn't we move this into [public types](https://github.com/mParticle/mparticle-web-sdk/blob/master/src/public-types.ts)?

src/identity-utils.ts
```diff
@@ -7,6 +7,13 @@ import {
...
+        email_sha256?: string | undefined;
```
alexs-mparticle: I agree with the robot. This is redundant.

src/identity-utils.ts
```diff
@@ -282,6 +289,27 @@ const getExpireTimestamp = (maxAge: number = ONE_DAY_IN_SECONDS): number =>
 const parseIdentityResponse = (responseText: string): IdentityResultBody =>
     responseText ? JSON.parse(responseText) : ({} as IdentityResultBody);
 
+const SHA256_IDENTITY_ALIASES: Readonly<Record<string, string>> = {
```
alexs-mparticle: I think if you set this to `UserIdentities` you won't have to keep casting later.

```suggestion
const SHA256_IDENTITY_ALIASES: Readonly<UserIdentities> = {
```

src/identity-utils.ts
```diff
@@ -282,6 +289,27 @@ const getExpireTimestamp = (maxAge: number = ONE_DAY_IN_SECONDS): number =>
...
+    email_sha256: 'other',
+    mobile_sha256: 'other2',
+};
+
+export const normalizeUserIdentityKeys = (
+    userIdentities: UserIdentities
+): UserIdentities => {
+    const normalized: Record<string, string | null | undefined> = {
+        ...(userIdentities as Record<string, string | null | undefined>),
+    };
```
alexs-mparticle:
```suggestion
    const normalized: Record<UserIdentities> = {
        ...(userIdentities),
    };
```

src/identity-utils.ts
```diff
@@ -282,6 +289,27 @@ const getExpireTimestamp = (maxAge: number = ONE_DAY_IN_SECONDS): number =>
...
+    for (const alias of Object.keys(SHA256_IDENTITY_ALIASES)) {
+        if (alias in normalized) {
+            const value = normalized[alias];
+            delete normalized[alias];
+            normalized[SHA256_IDENTITY_ALIASES[alias]] = value;
+        }
+    }
+    return normalized as UserIdentities;
```
alexs-mparticle:
```suggestion
    return normalized;
```

alexs-mparticle (changes requested):

src/identity-utils.ts
```diff
@@ -282,6 +289,31 @@ const getExpireTimestamp = (maxAge: number = ONE_DAY_IN_SECONDS): number =>
...
+            const value = normalized[alias];
+            delete normalized[alias];
+            if (value !== null) {
+                normalized[SHA256_IDENTITY_ALIASES[alias]] = value;
+            }
+        }
+    }
+    return normalized as UserIdentities;
+};
+
```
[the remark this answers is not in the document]
rmi22186: Fixed in 1c97df20. `hasIdentityRequestChanged` now normalizes the new request before the stringify-compare; added a regression test in `test/src/tests-identity-utils.ts` covering the alias→canonical round-trip.

src/identity-utils.ts
rmi22186: Moved to `src/public-types.ts` in 1c97df20. That is also the file `package.json` `"types"` points at, so the augmentation reaches consumers via declaration merging on `@mparticle/web-sdk`.

src/identity-utils.ts
rmi22186: Done in 1c97df20. The augmentation now uses `?: string | null` to match the existing `UserIdentities` style on DT master. Also opened DefinitelyTyped/DefinitelyTyped#74955 to add the fields canonically — once that publishes, the augmentation here can go.

src/identity-utils.ts
rmi22186: Taken in 1c97df20. All three downstream casts in `normalizeUserIdentityKeys` dropped as a result — thanks for the nudge.

src/identity-utils.ts
rmi22186: Took the intent in 1c97df20. Used `UserIdentities` directly — `Record<UserIdentities>` does not compile (Record requires two generic args, e.g. `Record<K, V>`). Same outcome: cast and the verbose `Record<string, string | null | undefined>` typing both gone.

src/identity-utils.ts
rmi22186: Done in 1c97df20.

src/identity.js
```diff
@@ -34,11 +35,21 @@ export default function Identity(mpInstance) {
             );
 
             // First, remove any falsy identity values and warn about them
-            const cleanedIdentityApiData = mpInstance._Helpers.Validators.removeFalsyIdentityValues(
+            const removedFalsyIdentityData = mpInstance._Helpers.Validators.removeFalsyIdentityValues(
                 identityApiData,
                 mpInstance.Logger
             );
 
+            // Normalize convenience aliases (email_sha256 → other, mobile_sha256 → other)
```
rmi22186:
```suggestion
            // Normalize convenience aliases (email_sha256 → other, mobile_sha256 → other2)
```

src/sdkRuntimeModels.ts
```diff
@@ -306,7 +306,7 @@ export const LogLevelType = {
 // Currently, this extends MPConfiguration in @types/mparticle__web-sdk
 // and the two will be merged in once the Store module is refactored
 export interface SDKInitConfig
-    extends Omit<MPConfiguration, 'dataPlan' | 'logLevel'> {
+    extends Omit<MPConfiguration, 'dataPlan' | 'logLevel' | 'identityCallback'> {
```
rmi22186, line 309: This was required due to build issues with definitelytyped - the local SDK is actually typed more strongly and is more accurate than DT's definition. Once we deprecated DT, this will be resolved.

src/identity-utils.ts
```diff
@@ -282,6 +282,25 @@ const getExpireTimestamp = (maxAge: number = ONE_DAY_IN_SECONDS): number =>
 const parseIdentityResponse = (responseText: string): IdentityResultBody =>
     responseText ? JSON.parse(responseText) : ({} as IdentityResultBody);
 
+const SHA256_IDENTITY_ALIASES: Readonly<UserIdentities> = {
+    email_sha256: 'other',
+    mobile_sha256: 'other2',
+};
```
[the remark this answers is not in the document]
alexs-mparticle: Nit: @rmi22186 I think we have a `keyOf` utility type that can help with this.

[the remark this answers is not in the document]
rmi22186: @alexs-mparticle updated in 61c828be. Switched the alias map from `Readonly<UserIdentities>` to `Readonly<Partial<Record<keyof UserIdentities, keyof UserIdentities>>>` so both keys and values must be valid identity key names — `null` / `undefined` / arbitrary strings now fail type-check at the declaration site, closing the silent `normalized[null] = value` corruption path Bugbot flagged.

src/identity-utils.ts
```diff
@@ -293,7 +314,9 @@ export const hasIdentityRequestChanged = (
     const currentUserIdentities =
         currentUser.getUserIdentities().userIdentities;
 
-    const newIdentities = newIdentityRequest.userIdentities;
+    const newIdentities = normalizeUserIdentityKeys(
+        newIdentityRequest.userIdentities
+    );
```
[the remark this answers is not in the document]
rmi22186: fixed

alexs-mparticle (approved):

PR #1254, opened 2026-04-27
feat: Expose onShoppableAdsReady on RoktManager

zach-will:
## Background
As part of embedding the Rokt Thank You Element (TYE) into the Joint SDK, we need a way for partners to register callbacks and listen to TYE events.

Previously, partners were instructed to attach event listeners inside the TYE onLoad handler:

```
script.src = baseUrl + '/rokt-thank-you-element.js'
script.onLoad = () => {
  window.RoktThankYouElement.on('THANK_YOU_ELEMENT_LOADING_INITIATED').subscribe(() => {
    // Do something
  });
}
```

With TYE now internalized within the Joint SDK, this approach is no longer viable.

We also want to allow partners to register event listeners before mParticle or RoktKit is fully ready. To support this, we’ve introduced a new method, `onShoppableAdsReady`, which exposes a safe hook for registering TYE event listeners.

The updated integration pattern is:

```
  window.mParticle.Rokt.onShoppableAdsReady(() => {
    window.RoktThankYouElement.on('THANK_YOU_ELEMENT_LOADING_INITIATED').subscribe(() => {
      // Do something
    });
  });
  window.mParticle?.ready?.(() => {
```

## Changes to snippet js.
Integrations will now require `onShoppableAdsReady` to be included in the RoktMethods stub:

```
[
      'selectPlacements',
      'hashAttributes',
      'setExtensionData',
      'use',
      'onShoppableAdsReady',
    ].forEach(
    (name) => {
      window.mParticle.Rokt[name] = (...args: any[]) =>
        window.mParticle.config.rq.push([`Rokt.${name}`, ...args])
    },
  )
```

## What Changed
- Added a new public method: onShoppableAdsReady in RoktManager
- When ThankYouElement is included in the kit config:
  - RoktKit is initialized early via RoktManager, but is not marked as ready until the launcher is defined
  - Just before loading rokt-thank-you-element.js, the onShoppableAdsReady callback queue in RoktManager is flushed
    - This ensures all queued callbacks are registered as onLoad handlers
    - Any subsequent calls to onShoppableAdsReady are executed immediately (no longer deferred)

## Deployment order
- Deploy this PR
- Deploy changes to RoktKit: [PR-91](https://github.com/mparticle-integrations/mparticle-javascript-integration-rokt/pull/91)

## Media

https://github.com/user-attachments/assets/62203695-3c4b-4918-8abb-9a737372854c

### DIFF
--- a/src/roktManager.ts
+++ b/src/roktManager.ts
@@ -90,6 +90,8 @@ export interface IRoktOptions {
 
 export type IRoktLauncherOptions = Dictionary<any>;
 
+const ON_SHOPPABLE_ADS_READY_METHOD = 'onShoppableAdsReady'
+
 // The purpose of this class is to create a link between the Core mParticle SDK and the
 // Rokt Web SDK via a Web Kit.
 // The Rokt Manager should load before the Web Kit and stubs out many of the
@@ -470,7 +472,7 @@ export default class RoktManager {
         this.kit = kit;
 
         this.messageQueue.forEach((message, key) => {
-            if (message.methodName === 'onShoppableAdsReady') {
+            if (message.methodName === ON_SHOPPABLE_ADS_READY_METHOD) {
                 try {
                     kit.onShoppableAdsReady(message.payload as () => void);
                 } catch (e) {
--- a/src/roktManager.ts
+++ b/src/roktManager.ts
@@ -76,6 +76,7 @@ export interface IRoktKit {
     selectPlacements: (options: IRoktSelectPlacementsOptions) => Promise<IRoktSelection>;
     setExtensionData<T>(extensionData: IRoktPartnerExtensionData<T>): void;
     use: <T>(name: string) => Promise<T>;
+    onShoppableAdsReady(callback: () => void): void;
     launcherOptions?: Dictionary<any>;
     settings?: IRoktKitSettings;
     integrationName?: string;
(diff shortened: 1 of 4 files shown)


src/roktManager.ts
```diff
@@ -450,6 +452,37 @@ export default class RoktManager {
...
+            const errorMessage = error instanceof Error ? error.message : String(error);
+            this.logger.error(`Failed to register onShoppableAdsReady callback: ${errorMessage}`);
+        }
+    }
+
+    public flushOnShoppableAdsReadyMessageQueue(kit: IRoktKit): void {
+        this.kit = kit;
+
+        this.messageQueue.forEach((message, key) => {
+            if (message.methodName === 'onShoppableAdsReady') {
```
alexs-mparticle: Let's make this a constant

alexs-mparticle (changes requested):

src/roktManager.ts
zach-will: Updated.

src/roktManager.ts
```diff
@@ -450,6 +454,37 @@ export default class RoktManager {
...
+        try {
+            this.kit.onShoppableAdsReady(callback);
+        } catch (error) {
+            const errorMessage = error instanceof Error ? error.message : String(error);
+            this.logger.error(`Failed to register onShoppableAdsReady callback: ${errorMessage}`);
+        }
+    }
+
+    public flushOnShoppableAdsReadyMessageQueue(kit: IRoktKit): void {
+        this.kit = kit;
```
alexs-mparticle, line 472: Noticed you're ressigning `this.kit` here. Is this necessary?

zach-will: Assigning the kit here lets `onShoppableAdsReady` defer to it in cases where a partner registers their callback after the TYE script has loaded but before `RoktLauncher` is defined.

alexs-mparticle (approved):

PR #1255, opened 2026-04-28
feat: Add Identity.search for IDSync Search

rmi22186:
## Summary

- New public API `mParticle.Identity.search(workspaceApiKey, knownIdentities, callback)` POSTs to mParticle's IDSync `/v1/search` and invokes the callback with `{ httpCode, body? }`. Sits alongside `identify`/`login`/`logout`/`modify`/`aliasUsers` on the Identity surface. The caller supplies a workspace-scoped API key (sent as `x-mp-key`), held separately from the host SDK's `_Store.devToken`, so a single install can search against an arbitrary workspace without being reconfigured.
- Outcomes: 200 → `{ httpCode: 200, body }`; 404 → `{ httpCode: 404 }`; missing key or invalid email → `noHttpCoverage` (no network); SDK opted out → `loggingDisabledOrMissingAPIKey`; network or JSON-parse failure → `noHttpCoverage` plus a structured `ISDKError` through `_ErrorReportingDispatcher`, mirroring `identifyRequest` in `identityApiClient.ts`.
- Opt-out parity: gates on `mpInstance._Helpers.canLog()` before any network call, so `setOptOut(true)` short-circuits the request and no email is transmitted — matching `identify`/`login`/`logout`/`modify`/`aliasUsers`.
- Pairs with the Rokt Web Kit PR — https://github.com/mparticle-integrations/mparticle-javascript-integration-rokt/pull/92 — and depends on the Fastly edge allowlist for `/v1/search` being updated to include `x-mp-key` (curl already round-trips 200/404; the browser preflight is the only remaining blocker).
- Test plan: `npm run lint` + `npm run build` clean; Karma on ChromeHeadless 1077 passing covering 200/404, validation, opt-out, and error-reporter paths in `tests-search.ts`. Live browser smoke test still blocked on the Fastly allowlist update.


src/identity.js
```diff
@@ -730,6 +731,72 @@ export default function Identity(mpInstance) {
...
+                knownIdentities,
+                apiKey,
+                requestBuilder,
+                searchUrl,
+                callback,
+                mpInstance.Logger,
+                undefined,
+                mpInstance._ErrorReportingDispatcher
+            );
+        },
```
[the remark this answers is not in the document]
rmi22186: fixed

rmi22186:
Note: Check to see if onIdentifyComplete is ever called when identify fails.

src/identity.interfaces.ts
```diff
@@ -12,6 +12,10 @@ import {
     mParticleUserCart,
     IIdentityResponse,
 } from './identity-user-interfaces';
+import {
+    ISearchKnownIdentities,
+    SearchCallback,
```
alexs-mparticle: Nit:

```suggestion
    IIdentitySearchKnownIdentities,
    IdentitySearchCallback,
```

src/identity.js
```diff
@@ -730,6 +731,83 @@ export default function Identity(mpInstance) {
...
+         * @param {String} workspaceApiKey Workspace API key (sent as x-mp-key).
+         * @param {Object} knownIdentities `{ email: string }`
+         * @param {Function} callback Invoked with the `ISearchResult`.
+         */
+        search: function(workspaceApiKey, knownIdentities, callback) {
+            if (!mpInstance._Helpers.canLog()) {
+                mpInstance.Logger.verbose(
+                    Messages.InformationMessages.AbandonLogEvent
+                );
+                if (mpInstance._Helpers.Validators.isFunction(callback)) {
```
alexs-mparticle: This can be pulled from `utils.ts` instead of `Validators`.

```suggestion
                if (isFunction(callback)) {
```

src/identity.js
```diff
@@ -730,6 +731,83 @@ export default function Identity(mpInstance) {
...
+                    try {
+                        callback({
+                            httpCode: HTTPCodes.loggingDisabledOrMissingAPIKey,
+                        });
+                    } catch (e) {
+                        mpInstance.Logger.error(
+                            'Error invoking search callback: ' +
+                                ((e && e.message) || String(e))
```
alexs-mparticle: I like the robot's suggestion below.

src/identity.js
```diff
@@ -730,6 +731,83 @@ export default function Identity(mpInstance) {
...
+            // service can correlate requests across endpoints.
+            const requestBuilder = function() {
+                return {
+                    client_sdk: {
+                        platform: Constants.platform,
+                        sdk_vendor: Constants.sdkVendor,
+                        sdk_version: Constants.sdkVersion,
+                    },
+                    environment: environment,
+                    request_id: mpInstance._Helpers.generateUniqueId(),
```
alexs-mparticle: Use `utils.ts`.

```suggestion
                    request_id: generateUniqueId(),
```

src/identity.js
```diff
@@ -730,6 +731,83 @@ export default function Identity(mpInstance) {
...
+            const searchUrl = serviceUrl + 'search?cb=1';
+
+            const environment = mpInstance._Store.SDKConfig.isDevelopmentMode
+                ? 'development'
+                : 'production';
+
+            // Build the same envelope that /v1/identify uses (client_sdk,
+            // request_id, request_timestamp_ms, environment) so the IDSync
+            // service can correlate requests across endpoints.
+            const requestBuilder = function() {
```
alexs-mparticle: Nit: can we extract this method outside of the function to make it more readable?

src/identity.js
```diff
@@ -730,6 +731,83 @@ export default function Identity(mpInstance) {
...
+
+            // The Search endpoint is colocated with /v1/identify under
+            // identityUrl, so we reuse the same service URL builder. We do
+            // NOT append the apiKey to the URL — auth is done via x-mp-key.
+            const serviceUrl = mpInstance._Helpers.createServiceUrl(
+                mpInstance._Store.SDKConfig.identityUrl
+            );
+            const searchUrl = serviceUrl + 'search?cb=1';
+
+            const environment = mpInstance._Store.SDKConfig.isDevelopmentMode
```
alexs-mparticle: I think you can destructure a lot of these configvalues to make the code more readable.

src/search.ts
```diff
@@ -0,0 +1,234 @@
+import Constants, { HTTP_OK, HTTP_NOT_FOUND } from './constants';
```
alexs-mparticle: I think we should put this in `src/identity/search.ts` so that we can eventually split out the other identity functions into their own space for cleaner organization.

alexs-mparticle (changes requested):

src/identity/search.ts
```diff
@@ -0,0 +1,234 @@
...
+    requestBuilder: () => Omit<IIdentitySearchRequestBody, 'known_identities'>,
+    searchUrl: string,
+    callback: IdentitySearchCallback,
+    logger: SDKLoggerApi,
+    uploader?: AsyncUploader,
+    errorReporter?: IErrorReportingService,
+): Promise<void> => {
+    // Validate the callback up front. If it isn't a function we have nowhere
+    // to deliver a result to, so log and bail out without invoking anything.
+    if (typeof callback !== 'function') {
```
alexs-mparticle: utils.ts.

```suggestion
    if (!isFunction(callback)) {
```

src/identity/search.ts
```diff
@@ -0,0 +1,234 @@
...
+            logger.error(
+                'Error invoking search callback: ' +
+                    ((e as Error)?.message || String(e)),
+            );
+        }
+    };
+
+    // No valid email -> deliver httpCode: noHttpCoverage so callers waiting on
+    // the callback (e.g. to clear a loading state) don't hang.
+    if (!knownIdentities || typeof knownIdentities.email !== 'string' || !knownIdentities.email) {
```
alexs-mparticle, line 132: Nit: We should set up an email validation util function.

src/identity/search.ts
```diff
@@ -0,0 +1,234 @@
...
+                : new XHRUploader(searchUrl));
+
+        logger.verbose('Sending search request to ' + searchUrl);
+        const response: Response = await api.upload(fetchPayload, searchUrl);
+
+        let body: IIdentitySearchResponseBody | undefined;
+
+        // FetchUploader returns a real Response with .json(); XHRUploader
+        // returns an XHR-shaped object with `responseText`. We tolerate both.
+        if (typeof (response as Response).json === 'function') {
```
alexs-mparticle: `isFunction`

src/identity/search.ts
```diff
@@ -0,0 +1,234 @@
...
+            try {
+                body = (await (response as Response).json()) as IIdentitySearchResponseBody;
+            } catch (e) {
+                logger.verbose(
+                    'search response had no parseable JSON body.',
+                );
+            }
+        } else {
+            const xhrLike = (response as unknown) as XMLHttpRequest;
```
alexs-mparticle, line 195: Do we still need to support XHR any longer?

If we do, why isn't this check inside the `XHRUploader` and `FetchUploader`, rather than in this function?

src/identity/search.ts
```diff
@@ -0,0 +1,234 @@
...
+        safeInvoke({ httpCode: response.status, body });
+    } catch (e) {
+        const message = (e as Error)?.message || String(e);
+        const reportMessage = 'Error sending search request: ' + message;
+        logger.error(reportMessage);
+        // Mirror the identity-route pattern in identityApiClient.ts: log to
+        // console AND push a structured report through the dispatcher so any
+        // registered IErrorReportingService (e.g. the Rokt kit's) can observe
+        // the failure.
+        errorReporter?.report({
```
alexs-mparticle, line 228: I believe we decided that Error reporting such as this would be offloaded to the kit?

src/identity.js
```diff
@@ -730,6 +731,32 @@ export default function Identity(mpInstance) {
...
+         * @param {Function} callback Invoked with the `IIdentitySearchResult`.
+         */
+        search: function(workspaceApiKey, knownIdentities, callback) {
+            executeSearchRequest(
+                mpInstance,
+                workspaceApiKey,
+                knownIdentities,
+                callback
+            );
+        },
```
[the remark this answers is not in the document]
alexs-mparticle: @rmi22186 Is this a concern?

alexs-mparticle (changes requested):

src/identity/search.ts
rmi22186: I could go either way - the backend accepts any string, so I think that's why there never was validation on the SDK side.  Really its up to the client site to do the validation.

src/identity/search.ts
rmi22186: This matches what we have in our [identityApiClient](https://github.com/mParticle/mparticle-web-sdk/blob/master/src/identityApiClient.ts#L344-L348) for the other identity methods.

src/identity/search.ts
rmi22186: Also mimicing what we have in identityapiclient.ts.  This sounds like a larger refactor.  Also, we've flagged this as part of https://go.mparticle.com/work/SQDSDKS-6568 🥇

src/identity.js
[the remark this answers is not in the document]
rmi22186: Shouldn't be initially beacuse the kit is the only one that will call this right now, and will be loaded only after the SDK is, so `Identity.search` will always be available available.  However, for future customers who may call this directly, so i've added this.

src/identity/search.ts
```diff
@@ -0,0 +1,235 @@
...
+                : new XHRUploader(searchUrl));
+
+        logger.verbose('Sending search request to ' + searchUrl);
+        const response: Response = await api.upload(fetchPayload, searchUrl);
+
+        let body: IIdentitySearchResponseBody | undefined;
+
+        // FetchUploader returns a real Response with .json(); XHRUploader
+        // returns an XHR-shaped object with `responseText`. We tolerate both.
+        if (isFunction((response as Response).json)) {
```
alexs-mparticle: I think the robot is right. You can probably remove `as Response`.

src/identity/search.ts
```diff
@@ -0,0 +1,235 @@
...
+            try {
+                body = (await (response as Response).json()) as IIdentitySearchResponseBody;
+            } catch (e) {
```
alexs-mparticle, line 189: Nit: I don't think this needs to be a try/catch just to log a verbose warning. You probably can just check if response is empty, right?

src/identity-utils.ts
```diff
@@ -321,3 +329,76 @@ export const hasExplicitIdentifier = (store: IStore | undefined | null): boolean
...
+        return;
+    }
+
+    // The Search endpoint is colocated with /v1/identify under
+    // identityUrl, so we reuse the same service URL builder. We do
+    // NOT append the apiKey to the URL — auth is done via x-mp-key.
+    const serviceUrl: string = _Helpers.createServiceUrl(identityUrl);
+    const searchUrl: string = serviceUrl + 'search?cb=1';
+
+    const environment: 'development' | 'production' = isDevelopmentMode
```
alexs-mparticle:
```suggestion
    const environment: Environment = isDevelopmentMode
```

alexs-mparticle (changes requested):

src/identity/search.ts
rmi22186: The JSON has to be attempted. There's no way to look at a response and know up-front if it's valid JSON without actually trying to parse it. "Empty" and "not valid JSON" are two different failure modes. 

We could check Content-Length: 0 or response.body === null to detect the empty case, but that wouldn't help us if Fastly returns an HTML error page on a 5xx,  the body's not empty, it's just not JSON, and response.json() would still throw. Same for malformed JSONor a truncated response

alexs-mparticle (approved):

src/identity/search.ts
```diff
@@ -0,0 +1,235 @@
...
+/**
+ * Shape of `known_identities` accepted by `search`.
+ *
+ * The IDSync `/v1/search` endpoint accepts the same identity keys as
+ * `/v1/identify`, but for v1 of this client API we only support `email`.
+ * Additional identity types can be added here in the future without breaking
+ * existing consumers.
+ */
+export interface IIdentitySearchKnownIdentities {
+    email: string;
```
samdozor, line 27: why restrict this? if we're adding this as a public API we should have it be comprehensive. 

Also - I think several clients are only passing us hash email (eg other) so this will not work for them

samdozor (changes requested):
Overall comment - why create a new search.ts at all and not just fold this directly into how the rest of the identity API is implemented? the support identity inconsistency is an example of something that we fix for free if we do that - also we avoid a larger code surface to be adding to the SDK.

were you thinking this limits blast radius of something goes wrong?

rmi22186:
Discussed over DM - will merge this and fast follow with any cleanup.

PR #1256, opened 2026-05-05
fix: Allow all user identity types in Identity.search

rmi22186:
## Summary

Expands `Identity.search` so it forwards every identifier the partner has, not just `email`. Builds on #1255 (the initial IDSync Search client API), which restricted `known_identities` to `{ email }`.

In production, partners sometimes pass hashed email through `other`, `other2-10`, or `customerid` rather than the plain `email` field. The previous shape silently dropped those, so search returned 404 even when a match existed under a non-email key.

## What changed

**`src/identity/search.ts`**
- `sendSearchRequest` now accepts `UserIdentities` (the SDK's full identity surface) instead of the bespoke `{ email: string }` type.
- The "no usable identifier" guard now checks that *any* property has a non-empty string value (was: required `email` specifically).
- Forwards `{ ...knownIdentities }` to the wire instead of `{ email: knownIdentities.email }`.
- Removes the now-unused `IIdentitySearchKnownIdentities` interface in favor of `UserIdentities`.

**Public types (`src/public-types.ts`, `src/identity.interfaces.ts`)**
- Drops `IIdentitySearchKnownIdentities` from the public surface (it was never load-bearing for kits — only one prior consumer, this PR removes it).
- Adds `IUserIdentities` for consumption by Kit
- `SDKIdentityApi.search` and `executeSearchRequest` updated to use `UserIdentities`.

**Tests (`test/src/tests-search.ts`)**
- Renames the existing guard test to reflect the new "any non-empty string identifier" contract; expands its inputs to include `{ email: 12345, other: null }` (mixed wrong-typed + null).
- New: `forwards non-email identifiers (e.g. hashed email in 'other') to the wire` — asserts that `{ other: 'sha256:abc123', customerid: 'cust-1' }` produces a 200 round-trip with the full identifier set in the request body.

## Breaking changes

`IIdentitySearchKnownIdentities` is removed from the public exports. It was added in #1255 and (per \`git grep\`) had no external consumers, so the practical blast radius is zero — but worth flagging for changelog purposes.

## Test plan

- [x] \`npm run test\` — Mocha suite, including the renamed/extended guard test and the new \`forwards non-email identifiers\` test
- [x] \`npm run build\` — \`tsc -p tsconfig.types.json\` should emit \`IUserIdentities\` into \`dist/types/src/public-types.d.ts\`
- [x] Verify a kit consumer (Rokt) can \`import type { IUserIdentities } from '@mparticle/web-sdk'\` and call \`mParticle.Identity.search(apiKey, { other: '<sha256>' }, cb)\` against a real workspace with a non-email identifier

pair with https://github.com/mparticle-integrations/mparticle-javascript-integration-rokt/pull/93

### DIFF
--- a/src/identity.interfaces.ts
+++ b/src/identity.interfaces.ts
@@ -12,10 +12,7 @@ import {
     mParticleUserCart,
     IIdentityResponse,
 } from './identity-user-interfaces';
-import {
-    IIdentitySearchKnownIdentities,
-    IdentitySearchCallback,
-} from './identity/search';
+import { IdentitySearchCallback } from './identity/search';
 const { platform, sdkVendor, sdkVersion, HTTPCodes } = Constants;
 
 export type IdentityPreProcessResult = {
@@ -173,18 +170,40 @@ export interface SDKIdentityApi {
      */
     search?(
         workspaceApiKey: string,
-        knownIdentities: IIdentitySearchKnownIdentities,
+        knownIdentities: UserIdentities,
         callback: IdentitySearchCallback
     ): void;
 }
 
 export type {
-    IIdentitySearchKnownIdentities,
     IIdentitySearchResult,
     IIdentitySearchResponseBody,
     IdentitySearchCallback,
 } from './identity/search';
 
+export interface IUserIdentities {
+    customerid?: string | undefined;
+    email?: string | undefined;
+    other?: string | undefined;
+    other2?: string | undefined;
+    other3?: string | undefined;
+    other4?: string | undefined;
+    other5?: string | undefined;
+    other6?: string | undefined;
+    other7?: string | undefined;
+    other8?: string | undefined;
+    other9?: string | undefined;
+    other10?: string | undefined;
+    mobile_number?: string | undefined;
+    phone_number_2?: string | undefined;
+    phone_number_3?: string | undefined;
+    facebook?: string | undefined;
+    facebookcustomaudienceid?: string | undefined;
+    google?: string | undefined;
+    twitter?: string | undefined;
+    microsoft?: string | undefined;
+    yahoo?: string | undefined;
+}
 export interface IIdentity {
     audienceManager: AudienceManager;
     idCache: BaseVault<Dictionary<ICachedIdentityCall>>;
--- a/src/identity.interfaces.ts
+++ b/src/identity.interfaces.ts
@@ -182,27 +182,27 @@ export type {
 } from './identity/search';
 
 export interface IUserIdentities {
-    customerid?: string | undefined;
-    email?: string | undefined;
-    other?: string | undefined;
-    other2?: string | undefined;
-    other3?: string | undefined;
-    other4?: string | undefined;
-    other5?: string | undefined;
-    other6?: string | undefined;
-    other7?: string | undefined;
-    other8?: string | undefined;
-    other9?: string | undefined;
-    other10?: string | undefined;
-    mobile_number?: string | undefined;
-    phone_number_2?: string | undefined;
-    phone_number_3?: string | undefined;
-    facebook?: string | undefined;
-    facebookcustomaudienceid?: string | undefined;
-    google?: string | undefined;
-    twitter?: string | undefined;
-    microsoft?: string | undefined;
-    yahoo?: string | undefined;
+    customerid?: string;
+    email?: string;
+    other?: string;
+    other2?: string;
+    other3?: string;
+    other4?: string;
+    other5?: string;
+    other6?: string;
+    other7?: string;
+    other8?: string;
+    other9?: string;
+    other10?: string;
+    mobile_number?: string;
+    phone_number_2?: string;
+    phone_number_3?: string;
+    facebook?: string;
+    facebookcustomaudienceid?: string;
+    google?: string;
+    twitter?: string;
+    microsoft?: string;
+    yahoo?: string;
 }
 export interface IIdentity {
     audienceManager: AudienceManager;
--- a/src/identity/search.ts
+++ b/src/identity/search.ts
@@ -131,7 +131,7 @@ export const sendSearchRequest = async (
         )
     ) {
         logger.verbose(
-            'search called without any non-empty identifier; skipping request.',
+            'Identity search called with non empty identifiers; skipping request.',
         );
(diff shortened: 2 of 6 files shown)


src/identity/search.ts
```diff
@@ -127,11 +116,17 @@ export const sendSearchRequest = async (
...
-    if (!knownIdentities || typeof knownIdentities.email !== 'string' || !knownIdentities.email) {
+    // No usable identifier -> deliver httpCode: noHttpCoverage so callers
+    // waiting on the callback (e.g. to clear a loading state) don't hang.
+    const hasIdentifier =
+        knownIdentities &&
+        typeof knownIdentities === 'object' &&
+        Object.values(knownIdentities).some(
+            (v) => typeof v === 'string' && v.length > 0,
+        );
+    if (!hasIdentifier) {
```
alexs-mparticle: Making this a convenience method right before invocation doesn't really help with readability, IMO.

We probably have some utility methods that could help clean this up a bit instead, or we can roll this into a new utility method.

```suggestion
    if (!isEmpty(knownIdentities) && knownIdentities.some(v => !isEmpty(v)) {
```

src/identity/search.ts
```diff
@@ -155,9 +150,7 @@ export const sendSearchRequest = async (
         const requestEnvelope = requestBuilder();
         const requestBody: IIdentitySearchRequestBody = {
             ...requestEnvelope,
-            known_identities: {
-                email: knownIdentities.email,
-            },
+            known_identities: { ...knownIdentities },
```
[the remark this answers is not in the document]
alexs-mparticle: I think the server does a decent job of ignoring null values but we may want to clean this up before spreading just in case.

src/identity.interfaces.ts
```diff
@@ -173,18 +170,40 @@ export interface SDKIdentityApi {
...
 
 export type {
-    IIdentitySearchKnownIdentities,
     IIdentitySearchResult,
     IIdentitySearchResponseBody,
     IdentitySearchCallback,
 } from './identity/search';
 
+export interface IUserIdentities {
+    customerid?: string | undefined;
```
alexs-mparticle: I agree with the robot, optional and undefined is redundant

alexs-mparticle (changes requested):

src/identity.interfaces.ts
rmi22186: done

src/identity/search.ts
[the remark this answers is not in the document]
rmi22186: updated

src/identity/search.ts
rmi22186: fixed.  I had the same idea and the robot was telling me isEmpty would let things like {email: 1234} through. However, we do validation prior to this so it ended up being a non issue.

src/identity/search.ts
```diff
@@ -127,11 +117,21 @@ export const sendSearchRequest = async (
...
+    ).userIdentities;
+
+    // No usable identifier -> deliver httpCode: noHttpCoverage so callers
+    // waiting on the callback (e.g. to clear a loading state) don't hang.
+    if (
+        isEmpty(cleanedKnownIdentities) ||
+        !Object.values(cleanedKnownIdentities).some(
+            (v) => typeof v === 'string' && v.length > 0,
+        )
+    ) {
```
alexs-mparticle, line 131: Now that I look at this again, we can remove the need for `isEmpty`.

```suggestion
    if (
        !Object.values(cleanedKnownIdentities ?? {}).some(
            (v) => typeof v === 'string' && v.length > 0,
        )
    ) {
```

We probably can even make a utility function to reduce `(v) => typeof v === 'string' && v.length > 0,` into `isEmptyString()` but that might be unnecessary.

src/identity/search.ts
```diff
@@ -127,11 +117,21 @@ export const sendSearchRequest = async (
...
         logger.verbose(
-            'search called without a valid email; skipping request.',
+            'search called without any non-empty identifier; skipping request.',
```
alexs-mparticle: Nit: The verbose log reads like a double negative (or maybe I'm just getting tripped up by the phrasing)

```suggestion
            'Identity search called with non empty identifiers; skipping request.',
```

alexs-mparticle (approved):

src/identity/search.ts
rmi22186: I think you're right about the isEmpty, but forgetting the !.  I'll add it

src/identity/search.ts
```diff
@@ -153,11 +152,10 @@ export const sendSearchRequest = async (
     // rejecting and the caller hanging on a never-fired callback.
     try {
         const requestEnvelope = requestBuilder();
+
         const requestBody: IIdentitySearchRequestBody = {
             ...requestEnvelope,
-            known_identities: {
-                email: knownIdentities.email,
-            },
+            known_identities: { ...cleanedKnownIdentities },
```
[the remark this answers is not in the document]
rmi22186: Currently only the kit is consuming this.  Will fast follow when we do a broader refactor.

src/identity/search.ts
```diff
@@ -1,6 +1,6 @@
 import Constants, { HTTP_OK, HTTP_NOT_FOUND } from '../constants';
 import { SDKLoggerApi } from '../sdkRuntimeModels';
-import { Environment, isFunction } from '../utils';
+import { Environment, isEmpty, isFunction } from '../utils';
```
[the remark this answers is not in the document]
rmi22186: removed

src/identity/search.ts
```diff
@@ -127,11 +117,20 @@ export const sendSearchRequest = async (
...
+    // No usable identifier -> deliver httpCode: noHttpCoverage so callers
+    // waiting on the callback (e.g. to clear a loading state) don't hang.
+    if (
+        !Object.values(cleanedKnownIdentities ?? {}).some(
+            (v) => typeof v === 'string' && v.length > 0,
+        )
+    ) {
         logger.verbose(
-            'search called without a valid email; skipping request.',
+            'Identity search called with non empty identifiers; skipping request.',
```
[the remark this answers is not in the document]
rmi22186: updated.

PR #1258, opened 2026-05-12
refactor: convert events module to TS

jaissica12:
## Background

As part of the ongoing effort to migrate the mParticle Web SDK from JavaScript to TypeScript, this PR converts the events module and its corresponding test file to TypeScript for improved type safety and developer experience.

## What Has Changed

- Converted `src/events.js` to `src/events.ts` with full type annotations on all method parameters, return types, and internal variables
- Converted `test/src/tests-event-logging.js` to `test/src/tests-event-logging.ts` with proper global type declarations and type casts for intentionally invalid test inputs
- Updated `src/events.interfaces.ts`:
  - Added `TrackingCallback` type for geolocation tracking
  - Expanded `logImpressionEvent` signature to accept `SDKImpression | SDKImpression[] | SDKProductImpression | SDKProductImpression[]`
  - Widened `promotion` param to accept `SDKPromotion | SDKPromotion[]`
- Minor supporting type adjustments in:
  - `src/sdkRuntimeModels.ts` — widened `PromotionActionType` to accept typed union
  - `src/serverModel.ts` — added `as string` cast for `PromotionActionType`
  - `src/store.ts` — added `timeout` to `SDKConfig` interface
  - `test/src/config/constants.ts` — type cast for `window.mParticle`

## Screenshots/Video
Screenshot 2026-05-14 at 11 01 50 AM
Screenshot 2026-05-14 at 11 02 21 AM
Screenshot 2026-05-14 at 11 03 03 AM
Screenshot 2026-05-14 at 11 03 45 AM

## Checklist

- Closes https://go/j/SDKE-1106

### DIFF
--- a/src/events.interfaces.ts
+++ b/src/events.interfaces.ts
@@ -8,6 +8,7 @@ import {
     BaseEvent,
     SDKEvent,
     SDKEventCustomFlags,
+    SDKImpression,
     SDKProduct,
     SDKProductImpression,
     SDKPromotion,
@@ -45,11 +46,11 @@ export interface IEvents {
     ): void;
     logEvent(event: BaseEvent, eventOptions?: SDKEventOptions): void;
     logImpressionEvent(
-        impression: SDKProductImpression,
+        impression: SDKImpression | SDKImpression[] | SDKProductImpression | SDKProductImpression[],
         attrs?: SDKEventAttrs,
         customFlags?: SDKEventCustomFlags,
         eventOptions?: SDKEventOptions
-    );
+    ): void;
     logOptOut(): void;
     logProductActionEvent(
         productActionType: valueof<typeof ProductActionType>,
--- a/src/events.interfaces.ts
+++ b/src/events.interfaces.ts
@@ -62,7 +62,7 @@ export interface IEvents {
     ): void;
     logPromotionEvent(
         promotionType: valueof<typeof PromotionActionType>,
-        promotion: SDKPromotion,
+        promotion: SDKPromotion | SDKPromotion[],
         attrs?: SDKEventAttrs,
         customFlags?: SDKEventCustomFlags,
         eventOptions?: SDKEventOptions
--- a/src/events.interfaces.ts
+++ b/src/events.interfaces.ts
@@ -1,5 +1,4 @@
 import {
-    Callback,
     SDKEventAttrs,
     SDKEventOptions,
     TransactionAttributes,
@@ -15,6 +14,12 @@ import {
 } from './sdkRuntimeModels';
 import { valueof } from './utils';
 import { EventType, ProductActionType, PromotionActionType } from './types';
+
+export type TrackingCallback = ((
+    position?: GeolocationPosition | {
+        coords: { latitude: number | string; longitude: number | string };
+    }
+) => void) | null;
 
 // Supports wrapping event handlers functions that will ideally return a specific type
 type EventHandlerFunction<T> = (element: HTMLLinkElement | HTMLFormElement) => T;
@@ -79,6 +84,6 @@ export interface IEvents {
         attrs?: SDKEventAttrs,
         customFlags?: SDKEventCustomFlags
     ): void;
-    startTracking(callback: Callback): void;
+    startTracking(callback: TrackingCallback): void;
     stopTracking(): void;
 }
--- a/src/events.interfaces.ts
+++ b/src/events.interfaces.ts
@@ -51,6 +51,7 @@ export interface IEvents {
     ): void;
     logEvent(event: BaseEvent, eventOptions?: SDKEventOptions): void;
     logImpressionEvent(
+        // https://go/j/SDKE-1199
         impression: SDKImpression | SDKImpression[] | SDKProductImpression | SDKProductImpression[],
         attrs?: SDKEventAttrs,
         customFlags?: SDKEventCustomFlags,
--- a/src/events.ts
+++ b/src/events.ts
@@ -1,11 +1,36 @@
 import Types from './types';
 import Constants from './constants';
+import { IEvents } from './events.interfaces';
+import { IMParticleWebSDKInstance } from './mp-instance';
+import {
+    BaseEvent,
+    SDKEvent,
+    SDKEventCustomFlags,
+    SDKImpression,
+    SDKProduct,
+    SDKProductActionType,
+    SDKProductImpression,
+    SDKPromotion,
+} from './sdkRuntimeModels';
+import { SDKEventAttrs, SDKEventOptions, TransactionAttributes } from '@mparticle/web-sdk';
+import { valueof } from './utils';
+import { EventType, ProductActionType, PromotionActionType } from './types';
+
+interface DOMHandlerElement extends HTMLElement {
+    href?: string;
+    target?: string;
+    submit?: () => void;
+    attachEvent?: (event: string, handler: EventListener) => void;
+}
 
 var Messages = Constants.Messages;
 
-export default function Events(mpInstance) {
+export default function Events(
+    this: IEvents,
+    mpInstance: IMParticleWebSDKInstance
+): void {
     var self = this;
-    this.logEvent = function(event, options) {
+    this.logEvent = function(event: BaseEvent, options?: SDKEventOptions): void {
         mpInstance.Logger.verbose(
             Messages.InformationMessages.StartingLogEvent + ': ' + event.name
         );
@@ -19,7 +44,7 @@ export default function Events(mpInstance) {
         }
     };
 
-    this.startTracking = function(callback) {
+    this.startTracking = function(callback: Function | null): void {
         if (!mpInstance._Store.isTracking) {
             if ('geolocation' in navigator) {
                 mpInstance._Store.watchPositionId = navigator.geolocation.watchPosition(
@@ -37,7 +62,7 @@ export default function Events(mpInstance) {
             triggerCallback(callback, position);
         }
 
-        function successTracking(position) {
+        function successTracking(position: GeolocationPosition): void {
             mpInstance._Store.currentPosition = {
                 lat: position.coords.latitude,
                 lng: position.coords.longitude,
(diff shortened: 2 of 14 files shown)


src/events.ts
```diff
@@ -279,43 +311,48 @@ export default function Events(mpInstance) {
...
+                const imp = item as SDKImpression;
                 event.ProductImpressions.push({
-                    ProductImpressionList: impression.Name,
-                    ProductList: Array.isArray(impression.Product)
-                        ? impression.Product
-                        : [impression.Product],
+                    ProductImpressionList: imp.Name,
+                    ProductList: Array.isArray(imp.Product)
+                        ? imp.Product
+                        : [imp.Product],
```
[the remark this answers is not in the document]
jaissica12: Fixed by adding a 'Name' in item check to distinguish SDKImpression from SDKProductImpression. The new change correctly handles SDKProductImpression input that previously produced undefined values silently.

src/events.ts
```diff
@@ -116,26 +147,26 @@ export default function Events(mpInstance) {
...
 
-            self.logCommerceEvent(event, attrs);
+            this.logCommerceEvent(event, attrs);
         }
     };
 
     this.logProductActionEvent = function(
-        productActionType,
-        product,
-        customAttrs,
```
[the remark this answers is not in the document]
jaissica12: Fixed. Converted all affected methods (logAST, logCheckoutEvent, logProductActionEvent, logPurchaseEvent, logRefundEvent, logPromotionEvent, logImpressionEvent) from function to arrow functions, which lexically capture this from the constructor.

src/events.interfaces.ts
```diff
@@ -45,11 +46,11 @@ export interface IEvents {
     ): void;
     logEvent(event: BaseEvent, eventOptions?: SDKEventOptions): void;
     logImpressionEvent(
-        impression: SDKProductImpression,
+        impression: SDKImpression | SDKImpression[] | SDKProductImpression | SDKProductImpression[],
```
alexs-mparticle, line 55: I brought this up with @rmi22186 earlier. I would like for us to not support both arrays and non-arrays for these endpoints, but that might have to be a follow up project.

src/events.ts
```diff
@@ -19,25 +42,25 @@ export default function Events(mpInstance) {
...
-        if (!mpInstance._Store.isTracking) {
-            if ('geolocation' in navigator) {
-                mpInstance._Store.watchPositionId = navigator.geolocation.watchPosition(
-                    successTracking,
-                    errorTracking
-                );
-            }
-        } else {
-            var position = {
+    this.startTracking = function(callback: ((position?: GeolocationPosition | { coords: { latitude: number | string; longitude: number | string } }) => void) | null): void {
```
alexs-mparticle: We should define a `TrackingCallback` type here.

src/events.ts
```diff
@@ -19,25 +42,25 @@ export default function Events(mpInstance) {
...
+        if (mpInstance._Store.isTracking) {
+            const position = {
                 coords: {
                     latitude: mpInstance._Store.currentPosition.lat,
                     longitude: mpInstance._Store.currentPosition.lng,
                 },
             };
             triggerCallback(callback, position);
+        } else {
+            if ('geolocation' in navigator) {
```
alexs-mparticle, line 55: If this is a TypeScript migration, why are we adding a conditional here?

src/events.ts
```diff
@@ -50,14 +73,17 @@ export default function Events(mpInstance) {
...
+        function errorTracking(): void {
             triggerCallback(callback);
             // prevents callback from being fired multiple times
             callback = null;
             mpInstance._Store.isTracking = false;
         }
 
-        function triggerCallback(callback, position) {
+        function triggerCallback(
+            callback: ((position?: GeolocationPosition | { coords: { latitude: number | string; longitude: number | string } }) => void) | null,
```
alexs-mparticle: This should be a new type definition.

src/events.ts
```diff
@@ -163,35 +194,36 @@ export default function Events(mpInstance) {
         });
 
         if (event) {
-            event.EventCategory = mpInstance._Ecommerce.convertProductActionToEventType(
+            // TODO(SDKE-1106): Remove `as Function` casts when ecommerce.js is migrated to TS
```
alexs-mparticle:
```suggestion
            // https://go.mparticle.com/work/SDKE-1106
```

src/events.ts
```diff
@@ -279,43 +311,56 @@ export default function Events(mpInstance) {
...
             customFlags
         );
 
         if (event) {
             event.EventName += 'Impression';
             event.EventCategory = Types.CommerceEventType.ProductImpression;
-            if (!Array.isArray(impression)) {
-                impression = [impression];
-            }
+            const rawList = Array.isArray(impression)
```
alexs-mparticle: I would avoid code changes, even small ones, as part of this migration.

src/events.ts
```diff
@@ -361,47 +406,48 @@ export default function Events(mpInstance) {
...
-                        window.location.href = element.href;
-                    } else if (element.submit) {
-                        element.submit();
+        domEvent: string,
+        selector: string | Node,
+        eventName: ((element: HTMLLinkElement | HTMLFormElement) => string) | string,
+        data: ((element: HTMLLinkElement | HTMLFormElement) => SDKEventAttrs) | SDKEventAttrs,
+        eventType: valueof<typeof EventType>
+    ): void {
+        let elements: ArrayLike<Element> | Element[] = [],
```
alexs-mparticle: Let's put declarations on their own lines.

src/events.ts
```diff
@@ -361,47 +406,48 @@ export default function Events(mpInstance) {
...
+            handler = (e: Event): void => {
+                const el = element as DOMHandlerElement;
```
alexs-mparticle: Why are we reassigning `element` to `el`? It doesn't help with readability.

src/events.ts
```diff
@@ -410,8 +456,8 @@ export default function Events(mpInstance) {
                     );
                 }
             },
-            element,
-            i;
+            element: Element,
+            i: number;
```
alexs-mparticle:
```suggestion
            elementIndex: number;
```

src/sdkRuntimeModels.ts
```diff
@@ -112,7 +112,7 @@ export interface SDKShoppingCart {
 }
 
 export interface SDKPromotionAction {
-    PromotionActionType: string;
+    PromotionActionType: string | valueof<typeof PromotionActionType>;
```
alexs-mparticle, line 115: can we just use `valueof<typeof PromotionActionType>` so we can control the values? `string` would allow any string, thus negating the `valueof`

src/store.ts
```diff
@@ -96,6 +96,7 @@ export interface SDKConfig {
     workspaceToken?: string;
     requiredWebviewBridgeName?: string;
     isLoggingEnabled?: boolean;
+    timeout?: number;
```
alexs-mparticle, line 99: is this a new value introduced in this pr?

test/src/tests-event-logging.ts
```diff
@@ -9,6 +9,28 @@ import {
     MPConfig,
     MessageType,
 } from './config/constants';
+import { IMParticleInstanceManager } from '../../src/sdkRuntimeModels';
+
+declare global {
```
alexs-mparticle, line 14: Can't this be imported from elsewhere?

test/src/tests-event-logging.ts
```diff
@@ -9,6 +9,28 @@ import {
...
+        }
+    }
+    function Should(obj: unknown): Should.Assertion;
+    // geomock.js custom property
+    interface Geolocation {
+        shouldFail: boolean;
+    }
+}
+
+const mParticle = window.mParticle as IMParticleInstanceManager;
```
alexs-mparticle: IIRC, this should be imported from config?

alexs-mparticle (changes requested):

src/events.ts
jaissica12: done

src/events.ts
jaissica12: This conditional was already present in the original events.js, it wasn't added as part of the migration. The only change here was flipping the if/else branch order to address a SonarCloud "unexpected negated condition" warning (the original had `if (!mpInstance._Store.isTracking)` first).

src/events.ts
jaissica12: updated

src/events.ts
jaissica12: Intentional, this is a temporary cast that will be resolved when the ecommerce module is migrated to TS. Didn't want to create an extra ticket for this so just kept comment and existing migration ticket as refrence

src/events.ts
jaissica12: The original `impression = [impression]` can't compile in TypeScript — the type checker won't allow reassigning a union-typed parameter like that. `const impressionList` is the minimal workaround to keep the same logic without using `as any`. As you mentioned, simplifying the API to only accept arrays would eliminate this normalization entirely

src/events.interfaces.ts
jaissica12: Created a follow-up ticket to track this: https://go/j/SDKE-1199

src/events.ts
jaissica12: done

src/events.ts
jaissica12: good catch, keeping it element

src/events.ts
jaissica12: renamed `i` to `elementIndex`

src/sdkRuntimeModels.ts
jaissica12: Keeping string | valueof<typeof PromotionActionType> for now, tightening to just valueof causes a downstream cast failure in sdkToEventsApiConverter.ts where it converts to PromotionActionActionEnum. Fixing that properly would require changes beyond the scope of this migration PR. We can revisit this in a follow-up.

src/store.ts
jaissica12: `SDKConfig.timeout` was already used in the original events.js (for the DOM event handler's setTimeout). It just wasn't declared in the SDKConfig interface, so TypeScript would error on access. Added the type to match the existing runtime shape.

test/src/tests-event-logging.ts
jaissica12: Added it in constants and importing from there

test/src/tests-event-logging.ts
jaissica12: updated `constants.ts` to export `mParticle` with the `IMParticleInstanceManager` type, so this test file now imports it directly from config. Removed the local cast and the `declare global` Window declaration.

src/events.ts
```diff
@@ -381,14 +429,14 @@ export default function Events(mpInstance) {
                     'DOM event triggered, handling event'
                 );
 
-                self.logEvent({
+                this.logEvent({
```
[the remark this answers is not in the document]
jaissica12: done

jaissica12:
Closing this in favor of https://github.com/mParticle/mparticle-web-sdk/pull/1342. This was an old PR which had a lot of merge conflicts

PR #1259, opened 2026-05-12
refactor: update ecommerce module to TS

jaissica12:
## Background
- Migrate remaining JavaScript modules to TypeScript for improved type safety and developer experience across the SDK codebase.
## What Has Changed
- Converted `src/ecommerce.js` to `src/ecommerce.ts` with full type annotations on all methods and parameters
- Converted `test/src/tests-eCommerce.js` to `test/src/tests-eCommerce.ts` with necessary type casts for intentional type-mismatch tests that validate error handling

## Checklist

## Additional Notes
- `var` declarations replaced with `const`/`let` (SonarCloud compliance)
- `ShoppingCart` initialization corrected from `[]` to `{}` to align with `SDKShoppingCart` interface type
- No functional/behavioral changes

- Closes https://go/j/SDKE-1108

### DIFF
--- a/src/ecommerce.ts
+++ b/src/ecommerce.ts
@@ -1,19 +1,48 @@
 import Types from './types';
 import Constants from './constants';
 import { extend } from './utils';
+import { IECommerce } from './ecommerce.interfaces';
+import { IMParticleWebSDKInstance } from './mp-instance';
+import {
+    ProductAction,
+    Product,
+    Promotion,
+    CommerceEvent,
+} from '@mparticle/event-models';
+import {
+    SDKEventAttrs,
+    SDKEventOptions,
+    TransactionAttributes,
+} from '@mparticle/web-sdk';
+import {
+    SDKEvent,
+    SDKEventCustomFlags,
+    SDKProduct,
+    SDKProductAction,
+    SDKProductImpression,
+    SDKPromotion,
+    SDKShoppingCart,
+} from './sdkRuntimeModels';
+import { valueof } from './utils';
+import { ProductActionType, PromotionActionType } from './types';
 
-var Messages = Constants.Messages;
+const Messages = Constants.Messages;
 
-export default function Ecommerce(mpInstance) {
-    var self = this;
+export default function Ecommerce(
+    this: IECommerce,
+    mpInstance: IMParticleWebSDKInstance
+): void {
+    const self = this;
+    const _self = self as { [K in keyof IECommerce]: Function };
 
     // https://go.mparticle.com/work/SQDSDKS-4801
-    this.convertTransactionAttributesToProductAction = function(
-        transactionAttributes,
-        productAction
-    ) {
+    _self.convertTransactionAttributesToProductAction = function(
+        this: IECommerce,
+        transactionAttributes: TransactionAttributes & { Step?: number; Option?: string },
+        productAction: ProductAction & SDKProductAction
+    ): void {
         if (transactionAttributes.hasOwnProperty('Id')) {
-            productAction.TransactionId = transactionAttributes.Id;
+            productAction.TransactionId = transactionAttributes.Id as string;
         }
         if (transactionAttributes.hasOwnProperty('Affiliation')) {
             productAction.Affiliation = transactionAttributes.Affiliation;
@@ -23,19 +52,19 @@ export default function Ecommerce(mpInstance) {
         }
         if (transactionAttributes.hasOwnProperty('Revenue')) {
             productAction.TotalAmount = this.sanitizeAmount(
-                transactionAttributes.Revenue,
+                transactionAttributes.Revenue as number,
                 'Revenue'
             );
         }
         if (transactionAttributes.hasOwnProperty('Shipping')) {
             productAction.ShippingAmount = this.sanitizeAmount(
-                transactionAttributes.Shipping,
+                transactionAttributes.Shipping as string,
                 'Shipping'
             );
         }
         if (transactionAttributes.hasOwnProperty('Tax')) {
             productAction.TaxAmount = this.sanitizeAmount(
-                transactionAttributes.Tax,
+                transactionAttributes.Tax as number,
                 'Tax'
             );
         }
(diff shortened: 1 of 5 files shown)


src/ecommerce.ts
```diff
@@ -365,20 +428,26 @@ export default function Ecommerce(mpInstance) {
             Id: id,
             Affiliation: affiliation,
             CouponCode: couponCode,
-            Revenue: revenue,
-            Shipping: shipping,
+            Revenue: revenue as number,
+            Shipping: shipping as string,
```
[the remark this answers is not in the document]
jaissica12: fixed

src/ecommerce.ts
```diff
@@ -1,19 +1,40 @@
...
+    SDKProductAction,
+    SDKPromotion,
+} from './sdkRuntimeModels';
+import {
+    SDKEventAttrs,
+    SDKEventOptions,
+    TransactionAttributes,
+} from '@mparticle/web-sdk';
+
+const Messages = Constants.Messages;
```
crisryantan:
```suggestion
const { Messages } = Constants;
```

test/src/tests-eCommerce.ts
```diff
@@ -121,7 +144,7 @@ describe('eCommerce', function() {
                 '400');
 
         mParticle.eCommerce.logProductAction(
-            mParticle.ProductActionType.Typo, // <------ will result in a null when converting the product action type as this is not a real value
+            (mParticle.ProductActionType as any).Typo, // <------ will result in a null when converting the product action type as this is not a real value
```
mattbodle, line 147: Remove any
As part of these changes do we have linting to stop the usage of any on updated files?

test/src/tests-eCommerce.ts
```diff
@@ -136,7 +159,7 @@ describe('eCommerce', function() {
                 'Plus',
                 'Phones',
                 'Apple',
-                '1-foo',
+                '1-foo' as any,
```
mattbodle, line 162: same for the rest here

test/src/tests-eCommerce.ts
```diff
@@ -624,7 +647,7 @@ describe('eCommerce', function() {
         const product1 = mParticle.eCommerce.createProduct('iphone', 'iphoneSKU', 999);
         const product2 = mParticle.eCommerce.createProduct('galaxy', 'galaxySKU', 799);
 
-        mParticle.eCommerce.logProductAction(mParticle.ProductActionType.Checkout, [product1, product2], null, null, {Step: 4, Option: 'Visa'});
+        mParticle.eCommerce.logProductAction(mParticle.ProductActionType.Checkout, [product1, product2], null, null, {Step: 4, Option: 'Visa'} as unknown as TransactionAttributes);
```
mattbodle, line 650: Why does it not cast correctly?
It should not need to be cast to unknown, unless the test case is to verify receiving an invalid object

src/events.ts
```diff
@@ -334,7 +334,7 @@ export default function Events(
 
     this.logPromotionEvent = function(
         promotionType: valueof<typeof PromotionActionType>,
-        promotion: SDKPromotion,
+        promotion: SDKPromotion | SDKPromotion[],
```
mattbodle, line 345: Better notation is Array<SDKPromotion>
I suggest we implement the set of TS linting rules so these are automatically verified

mattbodle (changes requested):

src/ecommerce.ts
```diff
@@ -233,14 +270,20 @@ export default function Ecommerce(mpInstance) {
...
         }
     };
 
     // https://go.mparticle.com/work/SQDSDKS-4801
-    this.extractActionAttributes = function(attributes, productAction) {
+    this.extractActionAttributes = function(
+        attributes: { [key: string]: any },
+        productAction: SDKProductAction
+    ): void {
         self.extractTransactionId(attributes, productAction);
```
mattbodle, line 287: As per the sonarqube warning above, don't assign this to self in v3 as we're using es6+ JS.
You can make this function an arrow function and this.extract() will work fine

src/ecommerce.ts
```diff
@@ -304,17 +353,17 @@ export default function Ecommerce(mpInstance) {
...
+        name: string,
+        sku: string | number,
+        price: string | number,
+        quantity?: string | number,
+        variant?: string,
+        category?: string,
+        brand?: string,
+        position?: number,
+        couponCode?: string,
+        attributes?: SDKEventAttrs
```
mattbodle, line 365: Make this a single object

mattbodle (changes requested):

src/ecommerce.ts
jaissica12: done

src/events.ts
jaissica12: Updated to Array<SDKPromotion>, adding those lint rules is a follow-up

src/ecommerce.ts
jaissica12: Was planning to do in shot for all modules, to minimize diff. Update `extractActionAttributes` is now an arrow and uses this.

src/ecommerce.ts
jaissica12: My understanding is that this would change the public positional API

test/src/tests-eCommerce.ts
jaissica12: Agreed, this isn’t an invalid object. DT `TransactionAttributes` doesn’t include checkout Step/Option, so it needed the double cast, typed that internally and dropped the cast.

test/src/tests-eCommerce.ts
jaissica12: Same as the typo case, those tests pass invalid values on purpose, so they now use unknown instead of any.

test/src/tests-eCommerce.ts
jaissica12: Removed any, Typo is still an invalid action type, now typed as unknown.

crisryantan (approved):

mattbodle (approved):

PR #1260, opened 2026-05-12
refactor: update forwarder module to TS

jaissica12:
## Background

## What Has Changed
- Converted `src/forwarders.js` to `src/forwarders.ts` with full type annotations on all methods, parameters, and return types
- Updated `src/forwarders.interfaces.ts` with corrected and expanded interface definitions to support the conversion
- Updated `src/sdkRuntimeModels.ts` with type adjustments needed for forwarder-related models
- Updated `src/store.ts` with type refinements required by the forwarder module

## Screenshots/Video
- 
Screenshot 2026-05-20 at 11 30 43 PM
Screenshot 2026-05-20 at 11 33 06 PM
Screenshot 2026-05-20 at 11 33 59 PM
Screenshot 2026-05-20 at 11 34 39 PM

## Checklist

- Closes https://go/j/SDKE-1109

### DIFF
--- a/src/forwarders.ts
+++ b/src/forwarders.ts
@@ -2,7 +2,7 @@ import filteredMparticleUser from './filteredMparticleUser';
 import { isEmpty, extend } from './utils';
 import KitFilterHelper from './kitFilterHelper';
 import Constants from './constants';
-import APIClient from './apiClient';
+import APIClient, { IForwardingStatsData } from './apiClient';
 import {
     isBlockedByForwardingRule,
     isBlockedByEventFilter,
@@ -12,11 +12,36 @@ import {
     filterBatchEventAttributes,
     filterBatchIdentities,
 } from './forwarder-utils';
+import { IMParticleWebSDKInstance } from './mp-instance';
+import { SDKEvent } from './sdkRuntimeModels';
+import {
+    IForwarders,
+    MPForwarder,
+    RegisteredKit,
+    forwardingStatsCallback as ForwardingStatsCallback,
+    ConfiguredKit,
+} from './forwarders.interfaces';
+import {
+    IConfigResponse,
+    IKitConfigs,
+    IFilteringUserAttributeValue,
+    IKitFilterSettings,
+} from './configAPIClient';
+import { IPixelConfiguration } from './cookieSyncManager';
+import { Batch } from '@mparticle/event-models';
+import { IMParticleUser, ISDKUserIdentity } from './identity-user-interfaces';
+import KitBlocker from './kitBlocking';
+import { Dictionary } from './utils';
 
 const { Modify, Identify, Login, Logout } = Constants.IdentityMethods;
 
-export default function Forwarders(mpInstance, kitBlocker) {
-    var self = this;
+export default function Forwarders(
+    this: IForwarders,
+    mpInstance: IMParticleWebSDKInstance,
+    kitBlocker: KitBlocker
+): void {
+    const self = this;
+
     this.forwarderStatsUploader = new APIClient(
         mpInstance,
         kitBlocker
@@ -27,8 +52,11 @@ export default function Forwarders(mpInstance, kitBlocker) {
         removeUserAttribute: 'removeUserAttribute',
     };
 
-    this.initForwarders = function(userIdentities, forwardingStatsCallback) {
-        var user = mpInstance.Identity.getCurrentUser();
+    this.initForwarders = function(
+        userIdentities: ISDKUserIdentity[],
+        forwardingStatsCallback: ForwardingStatsCallback
+    ): void {
+        const user = mpInstance.Identity.getCurrentUser();
         if (
             !mpInstance._Store.webviewBridgeEnabled &&
             mpInstance._Store.configuredForwarders
@@ -53,7 +81,7 @@ export default function Forwarders(mpInstance, kitBlocker) {
                         return false;
                     }
                     if (
-                        !self.isEnabledForUserAttributes(
+                        !(self.isEnabledForUserAttributes as Function)(
                             forwarder.filteringUserAttributeValue,
                             user
                         )
@@ -61,7 +89,7 @@ export default function Forwarders(mpInstance, kitBlocker) {
                         return false;
                     }
                     if (
-                        !self.isEnabledForUnknownUser(
+                        !(self.isEnabledForUnknownUser as Function)(
                             forwarder.excludeAnonymousUser,
                             user
                         )
@@ -69,7 +97,7 @@ export default function Forwarders(mpInstance, kitBlocker) {
                         return false;
                     }
 
-                    var filteredUserIdentities = mpInstance._Helpers.filterUserIdentities(
+                    const filteredUserIdentities = (mpInstance._Helpers.filterUserIdentities as Function)(
                         userIdentities,
                         forwarder.userIdentityFilters
                     );
@@ -100,7 +128,10 @@ export default function Forwarders(mpInstance, kitBlocker) {
         }
     };
 
-    this.isEnabledForUserAttributes = function(filterObject, user) {
+    this.isEnabledForUserAttributes = function(
+        filterObject: IFilteringUserAttributeValue,
+        user: IMParticleUser
+    ): boolean {
         if (
             !filterObject ||
             !mpInstance._Helpers.isObject(filterObject) ||
@@ -109,7 +140,9 @@ export default function Forwarders(mpInstance, kitBlocker) {
             return true;
         }
 
-        var attrHash, valueHash, userAttributes;
+        let attrHash: string;
+        let valueHash: string;
+        let userAttributes: Record<string, string>;
 
         if (!user) {
             return false;
@@ -117,7 +150,7 @@ export default function Forwarders(mpInstance, kitBlocker) {
             userAttributes = user.getAllUserAttributes();
         }
 
-        var isMatch = false;
+        let isMatch = false;
 
         try {
             if (
@@ -125,7 +158,7 @@ export default function Forwarders(mpInstance, kitBlocker) {
                 mpInstance._Helpers.isObject(userAttributes) &&
                 Object.keys(userAttributes).length
             ) {
-                for (var attrName in userAttributes) {
+                for (const attrName in userAttributes) {
                     if (userAttributes.hasOwnProperty(attrName)) {
                         attrHash = KitFilterHelper.hashAttributeConditionalForwarding(
                             attrName
@@ -156,7 +189,10 @@ export default function Forwarders(mpInstance, kitBlocker) {
         }
     };
 
-    this.isEnabledForUnknownUser = function(excludeAnonymousUserBoolean, user) {
+    this.isEnabledForUnknownUser = function(
+        excludeAnonymousUserBoolean: boolean,
+        user: IMParticleUser
+    ): boolean {
         if (!user || !user.isLoggedIn()) {
             if (excludeAnonymousUserBoolean) {
                 return false;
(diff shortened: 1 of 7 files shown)


src/forwarders.ts
```diff
@@ -12,11 +12,34 @@ import {
...
+    IConfigResponse,
+    IKitConfigs,
+    IFilteringUserAttributeValue,
+    IKitFilterSettings,
+} from './configAPIClient';
+import { IPixelConfiguration } from './cookieSyncManager';
+import { Batch } from '@mparticle/event-models';
+import { IMParticleUser, ISDKUserIdentity } from './identity-user-interfaces';
+import KitBlocker from './kitBlocking';
+import { Dictionary } from './utils';
```
alexs-mparticle: Clean up these imports

src/forwarders.ts
```diff
@@ -53,23 +79,23 @@ export default function Forwarders(mpInstance, kitBlocker) {
                         return false;
                     }
                     if (
-                        !self.isEnabledForUserAttributes(
+                        !(this.isEnabledForUserAttributes as Function)(
```
alexs-mparticle, line 82: Why do we need `as Function`?

src/forwarders.ts
```diff
@@ -241,13 +278,13 @@ export default function Forwarders(mpInstance, kitBlocker) {
...
                 // Check user identity filtering rules
-                clonedEvent.UserIdentities = filterUserIdentities(
+                clonedEvent.UserIdentities = (filterUserIdentities as Function)(
                     clonedEvent.UserIdentities,
                     forwarder.userIdentityFilters
                 );
 
                 // Check user attribute filtering rules
-                clonedEvent.UserAttributes = KitFilterHelper.filterUserAttributes(
+                clonedEvent.UserAttributes = (KitFilterHelper.filterUserAttributes as Function)(
```
alexs-mparticle, line 287: Same here. This feels like we're trying to bypass type safety.

src/forwarders.ts
```diff
@@ -266,15 +303,21 @@ export default function Forwarders(mpInstance, kitBlocker) {
...
+    this.sendBatchToForwarders = function(batch: Batch): void {
         if (
             mpInstance._Store.webviewBridgeEnabled ||
             !mpInstance._Store.activeForwarders
         ) {
             return;
         }
 
-        for (const forwarder of mpInstance._Store.activeForwarders) {
+        for (
```
alexs-mparticle, line 314: Is there a reason we needed to change this to a standard for loop?

src/forwarders.ts
```diff
@@ -313,15 +356,19 @@ export default function Forwarders(mpInstance, kitBlocker) {
                 const result = forwarder.processBatch(batchCopy);
 
                 if (result) {
-                    mpInstance.Logger.verbose(result);
+                    mpInstance.Logger.verbose(result as string);
```
alexs-mparticle: Isn't this already a string?

src/forwarders.ts
```diff
@@ -576,10 +640,12 @@ export default function Forwarders(mpInstance, kitBlocker) {
     // In the future, when all kits are moved to the mpConfig rather than
     // there being a separate process for MP configured kits and
     // sideloaded kits, this will need to be refactored.
-    this.processSideloadedKits = function(mpConfig) {
+    this.processSideloadedKits = function(
+        mpConfig: IConfigResponse & { sideloadedKits?: { kitInstance: { register: Function; name: string }; filterDictionary: IKitFilterSettings }[] }
```
alexs-mparticle: If we don't have a type for this, let's make one.

src/sdkRuntimeModels.ts
```diff
@@ -195,7 +195,7 @@ export interface MParticleWebSDK {
     configurePixel(config: IPixelConfiguration): void;
     endSession(): void;
     init(apiKey: string, config: SDKInitConfig, instanceName?: string): void;
-    _getActiveForwarders(): ConfiguredKit[];
+    _getActiveForwarders(): MPForwarder[];
```
alexs-mparticle: Let's be careful with this because `MPForwarder` and `ConfiguredKit` might have specific "contracts".

src/sdkRuntimeModels.ts
```diff
@@ -389,6 +389,10 @@ export interface SDKHelpersApi {
         attrs: SDKEventAttrs,
         name: string
     ): Dictionary<string> | null;
+    filterUserIdentities?(
+        userIdentitiesObject: Record<string, string>,
```
alexs-mparticle: I think we may have this type defined for this.

alexs-mparticle (changes requested):

src/forwarders.ts
jaissica12: done

src/forwarders.ts
jaissica12: The method is defined later in the constructor body but called here, TS can't verify it exists at this point. Cast has no runtime effect, it's a TS workaround for the constructor-function pattern.

src/forwarders.ts
jaissica12: Type mismatch: UserAttributes is string | string[] | null values, but filterUserAttributes expects string only. Cast avoids the TS error, I thought fixing the signature in KitFilterHelper is out of scope for this migration.

src/forwarders.ts
jaissica12: Good catch, forwarder.process() and forwarder.processBatch() already return string per the interface, so as string is unnecessary there. Removed it from processBatch.

src/forwarders.ts
jaissica12: Created a SideloadedKit interface in forwarders.interfaces.ts that reuses the existing UnregisteredKit type. Replaced the inline type.

src/sdkRuntimeModels.ts
jaissica12: Good catch, reverted to ConfiguredKit[]. Active forwarders have gone through configuration, so ConfiguredKit is the correct contract. The MPForwarder (Dictionary) type was too loose.

src/sdkRuntimeModels.ts
jaissica12: Updated to use Dictionary<string>, which is already used in kitFilterHelper.ts for the same purpose and is already imported in this file.

src/forwarders.ts
jaissica12: Reverted to the original for...of loop.

src/sdkRuntimeModels.ts
```diff
@@ -389,6 +389,10 @@ export interface SDKHelpersApi {
         attrs: SDKEventAttrs,
         name: string
     ): Dictionary<string> | null;
+    filterUserIdentities?(
+        userIdentitiesObject: Dictionary<string>,
+        filterList: number[]
+    ): ISDKUserIdentity[];
```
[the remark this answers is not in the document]
jaissica12: The Dictionary<string> type matches the actual implementation in helpers.js - it iterates with for...in and Object.keys(), expecting an object like { email: "foo@bar.com" }. The callers in forwarders.ts pass ISDKUserIdentity[] which is a pre-existing type mismatch in the JS codebase - the as Function casts already bypass the type check. Fixing this properly would require refactoring helpers.filterUserIdentities to accept both shapes, which is out of scope for this migration.

src/store.ts
```diff
@@ -197,7 +197,7 @@ export interface IStore {
     requireDelay: boolean;
     isLocalStorageAvailable: boolean | null;
     storageName: string | null;
-    activeForwarders: ConfiguredKit[];
+    activeForwarders: MPForwarder[];
```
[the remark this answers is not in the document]
jaissica12: done

src/forwarders.ts
```diff
@@ -266,15 +296,21 @@ export default function Forwarders(mpInstance, kitBlocker) {
...
             return;
         }
 
-        for (const forwarder of mpInstance._Store.activeForwarders) {
+        for (
+            let i = 0;
+            i < mpInstance._Store.activeForwarders.length;
+            i++
+        ) {
+            const forwarder = mpInstance._Store.activeForwarders[i];
```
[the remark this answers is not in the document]
jaissica12: fixed

src/forwarders.ts
```diff
@@ -24,8 +44,11 @@ export default function Forwarders(mpInstance, kitBlocker) {
         removeUserAttribute: 'removeUserAttribute',
     };
 
-    this.initForwarders = function(userIdentities, forwardingStatsCallback) {
-        var user = mpInstance.Identity.getCurrentUser();
+    this.initForwarders = function(
+        userIdentities: UserIdentities,
+        forwardingStatsCallback: forwardingStatsCallback
+    ): void {
+        let user = mpInstance.Identity.getCurrentUser();
```
crisryantan: I think this can be `const` instead. I don't see any change to the variable below

src/forwarders.ts
```diff
@@ -66,7 +89,7 @@ export default function Forwarders(mpInstance, kitBlocker) {
                         return false;
                     }
 
-                    var filteredUserIdentities = mpInstance._Helpers.filterUserIdentities(
+                    let filteredUserIdentities = mpInstance._Helpers.filterUserIdentities(
```
crisryantan: this should be `const` as well

src/forwarders.ts
```diff
@@ -106,23 +132,23 @@ export default function Forwarders(mpInstance, kitBlocker) {
             return true;
         }
 
-        var attrHash, valueHash, userAttributes;
+        let attrHash, valueHash, userAttributes;
```
crisryantan: valueHash can be const

src/forwarders.ts
```diff
@@ -162,26 +191,26 @@ export default function Forwarders(mpInstance, kitBlocker) {
...
-    this.applyToForwarders = function(functionName, functionArgs) {
+    this.applyToForwarders = function(functionName: string, functionArgs: unknown): void {
         if (mpInstance._Store.activeForwarders.length) {
             mpInstance._Store.activeForwarders.forEach(function(forwarder) {
-                var forwarderFunction = forwarder[functionName];
+                let forwarderFunction = (forwarder as any)[functionName];
                 if (forwarderFunction) {
                     try {
-                        var result = forwarder[functionName](functionArgs);
+                        let result = (forwarder as any)[functionName](functionArgs);
```
crisryantan: const here as well

src/forwarders.ts
```diff
@@ -301,21 +334,21 @@ export default function Forwarders(mpInstance, kitBlocker) {
...
         mpInstance._Store.activeForwarders.forEach(function(forwarder) {
-            var filteredUserIdentities = mpInstance._Helpers.filterUserIdentities(
+            let filteredUserIdentities = mpInstance._Helpers.filterUserIdentities(
                 userIdentities,
                 forwarder.userIdentityFilters
             );
             if (forwarder.setUserIdentity) {
                 filteredUserIdentities.forEach(function(identity) {
-                    var result = forwarder.setUserIdentity(
+                    let result = forwarder.setUserIdentity(
```
crisryantan: same here

src/forwarders.ts
```diff
@@ -327,28 +360,31 @@ export default function Forwarders(mpInstance, kitBlocker) {
...
-        var result;
+    this.setForwarderOnIdentityComplete = function(
+        user: IMParticleUser,
+        identityMethod: string
+    ): void {
+        let result;
 
         mpInstance._Store.activeForwarders.forEach(function(forwarder) {
-            var filteredUser = filteredMparticleUser(
+            let filteredUser = filteredMparticleUser(
```
crisryantan, line 385: same here

crisryantan (changes requested):
almost there but just need some correctness in variable declaration

crisryantan (approved):

PR #1261, opened 2026-05-12
refactor: update persistence module to TS

jaissica12:
## Background
- `persistence.js` was one of the remaining JavaScript modules in the SDK core. Migrating it to TypeScript continues the module-by-module TS conversion effort (same pattern as the identity and polyfill migrations) so persistence is type-checked with the rest of `src/`.

## What Has Changed
- Renamed `src/persistence.js` to `src/persistence.ts`.
- Added type annotations and converted `var` to `let`/`const`. Function style is unchanged, including `self = this`.
- Updated `IPersistence` signatures in `src/persistence.interfaces.ts` to match what the JavaScript actually did: optional `currentMPID` on `storeDataInMemory`, `encodePersistence(persistence: string)`, `decodePersistence` returning `string | void`, `getDomain(doc: Document, ...)`, `getPersistence(): IPersistenceMinified | null`, and number-based first/last seen times.
- Typed `persistence` in the Jest spec as `IPersistence`, and cast `decodePersistence` results at the two `JSON.parse` call sites in the Karma tests.
- No runtime behavior changes, and no changes to `store.ts` or `sdkRuntimeModels.ts`. Where persistence touched properties missing from shared interfaces, it uses local casts instead of widening those shared types.

## Screenshots/Video
- N/A. No user-facing or UI changes.

## Checklist

## Additional Notes
- This PR is intentionally limited to changes required to compile. Sonar flags pre-existing cognitive complexity in six persistence methods because the file is new as `.ts`; those refactors land in the stacked follow-up branch `refactor/SDKE-1107-persistence-complexity-followup` rather than here.
- Verified locally with `npx tsc --noEmit --skipLibCheck`, `npm run lint`, and the persistence Jest specs.

- Closes https://go/j/SDKE-1107

### DIFF
--- a/src/persistence.ts
+++ b/src/persistence.ts
@@ -1,36 +1,44 @@
 import Constants from './constants';
 import Polyfill from './polyfill';
 import * as Utils from './utils';
+import { IMParticleWebSDKInstance } from './mp-instance';
+import { IPersistence, IPersistenceMinified, iForwardingStatsBatches } from './persistence.interfaces';
+import { MPID, IdentityApiData } from '@mparticle/web-sdk';
+import { CookieSyncDates } from './cookieSyncManager';
+import { Dictionary } from './utils';
 
-var Base64 = Polyfill.Base64,
+const Base64 = Polyfill.Base64,
     Messages = Constants.Messages,
     Base64CookieKeys = Constants.Base64CookieKeys,
     SDKv2NonMPIDCookieKeys = Constants.SDKv2NonMPIDCookieKeys,
     StorageNames = Constants.StorageNames;
 
-export default function _Persistence(mpInstance) {
-    var self = this;
+export default function _Persistence(
+    this: IPersistence,
+    mpInstance: IMParticleWebSDKInstance
+): void {
+    const self = this;
 
     // https://go.mparticle.com/work/SQDSDKS-5022
-    this.useLocalStorage = function() {
+    this.useLocalStorage = function(): boolean {
         return (
             !mpInstance._Store.SDKConfig.useCookieStorage &&
             mpInstance._Store.isLocalStorageAvailable
         );
     };
 
-    this.initializeStorage = function() {
+    this.initializeStorage = function(): void {
         try {
-            var storage,
+            let storage: Storage,
                 localStorageData = self.getLocalStorage(),
                 cookies = self.getCookie(),
-                allData;
+                allData: IPersistenceMinified;
 
             // https://go.mparticle.com/work/SQDSDKS-6045
             // Determine if there is any data in cookies or localStorage to figure out if it is the first time the browser is loading mParticle
             if (!localStorageData && !cookies) {
                 mpInstance._Store.isFirstRun = true;
-                mpInstance._Store.mpid = 0;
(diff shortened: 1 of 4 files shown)


src/persistence.ts
```diff
@@ -301,20 +309,20 @@ export default function _Persistence(mpInstance) {
...
+    this.getLocalStorage = function(): IPersistenceMinified | null {
         if (!mpInstance._Store.isLocalStorageAvailable) {
             return null;
         }
 
-        var key = mpInstance._Store.storageName,
-            localStorageData = self.decodePersistence(
-                window.localStorage.getItem(key)
+        let key = mpInstance._Store.storageName,
+            localStorageData: any = self.decodePersistence(
```
crisryantan, line 318: since we are already fixing the types. what are your thoughts on avoiding `any` in these typing. I also see [any](https://github.com/mParticle/mparticle-web-sdk/pull/1261/changes#diff-d74addbec17763255f65ef752a33e2f0355f77d42534c05a8518ba3b3fa3bfbfR321) lets try to avoid this pattern

src/persistence.ts
```diff
@@ -1,27 +1,35 @@
...
     SDKv2NonMPIDCookieKeys = Constants.SDKv2NonMPIDCookieKeys,
     StorageNames = Constants.StorageNames;
 
-export default function _Persistence(mpInstance) {
-    var self = this;
+export default function _Persistence(
+    this: IPersistence,
+    mpInstance: IMParticleWebSDKInstance
+) {
+    const self = this;
```
crisryantan, line 20: how big of a change is this to support?

src/persistence.ts
```diff
@@ -383,47 +391,50 @@ export default function _Persistence(mpInstance) {
...
                 date.getTime() +
-                    mpInstance._Store.SDKConfig.cookieExpiration *
+                    (mpInstance._Store.SDKConfig as Dictionary)
+                        .cookieExpiration *
                         24 *
                         60 *
                         60 *
                         1000
-            ).toGMTString(),
+            ) as any).toGMTString(),
```
crisryantan, line 437: yeah let's avoid `any` here and be more type accurate since the goal here is introducing typescript we should be more type safe and accurate

crisryantan (changes requested):
added feedbacks specifically in type safety and accuracy

src/persistence.ts
jaissica12: All of the modules have this and its a huge refactor for one PR. My plan is once all the migration PRs go in, after which i will start this

src/persistence.ts
jaissica12: Agreed, we don’t want any as the pattern going forward. This migrate PR is JS→TS compile-only (localStorageData: any matches the untyped JS). Accurate cookie types (including this getLocalStorage path) are on the stacked follow-up refactor/SDKE-1107-persistence-avoid-any. Same for the other any(s) in this file.

src/persistence.ts
jaissica12: fixed in https://github.com/mParticle/mparticle-web-sdk/pull/1360

crisryantan (approved):

PR #1262, opened 2026-05-12
refactor: migrate nativeSdkHelpers module to TS

jaissica12:
## Background

## What Has Changed
- Converted `src/nativeSdkHelpers.js` to `src/nativeSdkHelpers.ts` with full type annotations on all methods, parameters, and return types
- Updated `src/nativeSdkHelpers.interfaces.ts` to correct the `sendViaBridgeV2` parameter type from `boolean` to `string`

## Screenshots/Video

## Checklist

## Additional Notes

- Closes https://go/j/SDKE-1110

### DIFF
--- a/src/nativeSdkHelpers.ts
+++ b/src/nativeSdkHelpers.ts
@@ -1,13 +1,19 @@
 import Constants from './constants';
+import { IMParticleWebSDKInstance } from './mp-instance';
+import { INativeSdkHelpers } from './nativeSdkHelpers.interfaces';
+import { Dictionary } from './utils';
 
-var Messages = Constants.Messages;
+const Messages = Constants.Messages;
 
-var androidBridgeNameBase = 'mParticleAndroid';
-var iosBridgeNameBase = 'mParticle';
+const androidBridgeNameBase = 'mParticleAndroid';
+const iosBridgeNameBase = 'mParticle';
 
-export default function NativeSdkHelpers(mpInstance) {
-    var self = this;
-    this.initializeSessionAttributes = function(apiKey) {
+export default function NativeSdkHelpers(
+    this: INativeSdkHelpers,
+    mpInstance: IMParticleWebSDKInstance
+): void {
+    const self = this;
+    this.initializeSessionAttributes = function(apiKey: string): void {
         const { SetSessionAttribute } = Constants.NativeSdkPaths;
         const env = JSON.stringify({
             key: '$src_env',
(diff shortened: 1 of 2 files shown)


rmi22186:
**Request: type the bridge surface instead of casting through `Dictionary`**

Migration looks good overall, and the `requiredWebviewBridgeName: boolean` → `string` fix is correct.

One ask before merge: please avoid the `as Dictionary` casts on `_Store` / `window` / `mParticle` and type the real fields instead. This file ends up with ~15 casts that erase type safety right where the migration should add it.

**1. Add to `IStore` (next to `webviewBridgeEnabled`):**
```ts
bridgeV1Available?: boolean;
bridgeV2Available?: boolean;
```
Then drop every `(mpInstance._Store as Dictionary).bridgeV*` access and use `mpInstance._Store.bridgeV2Available` / `bridgeV1Available` directly.

**2. Type the native bridge globals** (small local interfaces or a `Window` augmentation is fine), e.g.:
- `window.webkit?.messageHandlers`
- `window.mParticleAndroid`
- `window.mParticle.uiwebviewBridgeName` / `mpInstance.uiwebviewBridgeName`

That should remove the remaining `(window as Dictionary)` / `(window.mParticle as Dictionary)` / `(mpInstance as Dictionary)` casts.

Happy to keep this scoped to this PR — no need to refactor bridge behavior, just declare the properties the module already reads/writes.

src/nativeSdkHelpers.ts
```diff
@@ -25,28 +31,31 @@ export default function NativeSdkHelpers(mpInstance) {
...
         ) {
             return true;
         }
         // other iOS v2 bridge
         // TODO: what to do about people setting things on mParticle itself?
         if (
             window.mParticle &&
-            window.mParticle.uiwebviewBridgeName &&
-            window.mParticle.uiwebviewBridgeName === iosBridgeName
+            (window.mParticle as Dictionary).uiwebviewBridgeName &&
```
alexs-mparticle, line 56: If we can, let's use `IMParticleWebSDKInstance`

src/nativeSdkHelpers.ts
```diff
@@ -58,23 +67,23 @@ export default function NativeSdkHelpers(mpInstance) {
...
 
     this.isWebviewEnabled = function(
-        requiredWebviewBridgeName,
-        minWebviewBridgeVersion
-    ) {
-        mpInstance._Store.bridgeV2Available = self.isBridgeV2Available(
+        requiredWebviewBridgeName: string,
+        minWebviewBridgeVersion: number
+    ): boolean {
+        (mpInstance._Store as Dictionary).bridgeV2Available = self.isBridgeV2Available(
```
alexs-mparticle, line 73: Same here. We should have an `IStore` interface

src/nativeSdkHelpers.ts
```diff
@@ -25,28 +31,31 @@ export default function NativeSdkHelpers(mpInstance) {
...
             androidBridgeNameBase + '_' + bridgeName + '_v2';
-        var iosBridgeName = iosBridgeNameBase + '_' + bridgeName + '_v2';
+        const iosBridgeName = iosBridgeNameBase + '_' + bridgeName + '_v2';
 
         // iOS v2 bridge
         if (
-            window.webkit &&
-            window.webkit.messageHandlers &&
-            window.webkit.messageHandlers.hasOwnProperty(iosBridgeName)
+            (window as Dictionary).webkit &&
```
crisryantan, line 44: why did you have to do the casting here? since window would be type of window? or would it be something like
```
window & {
  webkit: {
     messageHandlers: void;
  }
}
```

jaissica12: Window doesn’t declare webkit.messageHandlers, so TS won’t allow the check without a wider type. Dictionary was the compile hammer. I’ll replace it with a small Window augmentation (handlers as a map with postMessage), not messageHandlers: void.

src/nativeSdkHelpers.ts
jaissica12: mpInstance is already `IMParticleWebSDKInstance`. window.mParticle is the instance manager, so I’ll type `uiwebviewBridgeName` there rather than cast the manager to the workspace instance.

src/nativeSdkHelpers.ts
jaissica12: Agreed — adding bridgeV1Available? / bridgeV2Available? on IStore and dropping the _Store as Dictionary casts.

crisryantan (approved):

PR #1268, opened 2026-05-19
ci: prevent v2 release sync to main

rmi22186:
## Summary
- Remove the unused legacy `.github/workflows/release.yml` workflow.
- Update staging release Step 3 labels to describe v2 branch synchronization.
- Stop Step 3 from pushing v2 release commits to `main`; it now syncs `development`, `master`, and release-order branches only.

## Context
`main` is now the v3 release line, while `master` remains v2. The v2 release process should not push release commits into v3, because main will be maintained independently with its own CI/CD.

## Verification
- `git diff --check`
- Ruby YAML parse of `.github/workflows/staging-step-3.yml`
- Confirmed no remaining `git push origin HEAD:main` in `.github/workflows`


.github/workflows/staging-step-3.yml
```diff
@@ -27,14 +27,13 @@ jobs:
           run: |
             git push origin HEAD:development
         
-        # We will eventually replace master with main
-        - name: Push release commits to main and master
+        # main is the v3 release line. Do not push v2 release commits to it.
```
rmi22186:
```suggestion
        # As of 5/19/2026, master is v2. This is subject to change in the future, but we'll need to update mpServer if this changes.
```

PR #1270, opened 2026-05-21
feat: Report deprecated method usage through logging dispatcher

rmi22186:
## Summary
- Add a core helper for deprecated API usage that preserves existing console warnings and emits structured logging dispatcher events.
- Report deprecated usage with code MP_DEPRECATED_METHOD_USAGE and message set to the deprecated method name.
- Wire existing deprecated method warnings through the helper across core, identity, session, and consent code paths.

## Verification
- Focused Jest reporting logger spec passed.
- npm run lint passed.

## Notes
- Existing unrelated workspace changes were left unstaged and are not included in this PR.

### DIFF
--- a/src/consent.ts
+++ b/src/consent.ts
@@ -9,7 +9,7 @@ import KitFilterHelper from './kitFilterHelper';
 import Constants from './constants';
 import { IMParticleUser } from './identity-user-interfaces';
 import { IMParticleWebSDKInstance } from './mp-instance';
-import { logDeprecatedApiUsage } from './reporting/deprecatedApiLogger';
+import { logDeprecatedMethodUsage } from './reporting/deprecatedMethodLogger';
 
 const { CCPAPurpose } = Constants;
 
@@ -506,8 +506,8 @@ export default function Consent(this: IConsent, mpInstance: IMParticleWebSDKInst
 
         // TODO: Can we remove this? It is deprecated.
         function removeCCPAState(this: ConsentState) {
-            logDeprecatedApiUsage(mpInstance, {
-                methodName: 'removeCCPAState',
+            logDeprecatedMethodUsage(mpInstance, {
+                methodName: 'Consent.removeCCPAState',
                 warningMessage: 'removeCCPAState is deprecated and will be removed in a future release; use removeCCPAConsentState instead',
             });
             // @ts-ignore
--- a/src/consent.ts
+++ b/src/consent.ts
@@ -506,10 +506,14 @@ export default function Consent(this: IConsent, mpInstance: IMParticleWebSDKInst
 
         // TODO: Can we remove this? It is deprecated.
         function removeCCPAState(this: ConsentState) {
-            logDeprecatedMethodUsage(mpInstance, {
-                methodName: 'Consent.removeCCPAState',
-                warningMessage: 'removeCCPAState is deprecated and will be removed in a future release; use removeCCPAConsentState instead',
-            });
+            logDeprecatedMethodUsage(
+                mpInstance.Logger,
+                mpInstance._LoggingDispatcher,
+                {
+                    methodName: 'Consent.removeCCPAState',
+                    warningMessage: 'removeCCPAState is deprecated and will be removed in a future release; use removeCCPAConsentState instead',
+                }
+            );
             // @ts-ignore
             return removeCCPAConsentState();
         }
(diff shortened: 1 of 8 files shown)


src/reporting/deprecatedApiLogger.ts
```diff
@@ -0,0 +1,24 @@
...
+export function logDeprecatedApiUsage(
+    mpInstance: DeprecatedApiLoggerInstance,
+    usage: DeprecatedApiUsage
+): void {
+    mpInstance.Logger?.warning(usage.warningMessage);
+    mpInstance._LoggingDispatcher?.log({
+        message: usage.methodName,
+        code: ErrorCodes.MP_DEPRECATED_METHOD_USAGE,
+    });
+}
```
alexs-mparticle: Consistency: We're switching between Deprecated Method vs Deprecated API in this logger, let's stick to one.

```suggestion
export function logDeprecatedMethodUsage(
    mpInstance: DeprecatedMethodLoggerInstance,
    usage: DeprecatedMethodUsage
): void {
    mpInstance.Logger?.warning(usage.warningMessage);
    mpInstance._LoggingDispatcher?.log({
        message: usage.methodName,
        code: ErrorCodes.MP_DEPRECATED_METHOD_USAGE,
    });
}
```

src/mp-instance.ts
```diff
@@ -772,14 +773,15 @@ export default function mParticleInstance(this: IMParticleWebSDKInstance, instan
              * @deprecated
              */
             add: function(product, logEventBoolean) {
-                self.Logger.warning(
-                    generateDeprecationMessage(
+                logDeprecatedApiUsage(self, {
+                    methodName: 'eCommerce.Cart.add()',
+                    warningMessage: generateDeprecationMessage(
```
alexs-mparticle: Nit: Maybe we reference the module in the log so we know where the entry point was?

```suggestion
                logDeprecatedApiUsage(self, {
                    methodName: 'mPInstance.eCommerce.Cart.add()',
                    warningMessage: generateDeprecationMessage(
```

src/identity.js
```diff
@@ -1536,12 +1544,7 @@ export default function Identity(mpInstance) {
                 newUser = mpInstance.Identity.getCurrentUser();
 
                 // https://go.mparticle.com/work/SQDSDKS-6359
-                tryOnUserAlias(
-                    prevUser,
-                    newUser,
-                    identityApiData,
-                    mpInstance.Logger
-                );
+                tryOnUserAlias(prevUser, newUser, identityApiData, mpInstance);
```
alexs-mparticle: Wy reimport the whole `mPInstance`?

src/consent.ts
```diff
@@ -505,9 +506,10 @@ export default function Consent(this: IConsent, mpInstance: IMParticleWebSDKInst
 
         // TODO: Can we remove this? It is deprecated.
         function removeCCPAState(this: ConsentState) {
-            mpInstance.Logger.warning(
-                'removeCCPAState is deprecated and will be removed in a future release; use removeCCPAConsentState instead'
-            );
+            logDeprecatedApiUsage(mpInstance, {
+                methodName: 'removeCCPAState',
```
alexs-mparticle:
```suggestion
                methodName: 'Consent.removeCCPAState',
```

alexs-mparticle (changes requested):

src/identity.js
rmi22186: because `logDeprecatedApiUsage` takes it as an argument in order to pass both `Logger` and `_LoggingDispatcher`.  Would you prefer logDeprecatedApiUsage take `Logger` and  `_LoggingDispatcher` directly?

alexs-mparticle: Yes, I think that would be preferable.

src/reporting/deprecatedMethodLogger.ts
```diff
@@ -0,0 +1,19 @@
...
+interface DeprecatedMethodUsage {
+    methodName: string;
+    warningMessage: string;
+}
+
+export function logDeprecatedMethodUsage(
+    logger: Pick<SDKLoggerApi, 'warning'>,
+    loggingDispatcher: ILoggingService | undefined,
+    usage: DeprecatedMethodUsage
+): void {
```
alexs-mparticle, line 13: Somewhat of a nit, but I would argue that `usage` is probably the most important param and should come first.

I would argue that we may want `logger` and `loggingDispatcher` to be params for a constructor, so that way we just have to call `logDeprecatedMethodUsage(usage)` but I'm it's not a hill I'd die on.

```suggestion
export function logDeprecatedMethodUsage(
    usage: DeprecatedMethodUsage
    logger: Pick<SDKLoggerApi, 'warning'>,
    loggingDispatcher: ILoggingService | undefined,
): void {
```

alexs-mparticle (approved):

PR #1273, opened 2026-05-26
refactor: migrate helpers module to TS

jaissica12:
## Background

## What Has Changed
- Converted src/helpers.js to src/helpers.ts with full type annotations on all methods, parameters, and return types
- Converted test/src/tests-helpers.js to test/src/tests-helpers.ts with necessary type adjustments
- Updated src/sdkRuntimeModels.ts with expanded SDKHelpersApi interface definitions to support the conversion
- Updated src/consent.ts with import path adjustment for the renamed module
- Replaced var declarations with const/let to satisfy SonarCloud linting rules
- Replaced const self = this pattern with arrow functions using lexical this to resolve SonarCloud violation

## Screenshots/Video

## Checklist

## Additional Notes

- Closes https://go/j/SDKE-1111

### DIFF
--- a/src/helpers.ts
+++ b/src/helpers.ts
@@ -3,12 +3,20 @@ import Constants from './constants';
 import * as utils from './utils';
 import Validators from './validators';
 import KitFilterHelper from './kitFilterHelper';
+import { IMParticleWebSDKInstance } from './mp-instance';
+import { SDKHelpersApi } from './sdkRuntimeModels';
+import { Dictionary } from './utils';
+import { IMParticleUser } from './identity-user-interfaces';
+import { MPID } from '@mparticle/web-sdk';
 
-var StorageNames = Constants.StorageNames;
+const StorageNames = Constants.StorageNames;
 
-export default function Helpers(mpInstance) {
-    var self = this;
-    this.canLog = function() {
+export default function Helpers(
+    this: SDKHelpersApi,
+    mpInstance: IMParticleWebSDKInstance
+): void {
+    const self = this;
+    this.canLog = function(): boolean {
         if (
             mpInstance._Store.isEnabled &&
             (mpInstance._Store.devToken ||
@@ -20,7 +28,7 @@ export default function Helpers(mpInstance) {
         return false;
     };
 
-    this.getFeatureFlag = function(feature) {
+    this.getFeatureFlag = function(feature: string): boolean | string {
         if (mpInstance._Store.SDKConfig.flags.hasOwnProperty(feature)) {
             return mpInstance._Store.SDKConfig.flags[feature];
         }
@@ -28,12 +36,12 @@ export default function Helpers(mpInstance) {
     };
 
     this.invokeCallback = function(
-        callback,
-        code,
-        body,
-        mParticleUser,
-        previousMpid
-    ) {
+        callback: Function,
+        code: number,
+        body: string,
+        mParticleUser?: IMParticleUser,
+        previousMpid?: MPID
+    ): void {
         if (!callback) {
             mpInstance.Logger.warning('There is no callback provided');
         }
@@ -51,9 +59,9 @@ export default function Helpers(mpInstance) {
                     },
                     getPreviousUser: function() {
                         if (!previousMpid) {
-                            var users = mpInstance.Identity.getUsers();
-                            var mostRecentUser = users.shift();
-                            var currentUser =
+                            const users = mpInstance.Identity.getUsers();
+                            let mostRecentUser = users.shift();
+                            const currentUser =
                                 mParticleUser ||
                                 mpInstance.Identity.getCurrentUser();
                             if (
@@ -78,13 +86,17 @@ export default function Helpers(mpInstance) {
         }
     };
 
-    this.invokeAliasCallback = function(callback, code, message) {
+    this.invokeAliasCallback = function(
+        callback: Function,
+        code: number,
+        message?: string
+    ): void {
         if (!callback) {
             mpInstance.Logger.warning('There is no callback provided');
         }
         try {
             if (self.Validators.isFunction(callback)) {
-                var callbackMessage = {
+                const callbackMessage: Dictionary = {
                     httpCode: code,
                 };
                 if (message) {
@@ -101,12 +113,15 @@ export default function Helpers(mpInstance) {
 
     this.extend = utils.extend;
 
-    this.createServiceUrl = function(secureServiceUrl, devToken) {
-        var serviceScheme =
+    this.createServiceUrl = function(
+        secureServiceUrl: string,
+        devToken?: string
+    ): string {
+        const serviceScheme =
             window.mParticle && mpInstance._Store.SDKConfig.forceHttps
                 ? 'https://'
                 : window.location.protocol + '//';
-        var baseUrl;
+        let baseUrl;
         if (mpInstance._Store.SDKConfig.forceHttps) {
             baseUrl = 'https://' + secureServiceUrl;
         } else {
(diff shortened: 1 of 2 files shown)


src/consent.ts
```diff
@@ -90,7 +90,7 @@ export interface IConsentStateV2DTO {
 }
 
 export interface IConsentRulesValues {
-    consentPurpose: string;
+    consentPurpose: string | number;
```
alexs-mparticle: I don't think we've ever used this as a `number`.

src/helpers.ts
```diff
@@ -20,20 +28,20 @@ export default function Helpers(mpInstance) {
         return false;
     };
 
-    this.getFeatureFlag = function(feature) {
+    this.getFeatureFlag = function(feature: string): boolean | string | null {
```
alexs-mparticle: I don't think this is ever a `boolean`

src/helpers.ts
```diff
@@ -143,17 +158,23 @@ export default function Helpers(mpInstance) {
         return xhr;
     };
 
-    this.filterUserIdentities = function(userIdentitiesObject, filterList) {
-        var filteredUserIdentities = [];
+    this.filterUserIdentities = function(
+        userIdentitiesObject: Dictionary<string>,
+        filterList: number[]
+    ): Array<{ Type: number; Identity: string }> {
```
alexs-mparticle: Nit: I believe we have a user identity type you can use here.

jaissica12: Good call, done

src/consent.ts
jaissica12: Agreed, reverted to string on #1338. Always uses string hashes, | number only covered test fixtures.

src/helpers.ts
jaissica12: I believe it can be, `processFlags` converts most server 'True'/'False' strings into booleans before they come in SDKConfig.flags (cacheIdentity, autoLogPageView). A few stay strings (offlineStorage, captureIntegrationSpecificIds.V2). Let me know if I’m missing something here

src/helpers.ts
```diff
@@ -20,20 +27,20 @@ export default function Helpers(mpInstance) {
         return false;
     };
 
-    this.getFeatureFlag = function(feature) {
+    this.getFeatureFlag = function(feature: string): boolean | string {
```
alexs-mparticle, line 32: When does this ever return a string?

src/helpers.ts
```diff
@@ -20,20 +27,20 @@ export default function Helpers(mpInstance) {
...
     };
 
     this.invokeCallback = function(
-        callback,
-        code,
-        body,
-        mParticleUser,
-        previousMpid
-    ) {
+        callback: Function,
```
alexs-mparticle: We may have a callback interface that you can use instead

src/helpers.ts
```diff
@@ -51,9 +58,9 @@ export default function Helpers(mpInstance) {
                     },
                     getPreviousUser: function() {
                         if (!previousMpid) {
-                            var users = mpInstance.Identity.getUsers();
-                            var mostRecentUser = users.shift();
-                            var currentUser =
+                            const users = mpInstance.Identity.getUsers();
+                            let mostRecentUser = users.shift();
```
alexs-mparticle, line 64: I think this needs to be hoisted higher

src/helpers.ts
```diff
@@ -129,11 +143,13 @@ export default function Helpers(mpInstance) {
 
         if (xhr && cb && 'withCredentials' in xhr) {
             xhr.onreadystatechange = cb;
-        } else if (typeof window.XDomainRequest !== 'undefined') {
+        } else if (
+            typeof (window as utils.Dictionary).XDomainRequest !== 'undefined'
```
alexs-mparticle: you can just import `Dictionary` directly

src/helpers.ts
```diff
@@ -143,17 +159,20 @@ export default function Helpers(mpInstance) {
         return xhr;
     };
 
-    this.filterUserIdentities = function(userIdentitiesObject, filterList) {
-        var filteredUserIdentities = [];
+    this.filterUserIdentities = function(
+        userIdentitiesObject: utils.Dictionary<string>,
```
alexs-mparticle: If we don't have a valid user identities interface, we can just use dictionary dicretly

```suggestion
        userIdentitiesObject: Dictionary<string>,
```

src/helpers.ts
```diff
@@ -254,7 +278,7 @@ export default function Helpers(mpInstance) {
     this.isObject = utils.isObject;
     this.decoded = utils.decoded;
     this.parseStringOrNumber = utils.parseStringOrNumber;
-    this.generateHash = utils.generateHash;
+    this.generateHash = utils.generateHash as unknown as SDKHelpersApi['generateHash'];
```
alexs-mparticle, line 283: We shouldn't need to do all this.

src/helpers.ts
rmi22186: offlineStorage returns a string '0', and captureIntegrationSpecificIds.V2 returns 'none' | 'all' | 'roktonly'

src/helpers.ts
rmi22186: I think `IdentityCallback` from @mparticle/web-sdk types

src/helpers.ts
```diff
@@ -78,13 +85,17 @@ export default function Helpers(mpInstance) {
         }
     };
 
-    this.invokeAliasCallback = function(callback, code, message) {
+    this.invokeAliasCallback = function(
+        callback: Function,
```
rmi22186: `IAliasCallback` from Definitely Typed

src/helpers.ts
rmi22186: Used within the same if block and not anywhere else, so I don't think so?

src/helpers.ts
crisryantan: should we just split this into two functions specific to its usecase?

src/helpers.ts
jaissica12: Updated `invokeCallback` to `IdentityCallback` from `identity-user-interfaces`
I believe, DT’s `IdentityCallback` from `@mparticle/web-sdk` doesn’t assign here, since `identity_type` is `string` vs ours is `SDKIdentityTypeEnum`.

src/helpers.ts
jaissica12: @rmi22186's right its within the same block, so left it in the `if (!previousMpid)`

src/helpers.ts
jaissica12: Imported `Dictionary` directly from `./utils` and dropped `utils.Dictionary`.

src/helpers.ts
jaissica12: Done, updated `userIdentitiesObject` is `Dictionary<string>`

src/helpers.ts
jaissica12: utils.generateHash returns a number, but SDKHelpersApi.generateHash is still typed as string, so this file needs the cast to compile. Dropping it by correcting both SDKHelpersApi and MParticleWebSDK.generateHash to number is in the stacked PR refactor/SDKE-1111-helpers-type-corrections.

src/helpers.ts
jaissica12: I think the the public type in DT is `AliasUsersCallback`. Switched to that

src/helpers.ts
jaissica12: **[No code changes yet]** Agreed this should be a single typed getter, not two functions, flags are boolean and string. I’d type it from `IFeatureFlags` after the remaining modules are on TS so we can drop the as boolean / as string casts in one pass. Let me know your thoughts

crisryantan: sgtm

alexs-mparticle: Works for me. I forgot how janky that function was.

rmi22186 (approved):

src/helpers.ts
alexs-mparticle: Disregard. I was hallucinating.

alexs-mparticle (approved):

PR #1276, opened 2026-06-01
feat: capture page_url in events

khushi1033:
## Background
- Events currently don't capture `page_url`, so we have no visibility into where on a site an event was fired.
- This PR adds the capability to capture `page_url` from the client, and is related to [PR1 ](https://github.com/mparticle-by-rokt/mPServer/pull/41115)and [PR2](https://github.com/mparticle-by-rokt/mPServer/pull/41294)— the mPServer PRs that handles the new field on the backend, propagating, sanitizing, and gating it for safety in case the URL contains PII.

## What Has Changed
- Bumped `@mparticle/event-models` to `^1.3.0` (adds `page_url?: string` on `CommonEventData`).
- Added optional `PageUrl?: string` to the internal `SDKEvent` interface.
- Captured on the standard `createEventObject` path and on the `createBackgroundASTEvent` path. External callers using the `toEventAPIObject` escape hatch own the full event shape and are not modified
- `convertBaseEventData` maps `sdkEvent.PageUrl` → `page_url` in the events-API payload.
- New tests cover converter mapping (set + omitted) and capture-time behavior; existing fixtures left as-is since `PageUrl` is optional.

## Screenshots/Video
Screenshot 2026-06-01 at 9 45 55 AM

## Checklist

## Additional Notes

### DIFF
--- a/dist/types/src/kitBlocking.d.ts
+++ b/dist/types/src/kitBlocking.d.ts
@@ -66,6 +66,7 @@ export default class KitBlocker {
         LaunchReferral?: string;
         ExpandedEventCount: number;
         ActiveTimeOnSite: number;
+        PageUrl?: string;
         IsBackgroundAST?: boolean;
         AlreadySentToForwarders?: boolean;
     };
@@ -110,6 +111,7 @@ export default class KitBlocker {
         LaunchReferral?: string;
         ExpandedEventCount: number;
         ActiveTimeOnSite: number;
+        PageUrl?: string;
         IsBackgroundAST?: boolean;
         AlreadySentToForwarders?: boolean;
     };
--- a/dist/types/src/kitBlocking.d.ts
+++ b/dist/types/src/kitBlocking.d.ts
@@ -66,7 +66,6 @@ export default class KitBlocker {
         LaunchReferral?: string;
         ExpandedEventCount: number;
         ActiveTimeOnSite: number;
-        PageUrl?: string;
         IsBackgroundAST?: boolean;
         AlreadySentToForwarders?: boolean;
     };
@@ -111,7 +110,6 @@ export default class KitBlocker {
         LaunchReferral?: string;
         ExpandedEventCount: number;
         ActiveTimeOnSite: number;
-        PageUrl?: string;
         IsBackgroundAST?: boolean;
         AlreadySentToForwarders?: boolean;
     };
--- a/package.json
+++ b/package.json
@@ -103,7 +103,7 @@
         "@babel/preset-env": "^7.6.0",
         "@babel/preset-typescript": "^7.6.0",
         "@mparticle/data-planning-models": "^0.1.0",
-        "@mparticle/event-models": "^1.1.9",
+        "@mparticle/event-models": "^1.3.0",
         "@rollup/plugin-babel": "6.0.3",
         "@rollup/plugin-commonjs": "25.0.4",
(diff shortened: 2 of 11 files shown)


src/serverModel.ts
```diff
@@ -310,6 +310,7 @@ export default function ServerModel(
                         event.name
                     ),
                     ActiveTimeOnSite: mpInstance._timeOnSiteTimer?.getTimeInForeground(),
+                    PageUrl: window.location.href || null,
```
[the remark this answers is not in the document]
khushi1033: Fixed! added page url to background AST events as well.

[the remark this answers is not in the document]
khushi1033: Zero internal callers of `toEventAPIObject` so no need for code change, edited PR summary to be more accurate

package.json
```diff
@@ -103,7 +103,7 @@
         "@babel/preset-env": "^7.6.0",
         "@babel/preset-typescript": "^7.6.0",
         "@mparticle/data-planning-models": "^0.1.0",
-        "@mparticle/event-models": "^1.1.9",
+        "@mparticle/event-models": "^1.3.0",
```
alexs-mparticle, line 106: I see you're bumping `event-models` but they don't look like nothing new is being used.

khushi1033: The bump is needed `convertBaseEventData` in `sdkToEventsApiConverter.ts` builds an `EventsApi.CommonEventData` typed literal that [now includes ](https://github.com/mParticle/js-models/pull/133)`page_url`, so without it the converter would fail to typecheck.

src/serverModel.ts
alexs-mparticle: We already have a utility method `getHref()` that you can reuse here. Check `/src/utils.ts`

dist/types/src/kitBlocking.d.ts
```diff
@@ -66,6 +66,7 @@ export default class KitBlocker {
         LaunchReferral?: string;
         ExpandedEventCount: number;
         ActiveTimeOnSite: number;
+        PageUrl?: string;
```
alexs-mparticle: I don't think we need to manually commit these files as they should be part of the build process.

src/batchUploader.ts
```diff
@@ -193,6 +193,7 @@ export class BatchUploader {
             SessionStartDate: sessionStartDate?.getTime() || now,
             Debug: SDKConfig.isDevelopmentMode,
             ActiveTimeOnSite: _timeOnSiteTimer?.getTimeInForeground() || 0,
+            PageUrl: window.location.href || null,
```
alexs-mparticle: We already have a utility method getHref() that you can reuse here. Check /src/utils.ts

alexs-mparticle (changes requested):

src/serverModel.ts
khushi1033: Good call! Switched to use `getHref()`

src/batchUploader.ts
khushi1033: Same as above, switched to `getHref()`

dist/types/src/kitBlocking.d.ts
khushi1033: Good catch! Removed changes here and in `sdkRuntimeModels.d.ts`

alexs-mparticle (approved):

PR #1280, opened 2026-06-16
feat: auto-calculate commerce event total amount from products

jamesnrokt:
## Summary

A commerce event's transaction-level **Total Amount** (`ProductAction.TotalAmount`) was only ever populated from a caller-supplied `transactionAttributes.Revenue`. When `Revenue` was omitted, the Web SDK left the total at `0` — even though each product's individual amount was recorded correctly. The result: multi-item checkouts reported every line item but a **zero basket total**.

This PR derives the total from the product list (`Σ quantity × price`) plus shipping and tax whenever the caller doesn't supply `Revenue`, bringing the Web SDK's behavior in line with mParticle's other platform SDKs.

## Changes

- **`src/ecommerce.js`** — new `calculateProductActionTotalAmount(productAction)` helper: sums `quantity × price` across the product list, adds `ShippingAmount` and `TaxAmount`, and assigns the result to `TotalAmount`. A total the caller explicitly provided (including `0`) is never overwritten.
- **`src/events.js`** — invoked from `logCommerceEvent`, the single chokepoint all commerce events flow through, so it applies uniformly to checkout, purchase, refund, add-to-cart, etc.
- **`test/src/tests-eCommerce.js`** — integration tests (auto-sum, shipping/tax included, explicit revenue not overridden) and unit tests for the helper.

## Behavior change / compatibility

⚠️ This changes observed output for integrations that **omit** `Revenue`: they will now emit a non-zero `Total Amount` (`tr` / `total_amount`) downstream instead of `0`. Integrations that pass `Revenue` are unaffected.

## Testing

- `npm test` — 1090 passing, 0 failing (ChromeHeadless).

### DIFF
--- a/src/ecommerce.js
+++ b/src/ecommerce.js
@@ -1,6 +1,6 @@
 import Types from './types';
 import Constants from './constants';
-import { extend } from './utils';
+import { extend, parseNumber } from './utils';
 
 var Messages = Constants.Messages;
 
@@ -45,6 +45,33 @@ export default function Ecommerce(mpInstance) {
         if (transactionAttributes.hasOwnProperty('Option')) {
             productAction.CheckoutOptions = transactionAttributes.Option;
         }
+    };
+
+    // When the caller does not supply a transaction-level total via
+    // transactionAttributes.Revenue, derive it from the product list
+    // (quantity * price) plus shipping and tax. A total that the caller provided
+    // (including 0) is never overwritten.
+    this.calculateProductActionTotalAmount = function(productAction) {
+        if (!productAction || productAction.TotalAmount != null) {
+            return productAction;
+        }
+
+        let totalAmount = 0;
+
+        if (Array.isArray(productAction.ProductList)) {
+            productAction.ProductList.forEach(function(product) {
+                totalAmount +=
+                    parseNumber(product.Quantity) * parseNumber(product.Price);
+            });
+        }
+
+        totalAmount +=
+            parseNumber(productAction.ShippingAmount) +
+            parseNumber(productAction.TaxAmount);
+
+        productAction.TotalAmount = totalAmount;
+
+        return productAction;
     };
 
     this.getProductActionEventName = function(productActionType) {
--- a/src/events.js
+++ b/src/events.js
@@ -356,6 +356,14 @@ export default function Events(mpInstance) {
                 commerceEvent.EventAttributes = attrs;
             }
 
+            // When no transaction total (Revenue) was provided, derive it from
+            // the products, shipping, and tax.
+            if (commerceEvent.ProductAction) {
+                mpInstance._Ecommerce.calculateProductActionTotalAmount(
+                    commerceEvent.ProductAction
+                );
+            }
+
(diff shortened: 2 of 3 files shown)


src/ecommerce.js
```diff
@@ -47,6 +47,34 @@ export default function Ecommerce(mpInstance) {
...
+    // When the caller does not supply a transaction-level total via
+    // transactionAttributes.Revenue, derive it from the product list
+    // (quantity * price) plus shipping and tax. A total that the caller provided
+    // (including 0) is never overwritten.
+    this.calculateProductActionTotalAmount = function(productAction) {
+        if (!productAction || productAction.TotalAmount != null) {
+            return productAction;
+        }
+
+        var parseNumber = mpInstance._Helpers.parseNumber;
```
alexs-mparticle: Grab this from `utils.ts`

src/ecommerce.js
```diff
@@ -47,6 +47,34 @@ export default function Ecommerce(mpInstance) {
...
+        var totalAmount = 0;
```
alexs-mparticle:
```suggestion
        let totalAmount = 0;
```

src/ecommerce.js
```diff
@@ -47,6 +47,34 @@ export default function Ecommerce(mpInstance) {
         }
     };
 
+    // When the caller does not supply a transaction-level total via
+    // transactionAttributes.Revenue, derive it from the product list
+    // (quantity * price) plus shipping and tax. A total that the caller provided
+    // (including 0) is never overwritten.
+    this.calculateProductActionTotalAmount = function(productAction) {
```
alexs-mparticle, line 54: Can we just use a boolean flag to enable this calculation so that we don't affect customers who may be doing their own calculations? Then in v3 we can make the calculation default/breaking?

alexs-mparticle (changes requested):

jaissica12 (changes requested):
@jamesnrokt small thing, can you retarget this to development [development](https://github.com/mParticle/mparticle-web-sdk/tree/development)

jamesnrokt:
> @jamesnrokt small thing, can you retarget this to development [development](https://github.com/mParticle/mparticle-web-sdk/tree/development)

Done

src/ecommerce.js
jamesnrokt: Done

src/ecommerce.js
jamesnrokt: Done

src/ecommerce.js
jamesnrokt: This shouldn't impact any customers who are already doing their own calculations, from the code it's only going to add this in the scenario where it's not provided. There's even a check for if the customer is providing a deliberate zero value to prevent us overwriting something intentionally set.

alexs-mparticle (approved):

jaissica12 (approved):
LGTM 🚀

jamesnrokt:
Removed `wait to merge` label as confirmed that this not only aligns with other mParticle SDKs but also has defensive mechanisms to prevent interfering with deliberate 0 values passed in

PR #1287, opened 2026-06-26
fix: prevent storage quota errors during event logging

khushi1033:
## Background

- Partner error logs showed `QuotaExceededError` surfacing from the synchronous commerce event path, including `eCommerce.logImpression`.
- The SDK had unguarded Web Storage writes in persistence and vault storage paths, so localStorage/sessionStorage quota or blocked-storage failures could escape event logging.

## What Has Changed

- Makes vault storage best-effort by catching Web Storage read/write/remove failures and returning a boolean write result.
- Guards `savePersistence()` localStorage writes so quota failures are logged without throwing.
- Adds offline batch queue trimming that drops the oldest batches until localStorage accepts the payload. If no batch can be stored (storage blocked or a single batch exceeds quota), batches are retained in memory for retry rather than dropped.
- Adds regression tests for vault write failures, persistence write failures, and offline batch trimming/retention behavior.

## Verification

- `npm run build:iife`
- `npm run build:test-bundle`
- `npm run test:jest`
- `npx karma start test/karma.config.js --browsers ChromeHeadless --single-run`

## Screenshots/Video

## Checklist

## Additional Notes

### DIFF
--- a/src/batchUploader.ts
+++ b/src/batchUploader.ts
@@ -463,13 +463,47 @@ export class BatchUploader {
             // uploading saved batches at a later time. Batches should only be removed from
             // Local Storage once we can confirm they are successfully uploaded.
             
-            this.batchVault.store(this.batchesQueuedForProcessing);
-            // Clear batch queue since everything should be in Offline Storage
-            this.batchesQueuedForProcessing = [];
+            this.storeBatchesQueuedForProcessing();
         }
 
         if (triggerFuture && !this.destroyed) {
             this.triggerUploadInterval(triggerFuture, false);
+        }
+    }
+
+    private storeBatchesQueuedForProcessing(): void {
+        const batchesToStore = this.batchesQueuedForProcessing.slice();
+        let storedBatches = batchesToStore.slice();
+
+        while (storedBatches.length) {
+            if (this.batchVault.store(storedBatches)) {
+                this.logDroppedOfflineBatches(
+                    batchesToStore.length - storedBatches.length
+                );
+                this.batchesQueuedForProcessing = [];
+                return;
+            }
+
+            storedBatches = storedBatches.slice(1);
+        }
+
+        if (this.batchVault.store([])) {
+            this.logDroppedOfflineBatches(batchesToStore.length);
+            this.batchesQueuedForProcessing = [];
+            return;
+        }
+
+        this.mpInstance.Logger.warning(
+            'Offline batch storage is unavailable. Retaining batches in memory.'
+        );
+    }
+
+    private logDroppedOfflineBatches(droppedBatchCount: number): void {
+        if (droppedBatchCount > 0) {
+            this.mpInstance.Logger.warning(
+                'Offline batch storage is over quota. Dropped ' +
+                    `${droppedBatchCount} oldest batch(es).`
+            );
         }
     }
 
--- a/src/batchUploader.ts
+++ b/src/batchUploader.ts
@@ -473,6 +473,14 @@ export class BatchUploader {
 
     private storeBatchesQueuedForProcessing(): void {
         const batchesToStore = this.batchesQueuedForProcessing.slice();
+
+        // Nothing to persist: reset offline storage to an empty state.
+        if (isEmpty(batchesToStore)) {
+            this.batchVault.store([]);
+            this.batchesQueuedForProcessing = [];
+            return;
+        }
+
         let storedBatches = batchesToStore.slice();
 
         while (storedBatches.length) {
@@ -487,12 +495,7 @@ export class BatchUploader {
             storedBatches = storedBatches.slice(1);
         }
 
-        if (this.batchVault.store([])) {
-            this.logDroppedOfflineBatches(batchesToStore.length);
-            this.batchesQueuedForProcessing = [];
-            return;
-        }
-
+        // Could not persist any batches; retain them in memory to retry later.
         this.mpInstance.Logger.warning(
             'Offline batch storage is unavailable. Retaining batches in memory.'
         );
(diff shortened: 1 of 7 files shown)


src/vault.ts
```diff
@@ -25,17 +25,22 @@ export abstract class BaseVault<StorableItem> {
...
+                stringifiedItem = JSON.stringify(item);
+            } else {
+                stringifiedItem = '';
+            }
+
+            this.storageObject.setItem(this._storageKey, stringifiedItem);
+            this.contents = item;
+            return true;
+        } catch (e) {
+            return false;
```
alexs-mparticle: `catch` collapses every failure to `false`. The trim loop reads `false` as "over quota → drop oldest", but storage-disabled (`SecurityError`) fails identically for every subset and will loop-drop all batches before retaining. Worth distinguishing the cause so only real quota errors trigger trimming.

src/batchUploader.ts
```diff
@@ -463,16 +463,53 @@ export class BatchUploader {
...
+            }
+
+            storedBatches = storedBatches.slice(1);
+        }
+
+        // Could not persist any batches; retain them in memory to retry later.
+        this.mpInstance.Logger.warning(
+            'Offline batch storage is unavailable. Retaining batches in memory.'
+        );
+    }
```
alexs-mparticle, line 513: Decision, mutation, and I/O are tangled here: double `slice()`, and the drop count is computed at the call site even though `logDroppedOfflineBatches` re-guards `> 0`. An index-based loop drops both and the index *is* the count:
```suggestion
    private storeBatchesQueuedForProcessing(): void {
        const batches = this.batchesQueuedForProcessing;

        // Nothing to persist: reset offline storage to an empty state.
        if (isEmpty(batches)) {
            this.batchVault.store([]);
            this.batchesQueuedForProcessing = [];
            return;
        }

        // Drop oldest batches until storage accepts the payload.
        for (let dropped = 0; dropped < batches.length; dropped++) {
            if (this.batchVault.store(batches.slice(dropped))) {
                this.logDroppedOfflineBatches(dropped);
                this.batchesQueuedForProcessing = [];
                return;
            }
        }

        // Could not persist any batches; retain them in memory to retry later.
        this.mpInstance.Logger.warning(
            'Offline batch storage is unavailable. Retaining batches in memory.'
        );
    }
```
Note: each iteration still re-`JSON.stringify`s the whole array inside `store()`, so this stays O(N²) on the unload path. If backlogs can get large, size-budget the cut instead of retrying.

alexs-mparticle (commented):
A few notes toward making the storage paths purer — mostly separating decisions from side effects and consolidating the quota handling. Suggestions inline.

src/vault.ts
```diff
@@ -44,11 +49,14 @@ export abstract class BaseVault<StorableItem> {
...
-        // TODO: Handle cases where Local Storage is unavailable
         // https://go.mparticle.com/work/SQDSDKS-5022
-        const item: string = this.storageObject.getItem(this._storageKey);
+        try {
+            const item: string = this.storageObject.getItem(this._storageKey);
 
-        this.contents = item ? JSON.parse(item) : null;
+            this.contents = item ? JSON.parse(item) : null;
+        } catch (e) {
+            this.contents = null;
```
alexs-mparticle, line 77: `retrieve()` now swallows parse errors and returns `null`, but callers spread it unguarded — `push(...this.eventVault.retrieve())` (L94) and `unshift(...this.batchVault.retrieve())` (L429). So corrupt/empty storage becomes `TypeError: null is not iterable` at the call site (and `prepareAndUpload` rejects, so the interval never re-arms). Guard the spreads:
```ts
this.batchesQueuedForProcessing.unshift(...(this.batchVault.retrieve() ?? []));
```

src/vault.ts
khushi1033: Good catch - added a StorageResult enum so the caller knows why a write failed.

src/batchUploader.ts
khushi1033: Refactored `storeBatchesQueuedForProcessing` to a single index-based loop that tracks how many batches were dropped, removing the repeated slice()/length math

src/vault.ts
khushi1033: Fixed. Guarded both spreads with ?? [] (push(...(this.eventVault.retrieve() ?? [])) and the batch unshift), so corrupt/empty storage no longer throws null is not iterable and prepareAndUpload won't reject and leave the interval un-armed

src/batchUploader.ts
```diff
@@ -463,16 +467,56 @@ export class BatchUploader {
...
+
+            if (result === StorageResult.Unavailable) {
+                break;
+            }
+        }
+
+        // Could not persist any batches; retain them in memory to retry later.
+        this.mpInstance.Logger.warning(
+            'Offline batch storage is unavailable. Retaining batches in memory.'
+        );
```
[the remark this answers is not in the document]
khushi1033: Fixed. The loop now logs the message that matches the actual cause: Unavailable returns immediately with the "unavailable" warning, and a fully-exhausted quota path logs a dedicated "over quota" warning.

src/batchUploader.ts
```diff
@@ -463,16 +467,56 @@ export class BatchUploader {
...
     }
 
+    private storeBatchesQueuedForProcessing(): void {
+        const batches = this.batchesQueuedForProcessing;
+
+        // Nothing to persist: reset offline storage to an empty state.
+        if (isEmpty(batches)) {
+            this.batchVault.store([]);
+            this.batchesQueuedForProcessing = [];
+            return;
```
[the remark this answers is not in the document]
khushi1033: Looked into this and the duplicate path isn't actually reachable: the clear operations (store([]) / purge()) write an empty string / remove the key, which can't exceed quota; and if storage is truly unavailable for writes, the later retrieve() also throws and returns null (now guarded with ?? []), so nothing gets re-read or re-uploaded. I've left it as-is to avoid adding handling for an unreachable case.

alexs-mparticle (approved):

PR #1296, opened 2026-07-28
fix(pre-init): drain ready queue before processing to stop re-entrant re-execution

rmi22186:
## Summary
- Drain `_preInit.readyQueue` in place (`splice`) **before** iterating, so a synchronous re-entry from cache-hit `identify` → `parseIdentityResponse` → `processReadyQueue` sees an empty queue and cannot re-execute the same items.
- Keeps the item-level `.slice()` from #1292 as defense in depth.
- Updates Jest coverage: queue is emptied after drain, second drain is a no-op, and mid-drain re-entry does not duplicate calls.

## Why
After #1292 stopped mutating queued items, re-entrant drains started successfully re-running the same entries. That shows up in prod as:
- duplicate `Identity.*` / `logEvent` side effects
- recursively nested `Unable to compute proper mParticle function Error: Unable to compute...` reports (only on `2.73.0`, starting at the #1292 deploy)

## Test plan
- [x] `npm run test:jest -- --testPathPattern=pre-init-utils.spec`
- [ ] Confirm nested `Unable to compute proper mParticle function Error:` volume drops after rollout
- [ ] Spot-check that `IDENTITY_REQUEST` no longer shows duplicate-driven nesting for `2.73.x+`

### DIFF
--- a/src/pre-init-utils.ts
+++ b/src/pre-init-utils.ts
@@ -47,8 +47,6 @@ export const processReadyQueue = (
 };
 
 const formatReadyQueueItemError = (e: unknown): string => {
-    // Only Errors and strings are described; anything else would stringify to
-    // "[object Object]" and tell us nothing.
     let detail = 'unknown error';
     if (e instanceof Error) {
         detail = e.message;
--- a/src/pre-init-utils.ts
+++ b/src/pre-init-utils.ts
@@ -1,5 +1,6 @@
 import { IPixelConfiguration } from './cookieSyncManager';
 import { MPForwarder } from './forwarders.interfaces';
+import { Logger } from './logger';
 import { IntegrationDelays } from './mp-instance';
 import { isEmpty, isFunction } from './utils';
 
@@ -11,7 +12,10 @@ export interface IPreInit {
     isDevelopmentMode?: boolean;
 }
 
-export const processReadyQueue = (readyQueue): Function[] => {
+export const processReadyQueue = (
+    readyQueue,
+    logger: Logger
+): Function[] => {
     if (isEmpty(readyQueue)) {
         return [];
     }
@@ -35,18 +39,14 @@ export const processReadyQueue = (readyQueue): Function[] => {
                 processPreloadedItem(readyQueueItem);
             }
         } catch (e) {
-            logReadyQueueItemError(e);
+            logger.error(formatReadyQueueItemError(e));
         }
     });
 
     return [];
 };
 
-const logReadyQueueItemError = (e: unknown): void => {
-    if (typeof window === 'undefined') {
-        return;
-    }
-
+const formatReadyQueueItemError = (e: unknown): string => {
     // Only Errors and strings are described; anything else would stringify to
     // "[object Object]" and tell us nothing.
     let detail = 'unknown error';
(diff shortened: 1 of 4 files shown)


src/pre-init-utils.ts
```diff
@@ -12,25 +12,34 @@ export interface IPreInit {
...
+    // "Unable to compute proper mParticle function" errors when an item fails.
+    const items = readyQueue.splice(0, readyQueue.length);
+
+    items.forEach(readyQueueItem => {
+        if (isFunction(readyQueueItem)) {
+            readyQueueItem();
+        } else if (Array.isArray(readyQueueItem)) {
+            processPreloadedItem(readyQueueItem);
+        }
+    });
```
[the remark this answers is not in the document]
rmi22186: Agreed on the mechanics — drain-first + an uncaught throw would drop siblings.

Resolved by catching per item inside `processReadyQueue`: log via `Logger.error`, then continue the local `items` array so siblings still run once in this pass. Nothing is re-queued (that would reopen re-entrancy / nested wrapping).

Also worth noting: on the main identity path `isInitialized` is set *before* the drain, so leftovers that used to remain on the shared array after a throw were unlikely to be reliably re-drained anyway.

rmi22186:
Addressed the two SonarCloud findings on `logReadyQueueItemError` in 080058ab; quality gate is now passing with 0 open issues.

- **`typescript:S6551`** (object stringification) — `String(e)` on an `unknown` throw would render `[object Object]`, which is noise in the log. Now only `Error` and `string` throws are described; anything else logs `unknown error`. Every throw on this path is a real `Error` (`processPreloadedItem` throws `new Error(...)`, and queued callbacks are app code), so no useful detail is lost.
- **`typescript:S2486`** (empty catch) — the `try/catch` around the logger lookup is gone in favour of optional chaining, plus an early return when `window` is undefined so the CJS/Node build can't hit a `ReferenceError`. The residual risk is a pathological `Logger.error` that itself throws, which would escape the per-item catch and abort the rest of the drain. That is not a real failure mode — `Logger.error` is an internal SDK function that only formats a string and calls `console`.

src/pre-init-utils.ts
```diff
@@ -12,25 +12,57 @@ export interface IPreInit {
...
+    // Only Errors and strings are described; anything else would stringify to
+    // "[object Object]" and tell us nothing.
+    let detail = 'unknown error';
+    if (e instanceof Error) {
+        detail = e.message;
+    } else if (typeof e === 'string') {
+        detail = e;
+    }
+
+    const logger = (window as any)?.mParticle?.getInstance?.()?.Logger;
```
alexs-mparticle: This function seems to be doing a lot of bypassing of type safety and assumes the logger _might_ exist, rather than going through our standard logging flow.

I think we should either pass the logger into the module directly and use this method to just render a string and be more consistent with the rest of our logging behavior.

alexs-mparticle (changes requested):

src/pre-init-utils.ts
rmi22186: Done in 3f442813 — `processReadyQueue(readyQueue, logger)` now takes the instance `Logger` from each caller (`self.Logger` / `mpInstance.Logger` in mp-instance and identity). The helper only formats the message and calls `logger.error(...)`; no more `window.mParticle.getInstance()` lookup.

src/pre-init-utils.ts
```diff
@@ -11,26 +12,55 @@ export interface IPreInit {
...
+            logger.error(formatReadyQueueItemError(e));
+        }
+    });
+
     return [];
 };
 
+const formatReadyQueueItemError = (e: unknown): string => {
+    // Only Errors and strings are described; anything else would stringify to
+    // "[object Object]" and tell us nothing.
```
alexs-mparticle: This comment feels unnecessary.

```suggestion
```

alexs-mparticle (approved):

PR #1298, opened 2026-07-28
refactor: extract getErrorMessage utility for caught-error logging

rmi22186:
## Summary
- Add `getErrorMessage(e, fallback?)` in `utils.ts` to normalize caught values into loggable strings (Error → message, string → as-is, else → `'unknown error'`).
- Replace the local `formatReadyQueueItemError` helper introduced in #1296 with a one-line prefix + `getErrorMessage`.
- Migrate the duplicated `instanceof Error ? .message : String(e)` / `(e as Error).message || String(e)` patterns in `roktManager`, `identityApiClient`, `identity/search`, and `identity-utils`.

Follow-up from Alex's review on #1296: the ready-queue formatter was doing the same work already copied across several modules.

Left alone on purpose: `Promise.reject(error instanceof Error ? error : new Error(...))` in RoktManager — that preserves Error objects for rejection, not message extraction.

## Test plan
- [x] Jest: `utils.spec.ts` (new `getErrorMessage` cases) + `pre-init-utils.spec.ts`
- [x] `npm run build:iife` (rollup TypeScript check)
- [x] `npm run lint`
- [ ] CI green on this PR (Jest matrix, Core Tests, integrations)

### DIFF
--- a/src/identityApiClient.ts
+++ b/src/identityApiClient.ts
@@ -6,7 +6,7 @@ import {
     IFetchPayload,
 } from './uploaders';
 import { CACHE_HEADER } from './identity-utils';
-import { obfuscateData, parseNumber, valueof } from './utils';
+import { obfuscateData, parseNumber, valueof, getErrorMessage } from './utils';
 import {
     IAliasCallback,
     IAliasRequest,
@@ -175,7 +175,7 @@ export default function IdentityAPIClient(
             Logger.verbose(message);
             invokeAliasCallback(aliasCallback, response.status, errorMessage);
         } catch (e) {
-            const errorMessage = (e as Error).message || e.toString();
+            const errorMessage = getErrorMessage(e);
             Logger.error('Error sending alias request to mParticle servers. ' + errorMessage);
             invokeAliasCallback(
                 aliasCallback,
@@ -338,7 +338,7 @@ export default function IdentityAPIClient(
                 mpInstance.captureTiming(`${requestCount}-identityRequestEnd`);
             }
 
-            const errorMessage = (err as Error).message || err.toString();
+            const errorMessage = getErrorMessage(err);
             const msg = 'Error sending identity request to servers' + ' - ' + errorMessage;
             Logger.error(msg);
             errorReporter?.report({
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -625,11 +625,6 @@ function extend(...args: any[]): any {
     return target;
 }
 
-/**
- * Normalize a caught value into a loggable string.
- * Prefer Error.message and plain strings; anything else would stringify to
- * "[object Object]" and tell us nothing, so fall back to a fixed label.
- */
 const getErrorMessage = (
     e: unknown,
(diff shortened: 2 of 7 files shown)


src/identityApiClient.ts
```diff
@@ -175,7 +175,7 @@ export default function IdentityAPIClient(
             Logger.verbose(message);
             invokeAliasCallback(aliasCallback, response.status, errorMessage);
         } catch (e) {
-            const errorMessage = (e as Error).message || e.toString();
+            const errorMessage = getErrorMessage(e);
             Logger.error('Error sending alias request to mParticle servers. ' + errorMessage);
```
alexs-mparticle, line 179:
```suggestion
Logger.error('Error sending alias request to mParticle servers. ' + getErrorMessage(e));
```

src/identityApiClient.ts
```diff
@@ -338,7 +338,7 @@ export default function IdentityAPIClient(
                 mpInstance.captureTiming(`${requestCount}-identityRequestEnd`);
             }
 
-            const errorMessage = (err as Error).message || err.toString();
+            const errorMessage = getErrorMessage(err);
             const msg = 'Error sending identity request to servers' + ' - ' + errorMessage;
```
alexs-mparticle, line 342:
```suggestion
const msg = 'Error sending identity request to servers' + ' - ' + getErrorMessage(err);
```

src/utils.ts
```diff
@@ -625,6 +625,24 @@ function extend(...args: any[]): any {
     return target;
 }
 
+/**
+ * Normalize a caught value into a loggable string.
+ * Prefer Error.message and plain strings; anything else would stringify to
+ * "[object Object]" and tell us nothing, so fall back to a fixed label.
+ */
```
alexs-mparticle:
```suggestion
```

alexs-mparticle (changes requested):

src/identityApiClient.ts
rmi22186: Kept the `const` on purpose — `errorMessage` is reused by `invokeAliasCallback` below. Inlining only into `Logger.error` would either call `getErrorMessage` twice or leave a now-redundant local. Same pattern as the identity catch: one extraction, then log + callback both consume it.

src/identityApiClient.ts
rmi22186: Same as the alias catch — `errorMessage` is passed to `invokeCallback` (without the log prefix), so we need the local. Inlining into `msg` alone would force a second `getErrorMessage(err)` call for the callback.

src/utils.ts
rmi22186: Dropped in c99d3cab.

src/identityApiClient.ts
rmi22186: Since we are also using errorMessage on line 183 as part of invokeAliasCallback, we should keep this the way it is.

alexs-mparticle (approved):

PR #1299, opened 2026-07-28
feat: auto-log page view on init behind AutoLogPageView feature flag

alexs-mparticle:
## Summary

Implements Phase 1 (Web SDK) of Automatic Page Views: a server-delivered `AutoLogPageView` feature flag that, when enabled, logs a single page view on init immediately after the AST. Off by default.

## Changes

- **`src/constants.ts`** — add `AutoLogPageView: 'autoLogPageView'` to the `FeatureFlags` map (camelCase wire value, matching the case-sensitive server contract).
- **`src/store.ts`** — `processFlags` normalizes the flag with `=== 'True'` (defaults false when absent); add `autoLogPageView?: boolean` to `IFeatureFlags`.
- **`src/mp-instance.ts`** — when the flag is on, fire a page view right after `logAST()`. New lifecycle order: Session Start → Application State Transition → Page View. Fires via the internal `_Events.logEvent` path (not public `logPageView`) because `_Store.isInitialized` is still false at this point — the public method would defer the event onto the `readyQueue` and fire it out of lifecycle order. Mirrors the manual no-arg page view (name `PageView`, attrs `{ hostname, title }`); `PageUrl` is auto-stamped by `createEventObject`.

## Tests

- **`test/src/tests-store.ts`** — updated the two `processFlags` deep-equal unit tests for the new flag.
- **`test/src/tests-event-logging.js`** — new flag-on integration tests: flag-off baseline (no page view), explicit `'False'` (no page view), `'True'` logs exactly one page view with the expected name/attrs, and AST-before-PageView ordering within the batch.

Existing init tests are intentionally left unchanged as the flag-off baseline.


src/events.js
```diff
@@ -99,6 +99,18 @@ export default function Events(mpInstance) {
         self.logEvent({ messageType: Types.MessageType.AppStateTransition });
     };
 
+    this.logPageView = function() {
```
rmi22186, line 102: Should we be passing some sort of attribute through here to indicate it's an auto page view vs a regular pageview that was sent by the user?  I know the name may indicate this, but i think you mentioned previously an attribute that can be placed somewhere on the batch to indicate as such.

alexs-mparticle: We'd have to add this in later since it would require a backend change. It would be similar to `IsBackgroundAST`

rmi22186 (approved):

PR #1303, opened 2026-07-30
feat: forward active time-on-site to Rokt selectPlacements

alexs-mparticle:
## What

Wires the SDK's existing **active time-on-site** value into the `attributes` map passed to Rokt's `selectPlacements`, so it flows downstream (Rokt Kit → Rokt Web SDK → Transactions API).

The value is injected into `enrichedAttributes` in `roktManager.ts`, but only when there is a real, non-zero value to report:

```ts
const timeOnSite = this.store?.getTimeOnSite?.();

const enrichedAttributes: RoktAttributes = {
    ...mappedAttributes,
    ...(sandboxValue !== null ? { sandbox: sandboxValue } : {}),
    ...(timeOnSite ? { active_time_on_site_ms: timeOnSite } : {}),
};
```

## How

- **New `IStore.getTimeOnSite()` accessor** (`store.ts`) returning `mpInstance._timeOnSiteTimer?.getTimeInForeground()`. This keeps `RoktManager`'s existing dependency surface (it holds `store`, not `mpInstance`) rather than widening `RoktManager.init`.
- The accessor returns **`number | undefined`**: the live active-foreground time in ms when the timer exists, or `undefined` when `_timeOnSiteTimer` isn't initialized yet. It deliberately does **not** fall back to `0` — a `0` would tell downstream consumers the user spent zero active time on site, when the truth is "no data available." Returning `undefined` keeps the accessor honest.
- `getTimeInForeground()` computes the current active segment on read. It counts time the tab is visible + focused and excludes backgrounded time, exactly the value already stamped on events as `ActiveTimeOnSite` (`serverModel.ts`, `batchUploader.ts`).
- The call site reads the accessor directly and uses a truthy conditional-spread, mirroring the `sandboxValue` pattern directly above it. Because both `undefined` (uninitialized timer) and a legitimate `0` are falsy, the `timeOnSite ? … : {}` guard omits the attribute in both cases — a zero value is bad data for downstream processing at Rokt, so it is not sent.

## Decisions

| Question | Decision | Rationale |
|---|---|---|
| Which value | `getTimeInForeground()` | The single active-engagement value the batch already uses; no second value exists |
| Units | milliseconds | Timer returns ms; matches downstream precedent |
| Attribute key | `active_time_on_site_ms` | Already the canonical Events API field name in `sdkToEventsApiConverter.ts`; snake_case aligns with the Transactions API `time_on_site` lineage |
| Accessor return type | `number \| undefined` | A `0` fallback misreports "no data" as "zero active time"; `undefined` is the honest signal for an uninitialized timer |
| Zero / undefined | Omit the attribute | Zero is bad data downstream at Rokt; reading the accessor directly and using a truthy spread drops both `0` and `undefined`, mirroring `sandboxValue` |

## Testing

- `active_time_on_site_ms` is included in `selectPlacements` offers requests when the timer reports a positive value.
- The attribute is **omitted** when the timer reports `0`, when the timer is not initialized (`getTimeOnSite` returns `undefined`), and when the accessor is unavailable on the store.

### DIFF
--- a/src/roktManager.ts
+++ b/src/roktManager.ts
@@ -354,10 +354,12 @@ export default class RoktManager {
 
             this.setUserAttributes(mappedAttributes);
 
+            const timeOnSite = this.store?.getTimeOnSite?.() ?? 0;
+
             const enrichedAttributes: RoktAttributes = {
                 ...mappedAttributes,
                 ...(sandboxValue !== null ? { sandbox: sandboxValue } : {}),
-                active_time_on_site_ms: this.store?.getTimeOnSite?.() ?? 0,
+                ...(timeOnSite > 0 ? { active_time_on_site_ms: timeOnSite } : {}),
             };
 
             // Propagate email from current user identities if not already in attributes
--- a/src/roktManager.ts
+++ b/src/roktManager.ts
@@ -357,6 +357,7 @@ export default class RoktManager {
             const enrichedAttributes: RoktAttributes = {
                 ...mappedAttributes,
                 ...(sandboxValue !== null ? { sandbox: sandboxValue } : {}),
+                active_time_on_site_ms: this.store?.getTimeOnSite?.() ?? 0,
             };
 
             // Propagate email from current user identities if not already in attributes
--- a/src/roktManager.ts
+++ b/src/roktManager.ts
@@ -354,12 +354,12 @@ export default class RoktManager {
 
             this.setUserAttributes(mappedAttributes);
 
-            const timeOnSite = this.store?.getTimeOnSite?.() ?? 0;
+            const timeOnSite = this.store?.getTimeOnSite?.();
 
             const enrichedAttributes: RoktAttributes = {
                 ...mappedAttributes,
                 ...(sandboxValue !== null ? { sandbox: sandboxValue } : {}),
-                ...(timeOnSite > 0 ? { active_time_on_site_ms: timeOnSite } : {}),
+                ...(timeOnSite !== undefined ? { active_time_on_site_ms: timeOnSite } : {}),
             };
 
             // Propagate email from current user identities if not already in attributes
--- a/src/roktManager.ts
+++ b/src/roktManager.ts
@@ -354,12 +354,12 @@ export default class RoktManager {
 
             this.setUserAttributes(mappedAttributes);
 
-            const timeOnSite = this.store?.getTimeOnSite?.() || null;
+            const timeOnSite = this.store?.getTimeOnSite?.();
 
             const enrichedAttributes: RoktAttributes = {
                 ...mappedAttributes,
                 ...(sandboxValue !== null ? { sandbox: sandboxValue } : {}),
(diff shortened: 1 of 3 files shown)


rmi22186 (dismissed):
LGTM!

src/roktManager.ts
```diff
@@ -357,6 +357,7 @@ export default class RoktManager {
             const enrichedAttributes: RoktAttributes = {
                 ...mappedAttributes,
                 ...(sandboxValue !== null ? { sandbox: sandboxValue } : {}),
+                active_time_on_site_ms: this.store?.getTimeOnSite?.() ?? 0,
```
jamesnrokt: A zero value would be bad data for the downstream processing at Rokt. Would be better to only add the attribute if it's a non zero value

jamesnrokt (changes requested):

src/roktManager.ts
rmi22186: good call out

src/roktManager.ts
```diff
@@ -354,9 +354,12 @@ export default class RoktManager {
 
             this.setUserAttributes(mappedAttributes);
 
+            const timeOnSite = this.store?.getTimeOnSite?.() ?? 0;
```
jamesnrokt: Couldn't we get `getTimeOnSite` to just return undefined if not set due to ForegroundTimer not being init yet.

```suggestion
            const timeOnSite = this.store?.getTimeOnSite?.();
```

It would then align with how we're treating `sandboxValue` below too without us having to know if it's a legit zero or a hidden undefined

src/store.ts
```diff
@@ -232,6 +232,7 @@ export interface IStore {
     setUserIdentities?(mpid: MPID, userIdentities: UserIdentities): void;
     getRoktAccountId?(): string;
     setRoktAccountId?(accountId: string): void;
+    getTimeOnSite?(): number;
```
jamesnrokt: Related to the above using a legitimate undefined would be clearer here. It's not actually zero so we shouldn't tell consumers of the API that it is


```suggestion
    getTimeOnSite?(): number | undefined;
```

jamesnrokt (commented):
From a code clarity POV is there any reason not to be using undefined for this use case

src/roktManager.ts
```diff
@@ -354,9 +354,12 @@ export default class RoktManager {
 
             this.setUserAttributes(mappedAttributes);
 
+            const timeOnSite = this.store?.getTimeOnSite?.() || null;
+
             const enrichedAttributes: RoktAttributes = {
                 ...mappedAttributes,
                 ...(sandboxValue !== null ? { sandbox: sandboxValue } : {}),
+                ...(timeOnSite !== null ? { active_time_on_site_ms: timeOnSite } : {}),
```
rmi22186: this is nifty!

rmi22186 (dismissed):

jamesnrokt (approved):
Approving code but questioning if the base branch should be master?

PR #1318, opened 2026-08-12
feat: Gate google_hm behind enableHmTag setting + web-safe base64

crisryantan:
## Summary

Layers the opt-in gate onto #1246 (targets Rob's branch, not master):

- `google_hm` is now only appended for DoubleclickDFP (module 41) when the pixel config delivers the `enableHmTag` setting as `'True'` — the UI toggle shipped in mPServer #42288 ([CORECDP-3674]). Google errors on accounts not provisioned for Hosted Matching, so default is off; the check is case-insensitive and treats absent/`'False'`/anything-else as off.
- The value is now **web-safe unpadded base64** (RFC 4648 §5) via a new `toWebSafeBase64` util instead of raw `btoa` — numeric MPIDs always produce `=` padding, which Google's cookie matching spec rejects (their example is unpadded `MTIzNDU2`): https://developers.google.com/authorized-buyers/rtb/cookie-guide

Verified against a live prod config (`.../js/v2/e207c24e.../config`) which delivers `"settings": { "enableHmTag": "True" }` on the GMP pixel config.

## Test plan

- [x] Updated the two existing GMP specs (opt-in wired in; non-GMP module unaffected even with the setting present)
- [x] New specs: setting absent → no `google_hm`; `'False'` → no `google_hm`; lowercase `'true'` → sent; padded-MPID case asserts unpadded value
- [x] Unit tests for `toWebSafeBase64` (padding strip, `+/` → `-_`, already-safe values)
- [x] Local verification against the PR CI gates: full Jest suite 533/533 (23 suites); Karma core suite 1041/1041 (ChromeHeadless); `npm run lint` output identical to base branch; Prettier gate unaffected (checks `.js` only); `build:iife` clean

## Tested Locally
Opt-in ON (enableHmTag: 'True'): pixel fires with google_hm = unpadded web-safe base64 of the MPID, and Google's follow-up redirect echoes it back (google_tc=), confirming their service accepts the format:
hm_on_result
Opt-in OFF (setting absent): the cookie sync still fires normally — same URL, no google_hm parameter:
hm_off_result

### HAR validation (real browser, live Google endpoint)

A DevTools HAR capture of the opt-in ON flow confirms the full exchange end to end (file available on request — GitHub doesn't accept `.har` attachments):

1. `POST identity.mparticle.com/v1/identify` → 200, MPID `5405420562044398883`.
2. `GET cm.g.doubleclick.net/pixel?google_nid=mparticle_dmp&google_cm&MPID=5405420562044398883&esid=69467&env=2&google_hm=NTQwNTQyMDU2MjA0NDM5ODg4Mw` → **302**. The `google_hm` value decodes exactly to the MPID and is spec-compliant web-safe base64 (no `=` padding, no `+`/`/`).
3. **No `google_error` parameter anywhere in the chain** — per [Google's cookie guide](https://developers.google.com/authorized-buyers/rtb/cookie-guide#redirect-url-parameters), a rejected hosted-match-table write is signaled via `google_error=1–5` on the redirect, so its absence means the write was accepted.
4. Google's 302 redirected to `cookiesync.mparticle.com/v1/sync/?id=CAESEP…&google_cver=1` → **204**: Google handed its encrypted user ID back and mParticle stored it — i.e. both directions verified in one exchange (`google_hm` into Google's hosted match table, `GOOGLE_USER_ID` back into mParticle's, with the pre-existing Partner ID behavior intact).

## Note on the BrowserStack workflow failure

The push-triggered "BrowserStack Test" failure on this branch is pre-existing, not from this change:

- All modern browsers pass 1041/1041 with this commit (Chrome 50–152, Firefox 51–154, Safari 11, Opera 37, Edge 151).
- The failures are 45 cookie-sync *consent* tests on legacy **Edge 15 / EdgeHTML (2017)** plus two BrowserStack tunnel disconnects ("transport error").
- Master's most recent completed run (May 26, run 26452312354, pre-dating this change and #1246) fails the **identical Edge 15 consent tests — 47 of them**; all three April runs of the base branch (including the pre-feature refactor commit) also concluded failure.
- The Karma cookie-sync suite has no module-41/`google_hm` coverage, so this change cannot alter its behavior.


crisryantan:
Two checks are expected to stay red on this PR and are safe to ignore:

- **"Confirm that target branch for PR is main or build/"** — fails by design: this PR intentionally targets `worktree-gmp-cookie-sync-b64` (the head branch of #1246) so the changes fold into that PR. The mainline-target policy applies to #1246 itself.
- **BrowserStack Test** — pre-existing/chronic (details and evidence in the PR description; identical failures on master's May 26 run and on all April runs of the base branch).

Everything else is green: Jest on Node 18–25, Core (Karma) tests, SonarCloud quality gate (A maintainability after cc8d7e3), Semgrep, semantic title/branch-name, bundle-size report.

src/utils.ts
```diff
@@ -236,6 +236,32 @@ const createCookieSyncUrl = (
...
+    const unpaddedLength = base64.length - paddingLength;
+    const webSafeChars: string[] = [];
+
+    for (let i = 0; i < unpaddedLength; i++) {
+        const char = base64.charAt(i);
+        webSafeChars.push(WEB_SAFE_BASE64_REPLACEMENTS[char] || char);
+    }
+
+    return webSafeChars.join('');
+};
```
rmi22186: replaceAll may be an issue, but replace isn't.  this might be simpler and doesn't deal with padding.

```
btoa(value).replace(/[+/=]/g, (c) =>
    c === '+' ? '-' : c === '/' ? '_' : ''
);
```

test/jest/cookieSyncManager.spec.ts
```diff
@@ -861,6 +865,117 @@ describe('CookieSyncManager', () => {
...
+                    'https://cm.g.doubleclick.net/pixel?google_nid=abc123',
+                ],
+                [
+                    "is 'False'",
+                    { enableHmTag: 'False' },
+                    'https://cm.g.doubleclick.net/pixel?google_nid=abc123',
+                ],
+                [
+                    "is 'true' (case-insensitive)",
+                    { enableHmTag: 'true' },
```
rmi22186, line 887: server sends back capitalized strings

```suggestion
                    { enableHmTag: 'True' },
```

src/utils.ts
crisryantan: ah i actually did replace at first but the sonarcloud suggestion was to use replaceAll for performance purposes 🤔

crisryantan: Adopted in 2cfdcfe8 — agreed, simpler and handles padding in the same pass. (Sonar's replaceAll hint is MINOR-only and doesn't fail the gate; replaceAll isn't in our ES5 lib anyway.)

test/jest/cookieSyncManager.spec.ts
crisryantan: Added a 'True' (server contract) row in 2cfdcfe8, and kept the lowercase row since it pins the case-insensitive gate behavior.

src/utils.ts
```diff
@@ -240,10 +240,14 @@ const createCookieSyncUrl = (
...
+    '+': '-',
+    '/': '_',
+    '=': '',
+};
+
 const toWebSafeBase64 = (value: string): string =>
-    btoa(value).replace(/[+/=]/g, c =>
-        c === '+' ? '-' : c === '/' ? '_' : ''
-    );
+    btoa(value).replace(/[+/=]/g, c => WEB_SAFE_BASE64_REPLACEMENTS[c]);
```
rmi22186, line 250: love it

rmi22186 (approved):

PR #1320, opened 2026-08-12
fix: stabilize BrowserStack cross-browser suite on legacy Edge (EdgeHTML 15)

crisryantan:
## Problem

The **BrowserStack Test** workflow has failed on every run since late 2025 (last green: run 18953767668, Oct 30 2025), with three failure modes:

1. **Edge 15.15063 (EdgeHTML, Windows 10) deterministically fails 47 of ~1050 tests** — identical failure set on `master` run 26452312354 (May 26) and PR run 31624825318 (Aug 12). Every other browser in the matrix (Chrome 50, Firefox 51, Safari 11.1, Opera 37) passes.
2. **Occasional 6-hour hangs** until GitHub's default job timeout cancels the run (runs 26465972950, 25018891556, 23265634385).
3. **BrowserStack tunnel flake** — "Client disconnected from CONNECTED state (transport error)" marking otherwise-passing browsers DISCONNECTED.

## Root causes (from run logs + git archaeology)

**42 of the 47 Edge failures are a single cascade.** The test `Set width and height to 0 when window is defined but screen is not defined` (`tests-runtimeToBatchEventsDTO.ts`) does `delete window.screen`. On EdgeHTML, `window.screen` is a non-configurable host object: the delete corrupts it, the test throws (`TypeError: Object expected` inside `convertEvents`) *before* the line that restores `window.screen`, and every batch upload for the remainder of the run throws inside promise chains when `convertEvents` reads `window.screen.width`. The result is that every upload-asserting suite imported *after* it in `_test.index.ts` fails (`tests-batchUploader_3`, `tests-legacy-alias-requests`, `tests-integration-capture`, `tests-batchUploader_4`, `tests-identity` — e.g. `expected 'https://identity.mparticle.com/v1/identify' to be 'https://jssdks.mparticle.com/v3/JS/test_key/events'`, `expected 1 to equal 3`, `Unable to get property 'requestBody' of undefined`), while identical suites imported before it pass.

**The other 5 are the cookie-sync consent tests.** The shared fixture (`pixelUrl: 'https://test.com'`, `redirectUrl: '?redirect=...'`) makes `createCookieSyncUrl` produce `https://test.com%3Fredirect%3D...` — percent-encoding in the *host* component. EdgeHTML's strict URL parser throws a synchronous `Error: Invalid argument.` when this is assigned to `img.src` in `performCookieSync`. The 5 failing tests call `user.setConsentState(...)` directly, so the throw propagates into the test (the other cookie-sync tests trigger syncs inside SDK callbacks that swallow it). This is also a real (if unlikely) production hazard: pixel URLs come from server-side configuration, and on EdgeHTML one malformed pixel URL would make public API calls (`identify`, `setConsentState`) throw into customer code and abort the remaining pixels.

**Why it "broke" without a code change:** the failing tests and SDK code are byte-identical to the last green run (the screen test dates to #936). The karma matrix has requested `edge 15.0` unchanged since #798, but run logs show BrowserStack substituting browsers when an image is unavailable (e.g. modern Chrome/Edge "ghost" sessions alongside the requested ones). The evidence says the Oct 2025 greens ran against a substituted browser, and BrowserStack later restored a genuine EdgeHTML 15 image — the tests were never actually compatible with real EdgeHTML.

## Changes

- **`src/cookieSyncManager.ts`** — wrap `performCookieSync` in try/catch and log via `Logger.error`. A malformed pixel URL (server-provided config = external boundary) can no longer throw through public API calls or block the remaining pixels. Jest coverage added for the contained-error path (no throw, error logged, no sync date recorded).
- **`test/src/tests-runtimeToBatchEventsDTO.ts`** — skip the `window.screen` deletion test on EdgeHTML/IE (UA guard *before* the delete), and restore `window.screen` in a `finally` so an assertion failure can never poison later suites. Tradeoff stated below.
- **`.github/workflows/cross-browser-testing.yml`** — `timeout-minutes: 30` (healthy runs take <15 min; kills the 6h hangs) and a `concurrency` group cancelling superseded runs per ref (reduces BrowserStack parallel-session pressure that causes queueing/tunnel flake).
- **`test/cross-browser-testing/browserstack.karma.config.js`** — `captureTimeout: 300000` (BrowserStack session queueing regularly exceeds the 60s default) and `browserNoActivityTimeout: 120000` (mid-run stalls on slow VMs are not disconnects). Existing `browserDisconnectTimeout`/`browserDisconnectTolerance` retained.

## Coverage tradeoff

One test (`Set width and height to 0 when window is defined but screen is not defined`) is now skipped on EdgeHTML/IE only. It exercises a pure JS branch of `convertEvents` (not browser behavior) by deleting a host object, which EdgeHTML fundamentally does not support; it continues to run on every other browser in the BrowserStack matrix and in the local karma/Chrome suites. No consent or cookie-sync coverage is skipped anywhere — those tests now genuinely pass on EdgeHTML because the SDK no longer throws.

## Verification

- Local: `npm run lint` clean; Jest 589/589; `build:iife` + `build:cbt` compile; karma ChromeHeadless 1069/1069 SUCCESS.
- CI: the push to this branch triggers the BrowserStack workflow; all matrix browsers should finish green well within the new 30-minute cap. Will follow up with the run result.

### DIFF
--- a/.github/workflows/cross-browser-testing.yml
+++ b/.github/workflows/cross-browser-testing.yml
@@ -4,10 +4,21 @@
 name: 'BrowserStack Test'
 on: [push, pull_request, workflow_dispatch]
 
+# Cancel superseded runs for the same ref to reduce pressure on the limited
+# pool of BrowserStack parallel sessions, which causes queueing, tunnel
+# disconnects, and browser capture timeouts.
+concurrency:
+  group: browserstack-test-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   browserstack-test:
     name: 'BrowserStack Test'
     runs-on: ubuntu-latest
+    # A healthy run takes < 15 minutes. Without a timeout, karma occasionally
+    # waits on half-open BrowserStack tunnel sockets until GitHub's 6 hour
+    # default cancels the job.
+    timeout-minutes: 30
     steps:
 
       - name: 'BrowserStack Env Setup'  # Invokes the setup-env action
--- a/src/cookieSyncManager.ts
+++ b/src/cookieSyncManager.ts
@@ -156,18 +156,32 @@ export default function CookieSyncManager(
         mpid: MPID,
         cookieSyncDates: CookieSyncDates,
     ): void => {
-        const img = document.createElement('img');
+        try {
+            const img = document.createElement('img');
 
-        mpInstance.Logger.verbose(InformationMessages.CookieSync);
-        img.onload = function() {
-            cookieSyncDates[moduleId] = new Date().getTime();
+            mpInstance.Logger.verbose(InformationMessages.CookieSync);
+            img.onload = function() {
+                cookieSyncDates[moduleId] = new Date().getTime();
 
-            mpInstance._Persistence.saveUserCookieSyncDatesToPersistence(
-                mpid,
-                cookieSyncDates
+                mpInstance._Persistence.saveUserCookieSyncDatesToPersistence(
+                    mpid,
+                    cookieSyncDates
+                );
+            };
+            img.src = url;
+        } catch (error) {
+            // Pixel URLs come from server-side configuration, and some browsers
+            // (e.g. legacy EdgeHTML) throw synchronously when an invalid URL is
+            // assigned to img.src. Contain the error so a single bad pixel cannot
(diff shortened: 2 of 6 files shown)


crisryantan:
will revive later

PR #1324, opened 2026-08-17
feat(rokt): send the current mParticle session id as a selectPlacements attribute

jamesnrokt:
## Summary

Applies the same change as mparticle-integrations/mparticle-javascript-integration-rokt#116 to the in-repo copy of the Rokt kit, so the two do not drift while the monorepo migration is in progress.

The kit read the mParticle session id once, when creating the launcher, and passed it through as the `mpSessionId` launcher option. That option is no longer used, and reading the value at launcher-creation time meant it went stale for the life of the page: mParticle mints a new session id on inactivity timeout (`sessionTimeout`, 30 minutes by default), when another tab starts a new session, and after an explicit `endSession()` / `startNewSession()`. None of that reached the launcher.

This PR:

- Sends the session id as the **`mparticle_session_id` attribute** on each `selectPlacements` call. Because it is read per call, a rotated session is reflected on the next call.
- Removes the `mpSessionId` launcher option from `attachLauncher`.
- Adds `readMpSessionId()`, which reads `SessionManager.getSessionId()` if the facade exposes it and otherwise falls back to `getSession()`.

Review follow-ups applied:

- Reinstated the `getSession()` fallback. An earlier revision dropped it on the assumption that `getSessionId()` was available from core `v2.23.3`; that is true of the internal `_SessionManager` but **not** of the public `mParticle.sessionManager` facade, which exposes only `getSession()` in every published core through `2.78.0`:

  ```js
  // mparticle-instance-manager.ts
  this.sessionManager = {
      getSession: function() {
          return self.getInstance()._SessionManager.getSessionId();
      },
  };
  ```

  Dropping the fallback made the attribute silently absent — `isFunction(undefined)` is `false`, so `readMpSessionId()` always returned `undefined`. Confirmed in production against the standalone kit `1.33.0` on core `2.78.0`, where the attribute was absent from all 79,987 `wsdk` events in a sampled window. The public `getSession()` is not deprecated and emits no warning: it calls `_SessionManager.getSessionId()` directly, bypassing the deprecated internal `_SessionManager.getSession()`. Only the internal method carries `@deprecated` and `logDeprecatedMethodUsage`.

- Widened the `sessionManager` interface back to accept both accessors. It had been narrowed to an optional `getSessionId?()`, so TypeScript could not catch the absence.
- Added an `isFunction` helper and used it for all five `typeof x === 'function'` checks in the kit rather than leaving the new one inconsistent with the rest.

## Testing Plan

- `kits/rokt`: 237/237 tests pass; `vite build` clean.
- New unit tests cover: the session id current at the time of each call (asserted across a rotation between two calls), the attribute being omitted when `sessionManager` is unavailable, a facade exposing only `getSession` (the real shape) yielding the attribute, `getSessionId` being preferred when both are present, the attribute being omitted when neither accessor exists, and `createLauncher` no longer receiving `mpSessionId`.
- 35 existing full-payload `toEqual` assertions were updated, since the attribute is present on every call. The `#selectPlacements` `beforeEach` now pins a known `sessionManager`; previously these tests depended on an earlier test deleting `window.mParticle.sessionManager` without restoring it, which made results order-dependent.
- Test fixtures in `test/vitest.setup.ts` and `tests.spec.ts` use `getSession`, matching the real public facade. An earlier revision moved them to `getSessionId`, which is why the suite went green against a facade that does not exist.
- `kits/rokt/dist` is intentionally not included. Note this copy has no `src/utils.ts`, so `isFunction` sits with the existing module-level helpers rather than in a utils module.
- `kits/rokt` has 7,771 pre-existing `prettier/prettier` lint errors on a pristine tree (an indentation-width mismatch across all four files). This PR does not change that count materially and does not attempt to reformat the kit.
- Additional testing worth doing: confirm on a long-lived page that a session which times out is reflected on the next `selectPlacements` call, and that the offers request body carries `attributes.mparticle_session_id`.

- Note for reviewers running tests locally: `kits/rokt` uses vitest 4, which needs Node 22. On Node 20 it fails with `bad option: --no-experimental-webstorage`.


thomson-t (approved):

crisryantan:
Closing: this change shipped upstream in the standalone kit repo as mparticle-integrations/mparticle-javascript-integration-rokt#116 (released in kit 1.33.0) and was then corrected upstream by #117, which reads the session id from the public sessionManager facade instead. Both are now in main via the v1.35.0 sync (#1365), so merging this PR would reapply the uncorrected version.

PR #1326, opened 2026-08-18
fix: resolve TypeScript build errors in type declarations

alexs-mparticle:
## Summary

- Fix `generateHash` interface return type (`string` → `number`) to match the implementation in `utils.ts`
- Rename `request_timestamp_ms` → `request_timestamp_unixtime_ms` in `identity.js` to satisfy `IIdentityAPIRequestData`
- Fix `convertProductActionToEventType`/`convertPromotionActionToEventType` return types: `typeof CommerceEventType` (namespace object) → `valueof<typeof CommerceEventType>` (a member value)
- Fix `createIdentityChanges` return type in `IIdentityRequest`: single object → `IIdentityAPIIdentityChangeData[]` (it actually returns an array)
- Add JSDoc return types to `persistence.getLocalStorage`, `helpers.sanitizeAttributes`, and `identity.mParticleUser`/`mParticleUserCart` to unblock TS9005 declaration emit
- Cast `window.mParticle as any` when accessing internal pre-init properties (`config`, `isIOS`, `Store`, `_BatchValidator`) not on the public `@types/mparticle__web-sdk` type
- Widen `IConsentRulesValues.consentPurpose` to `string | number` — tests use `generateHash()` (returns `number`); JS auto-coerces numbers to strings as object keys so both work at runtime

## Test Plan

- [x] `npm run build` — clean, zero errors
- [x] `npm run build:types` — clean, zero TypeScript errors
- [x] `npm run test:jest` — 540 tests passing across 23 suites

### DIFF
--- a/src/identity.interfaces.ts
+++ b/src/identity.interfaces.ts
@@ -2,7 +2,7 @@ import { IdentityApiData, MPID, UserIdentities } from '@mparticle/web-sdk';
 import AudienceManager from './audienceManager';
 import { ICachedIdentityCall, IKnownIdentities } from './identity-utils';
 import { BaseVault } from './vault';
-import { Dictionary, Environment, valueof } from './utils';
+import { Dictionary, valueof } from './utils';
 import Constants from './constants';
 import {
     IdentityCallback,
@@ -55,7 +55,7 @@ export interface IIdentityAPIRequestData {
     context: string | null;
     environment: string;
     request_id: string;
-    request_timestamp_unixtime_ms: number;
+    request_timestamp_ms: number;
     previous_mpid: MPID | null;
     known_identities: IKnownIdentities;
 }
--- a/src/identity.interfaces.ts
+++ b/src/identity.interfaces.ts
@@ -53,7 +53,7 @@ export interface IIdentityAPIRequestData {
         sdk_version: typeof sdkVersion;
     };
     context: string | null;
-    environment: Environment;
+    environment: string;
     request_id: string;
     request_timestamp_unixtime_ms: number;
     previous_mpid: MPID | null;
@@ -69,7 +69,7 @@ export interface IIdentityAPIModifyRequestData
 }
 
 export interface IIdentityAPIIdentityChangeData {
-    identity_type: SDKIdentityTypeEnum;
+    identity_type: string;
     old_value: string;
     new_value: string;
 }
@@ -98,8 +98,8 @@ export interface IIdentityRequest {
     ): IIdentityAPIModifyRequestData;
     createIdentityChanges(
         previousIdentities: UserIdentities,
-        newIdentitie: UserIdentities
-    ): IIdentityAPIIdentityChangeData;
+        newIdentities: UserIdentities
+    ): IIdentityAPIIdentityChangeData[];
     preProcessIdentityRequest(
         identityApiData: IdentityApiData,
         callback: IdentityCallback,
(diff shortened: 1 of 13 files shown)


src/identity.js
```diff
@@ -94,7 +94,7 @@ export default function Identity(mpInstance) {
                     ? 'development'
                     : 'production',
                 request_id: mpInstance._Helpers.generateUniqueId(),
-                request_timestamp_ms: new Date().getTime(),
+                request_timestamp_unixtime_ms: new Date().getTime(),
```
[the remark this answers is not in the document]
rmi22186: revert

src/identity.interfaces.ts
rmi22186: Remove Environment, like the bot says, not that you're not using it

rmi22186: Although is there a reason to not use it?

src/identity.js
```diff
@@ -137,7 +124,7 @@ export default function Identity(mpInstance) {
...
```
rmi22186: ditto

src/mp-instance.ts
```diff
@@ -209,9 +198,10 @@ export default function mParticleInstance(this: IMParticleWebSDKInstance, instan
     this.getDeviceId = this._Persistence.getDeviceId;
 
     if (typeof window !== 'undefined') {
-        if (window.mParticle && window.mParticle.config) {
-            if (window.mParticle.config.hasOwnProperty('rq')) {
-                this._preInit.readyQueue = window.mParticle.config.rq;
+        const winMp = window.mParticle as any;
```
rmi22186: Shouldn't we not be using `any` if possible? why the reason for this change?

src/mparticle-instance-manager.ts
```diff
@@ -36,19 +36,13 @@ function mParticleInstanceManager(this: IMParticleInstanceManager) {
...
     if (typeof window !== 'undefined') {
-        this.isIOS =
-            window.mParticle && window.mParticle.isIOS
-                ? window.mParticle.isIOS
-                : false;
-        this.config =
-            window.mParticle && window.mParticle.config
-                ? window.mParticle.config
-                : {};
+        const winMp = window.mParticle as any;
```
rmi22186: ditto

src/mparticle-instance-manager.ts
```diff
@@ -61,11 +55,12 @@ function mParticleInstanceManager(this: IMParticleInstanceManager) {
      * @param {String} [instanceName] If you are self hosting the JS SDK and working with multiple instances, you would pass an instanceName to `init`. This instance will be selected when invoking other methods. See the above link to the doc site for more info and examples.
      */
     this.init = function(apiKey, config, instanceName) {
-        if (!config && (window.mParticle && window.mParticle.config)) {
+        const winMp = window.mParticle as any;
```
rmi22186: tritto

src/store.ts
```diff
@@ -358,10 +354,8 @@ export default function Store(
         if (config.hasOwnProperty('isIOS')) {
             this.SDKConfig.isIOS = config.isIOS;
         } else {
-            this.SDKConfig.isIOS =
-                window.mParticle && window.mParticle.isIOS
-                    ? window.mParticle.isIOS
-                    : false;
+            const winMp = window.mParticle as any;
```
rmi22186: 4itto

alexs-mparticle:
Closing as this work will be done as part of the other typescript migration work

PR #1328, opened 2026-08-18
feat(rokt): expose terminate on the window.mParticle.Rokt interface

jamesnrokt:
## Summary

Adds `RoktManager.terminate()`, exposing `window.mParticle.Rokt.terminate()` as a supported teardown path for partners who cannot simply destroy the container element — SPA route changes and confirmation pages in particular.

Today partners reach for the undocumented `window.Rokt.currentLauncher?.terminate()`. This routes the same call through the documented interface.

Pairs with mparticle-integrations/mparticle-javascript-integration-rokt#122, which adds the `terminate()` the manager delegates to. **That PR should merge and release first** — until a kit carrying it is on the page, this method logs that the attached kit version does not support terminate.

## Changes

- `RoktManager.terminate(): Promise<void>` — delegates to `this.kit.terminate()`, which forwards to the Rokt launcher.
- `IRoktKit.terminate` added as **optional**, guarded at the call site with `isFunction`. The Rokt Kit releases on its own cadence, so a kit published before this change genuinely will not implement it — version skew is real rather than theoretical.

`IRoktLauncher` is untouched: the manager only ever calls through the kit, and the kit owns the launcher-facing type.

## Behaviour

| State | Result |
|---|---|
| Kit + launcher attached | `await`s `kit.terminate()` |
| Nothing attached | Resolves as a no-op, logs verbose |
| Kit attached, no launcher | Resolves as a no-op |
| Kit predates `terminate` | Resolves, logs error |
| `kit.terminate()` rejects | Resolves, logs error |

`terminate()` never rejects — teardown should not be something callers have to defend against.

### A pre-ready call is not queued

Every other proxied method uses `deferredCall` when the kit isn't ready. `terminate` deliberately does not. Replaying a teardown against a launcher created *later* would silently tear down placements the caller never asked to remove; with no launcher there is nothing to terminate, so the call resolves as a no-op instead. This asymmetry is called out in the method's doc comment and pinned by tests.

## Also fixes a latent snippet bug

`terminate` is already in the snippet's `roktMethods` stub list (`snippet.js`), so a snippet-loaded page could queue `Rokt.terminate()` before init. At flush time `processPreloadedItem` hit an undefined method, threw, and `processReadyQueue` logged `Unable to compute proper mParticle function`. That path now dispatches correctly.

⚠️ **`getVersion` has the same latent gap** — it is in the snippet stub list but has no `RoktManager` implementation. Left out of scope here; happy to add it in a follow-up if wanted.

## Not included

`kits/rokt/` in this repo is a stale vendored snapshot of the Rokt kit (missing `storage.ts`, `utils.ts`, `pageViewStorage.ts`) and is not loaded by any build or test target, so it was intentionally left untouched. The live kit change is in the companion PR.

## Testing

- `npm run lint` — pass
- `npm run test:jest` — 632/632 across 26 suites
- Karma (ChromeHeadless) — 1087/1087

7 new `#terminate` cases in `test/jest/roktManager.spec.ts` cover each row of the behaviour table above, plus a case asserting the manager stays ready so repeat calls keep reaching the kit. The suite fails to compile against `development` without the source change.

> Note for reviewers running locally: Karma's default config includes `FirefoxHeadless` and aborts the whole run if Firefox isn't installed. Use `npx karma start test/karma.config.js --browsers ChromeHeadless`.

### DIFF
--- a/src/roktManager.ts
+++ b/src/roktManager.ts
@@ -83,6 +83,9 @@ export interface IRoktKit {
     setExtensionData<T>(extensionData: IRoktPartnerExtensionData<T>): void;
     use: <T>(name: string) => Promise<T>;
     onShoppableAdsReady(callback: () => void): void;
+    // Optional because the Rokt Kit ships on its own release cadence; a kit
+    // published before terminate() existed will not implement it.
+    terminate?: () => Promise<void>;
     launcherOptions?: Dictionary<any>;
     settings?: IRoktKitSettings;
     integrationName?: string;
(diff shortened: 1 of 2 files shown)


kits/adobe-target/dist/AdobeTarget-Kit.iife.js
```diff
@@ -0,0 +1,829 @@
+console.info('[mParticle QA kit path smoke] Loaded kits/adobe-target/dist/AdobeTarget-Kit.iife.js from monorepo GitHub path');
```
crisryantan: should this be gitignored?

crisryantan: seems generated from build including kits/adobe-target/dist/AdobeTarget-Kit.common.js

jamesnrokt: I targeted the wrong base when branching. Have corrected the base so only my changes are now shown

rmi22186: @jamesnrokt / @crisryantan  - Thanks for catching this. Note that this is actually a commit I had added in `main` in order to test that updated QA paths are now working in the monorepo.

rmi22186 (approved):
This LGTM.  @mattbodle's comment over at https://github.com/mparticle-integrations/mparticle-javascript-integration-rokt/pull/122 _shouldn't_ affect the implementation here, though I'm not 100% clear on the use case of creating a new launcher right after terminating.  That seems unlkely?


Edit: Oh I see, for SPAs, so a customer may go back and make another purchase, right?

PR #1331, opened 2026-08-18
fix(apv): prevent duplicate page views on SPA re-init (stacked wrapper + init guard)

alexs-mparticle:
## Summary

- Adds a `window.__mpApvTracker__` singleton so that when Next.js re-evaluates the SDK module on each SPA navigation, the new module's tracker tears down the previous module's tracker before patching `pushState`. This eliminates stacked wrappers (N page views per navigation).
- Guards `logPageView()` in `completeSDKInitialization` behind `window.__mpApvInitPVLogged__` so the initial page view fires only once per hard page load, not on every `mParticle.init()` call.
- Fixes `_resetForTests` to tear down any active `PageViewTracker` and clear both APV window flags between tests, preventing stale state from leaking across test cases.

## Test plan

- [ ] Verify a single page view is logged on hard load with `AutoLogPageView: True`
- [ ] Verify SPA navigation (pushState) logs exactly one page view per route change
- [ ] Verify calling `mParticle.init()` again on re-navigation does not fire a duplicate initial page view
- [ ] `npm test` passes (2176 tests, 0 failures)

## Notes

The companion PR for kit double-registration via `configuredForwarders` accumulation is being reviewed separately as a draft.

### DIFF
--- a/src/mp-instance.ts
+++ b/src/mp-instance.ts
@@ -57,13 +57,7 @@ import { LoggingDispatcher } from './reporting/loggingDispatcher';
 import { IErrorReportingService, ILoggingService } from './reporting/types';
 import { logDeprecatedMethodUsage } from './reporting/deprecatedMethodLogger';
 import { normalizeRoktLauncherOptions } from './roktLauncherOptions';
-import { PageViewTracker, WIN_TRACKER_KEY } from './pageViewTracker';
-
-const WIN_INIT_PV_KEY = '__mpApvInitPVLogged__' as const;
-type WindowWithApvFlags = Window & {
-    [WIN_INIT_PV_KEY]?: boolean;
-    [WIN_TRACKER_KEY]?: PageViewTracker;
-};
+import { PageViewTracker, WIN_TRACKER_KEY, WIN_INIT_PV_KEY, WindowWithApvFlags } from './pageViewTracker';
 
 export interface IErrorLogMessage {
     message?: string;
--- a/src/mp-instance.ts
+++ b/src/mp-instance.ts
@@ -304,11 +304,19 @@ export default function mParticleInstance(this: IMParticleWebSDKInstance, instan
         );
         instance._Events.stopTracking();
 
-        if (instance._PageViewTracker) {
-            instance._PageViewTracker.teardown();
+        const instanceTracker = instance._PageViewTracker;
+        if (instanceTracker) {
+            instanceTracker.teardown();
             instance._PageViewTracker = undefined;
         }
         const win = window as WindowWithApvFlags;
+        // Tear down any window-registered tracker that differs from the instance
+        // tracker (e.g. left by a previous module load) before clearing the
+        // window flag, so its history wrappers and popstate listener are removed.
+        const windowTracker = win[WIN_TRACKER_KEY];
+        if (windowTracker && windowTracker !== instanceTracker) {
+            windowTracker.teardown();
+        }
         delete win[WIN_INIT_PV_KEY];
         delete win[WIN_TRACKER_KEY];
 
@@ -1600,6 +1608,9 @@ function completeSDKInitialization(apiKey, config, mpInstance) {
         mpInstance._SessionManager.initialize();
         mpInstance._Events.logAST();
 
+        // Note: APV uses window-level singletons (WIN_TRACKER_KEY, WIN_INIT_PV_KEY)
+        // that assume a single active SDK instance. Multiple-instance support is
+        // out of scope for this Rokt/legacy APV feature.
         if (getFeatureFlag(AutoLogPageView)) {
             if (!mpInstance._PageViewTracker) {
                 mpInstance.Logger.verbose(
@@ -1627,6 +1638,8 @@ function completeSDKInitialization(apiKey, config, mpInstance) {
             }
             mpInstance._PageViewTracker.init();
         } else if (mpInstance._PageViewTracker) {
+            // AutoLogPageView was disabled on re-init: tear down the active tracker
+            // so it stops listening. Re-created if the flag is re-enabled later.
             mpInstance._PageViewTracker.teardown();
         }
 
--- a/src/mp-instance.ts
+++ b/src/mp-instance.ts
@@ -1595,7 +1595,7 @@ function completeSDKInitialization(apiKey, config, mpInstance) {
 
         // Note: APV uses window-level singletons (WIN_TRACKER_KEY, WIN_INIT_PV_KEY)
         // that assume a single active SDK instance. Multiple-instance support is
-        // out of scope for this Rokt/legacy APV feature.
+        // out of scope
         if (getFeatureFlag(AutoLogPageView)) {
             if (!mpInstance._PageViewTracker) {
                 mpInstance.Logger.verbose(
--- a/src/mp-instance.ts
+++ b/src/mp-instance.ts
@@ -57,7 +57,13 @@ import { LoggingDispatcher } from './reporting/loggingDispatcher';
 import { IErrorReportingService, ILoggingService } from './reporting/types';
 import { logDeprecatedMethodUsage } from './reporting/deprecatedMethodLogger';
 import { normalizeRoktLauncherOptions } from './roktLauncherOptions';
-import { PageViewTracker } from './pageViewTracker';
+import { PageViewTracker, WIN_TRACKER_KEY } from './pageViewTracker';
+
+const WIN_INIT_PV_KEY = '__mpApvInitPVLogged__' as const;
+type WindowWithApvFlags = Window & {
+    [WIN_INIT_PV_KEY]?: boolean;
+    [WIN_TRACKER_KEY]?: PageViewTracker;
+};
 
 export interface IErrorLogMessage {
     message?: string;
@@ -297,6 +303,15 @@ export default function mParticleInstance(this: IMParticleWebSDKInstance, instan
             window.localStorage
         );
         instance._Events.stopTracking();
+
+        if (instance._PageViewTracker) {
+            instance._PageViewTracker.teardown();
+            instance._PageViewTracker = undefined;
+        }
+        const win = window as WindowWithApvFlags;
+        delete win[WIN_INIT_PV_KEY];
+        delete win[WIN_TRACKER_KEY];
+
         if (!keepPersistence) {
             instance._Persistence.resetPersistence();
         }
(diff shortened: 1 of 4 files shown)


src/mp-instance.ts
```diff
@@ -1586,10 +1601,29 @@ function completeSDKInitialization(apiKey, config, mpInstance) {
...
+                // fire duplicate logPageView() events for the same page.
+                const win = window as WindowWithApvFlags;
+                if (win[WIN_INIT_PV_KEY]) {
+                    mpInstance.Logger.verbose(
+                        'mParticle APV: [sdk-init] initial page view already logged this page load, skipping'
+                    );
+                } else {
+                    win[WIN_INIT_PV_KEY] = true;
+                    mpInstance._Events.logPageView();
+                }
```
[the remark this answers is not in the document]
rmi22186: @alexs-mparticle - Can we note somewhere that this doesn't support multiple instances, and we decided we are okay with that?

src/mp-instance.ts
```diff
@@ -1586,10 +1601,29 @@ function completeSDKInitialization(apiKey, config, mpInstance) {
...
+            } else {
+                mpInstance.Logger.verbose(
+                    'mParticle APV: [sdk-init] tracker already on this instance, skipping logPageView'
+                );
```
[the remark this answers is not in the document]
rmi22186: @alexs-mparticle is this a concern? If the PageViewTracker is watching navigation, I'm not sure that it is, but would be good to comment back as to why it isn't if not.  (I'm not saying it's not)

src/pageViewTracker.ts
```diff
@@ -37,16 +49,26 @@ export class PageViewTracker {
...
         }
 
+        // Tear down any tracker left over from a previous module load.
+        // window[WIN_TRACKER_KEY] outlives module re-evaluation; this is the
+        // only way to reach a tracker instance in a dead module scope.
+        const win = window as WindowWithTracker;
+        const prev = win[WIN_TRACKER_KEY];
+        if (prev && prev !== this) {
+            this.mpInstance.Logger.verbose(
+                'mParticle APV: [init] found stale tracker from previous module load — tearing down to prevent stacked wrappers'
```
rmi22186: Could this handoff preserve or flush a page view already queued by `prev`? A route change queues its event with `setTimeout`; if module re-evaluation reaches this synchronous init before that timer runs, `teardown()` marks the previous tracker inactive and its callback aborts. The replacement tracker seeds the destination as `lastPath`, so that navigation is never logged. A regression test where the old tracker detects a route, the replacement initializes, and then timers flush would cover the target scenario.

src/mp-instance.ts
```diff
@@ -297,6 +303,15 @@ export default function mParticleInstance(this: IMParticleWebSDKInstance, instan
...
         );
         instance._Events.stopTracking();
+
+        if (instance._PageViewTracker) {
+            instance._PageViewTracker.teardown();
+            instance._PageViewTracker = undefined;
+        }
+        const win = window as WindowWithApvFlags;
+        delete win[WIN_INIT_PV_KEY];
+        delete win[WIN_TRACKER_KEY];
```
rmi22186: This can delete the only reference to an active stale tracker without tearing it down. After module re-evaluation, `instance._PageViewTracker` may be absent or differ from `win[WIN_TRACKER_KEY]`; in that case its history wrapper/listener remains installed but becomes undiscoverable. Please teardown the tracker referenced by the window registry before clearing it (while avoiding a second teardown when both references are the same).

test/src/tests-event-logging.ts
```diff
@@ -611,6 +611,37 @@ describe('event logging', function() {
                 window.document.title
             );
         });
+
+        it('should not log a duplicate page view when init() is called again (SPA re-init)', async () => {
```
rmi22186, line 630: This verifies repeated `init()` on the same SDK instance, where `_PageViewTracker` already exists. It does not exercise the new window-level handoff or `WIN_INIT_PV_KEY` path used when a fresh module creates a fresh tracker. Could we add a focused module-replacement test with an old tracker stored on `window`, including a pending route event and cleanup of stale global state?

rmi22186 (commented):
Re-reviewed with multiple-instance support explicitly treated as out of scope for this Rokt/legacy APV feature. I found two lifecycle edge cases and a test-coverage gap. One additional question: `development` already marks wrappers with `__mpApvWrapped__` and skips re-wrapping, so it would be useful to confirm whether the observed production duplication is from accumulated listeners/stale ownership rather than stacked `pushState` wrappers.

src/mp-instance.ts
```diff
@@ -1585,14 +1593,36 @@ function completeSDKInitialization(apiKey, config, mpInstance) {
         mpInstance._SessionManager.initialize();
         mpInstance._Events.logAST();
 
+        // Note: APV uses window-level singletons (WIN_TRACKER_KEY, WIN_INIT_PV_KEY)
+        // that assume a single active SDK instance. Multiple-instance support is
+        // out of scope for this Rokt/legacy APV feature.
```
alexs-mparticle:
```suggestion
        // out of scope
```

src/pageViewTracker.ts
```diff
@@ -0,0 +1,307 @@
...
+    private replaceStateWrapper: HistoryStateMethod | null = null;
+
+    private popStateListener: (() => void) | null = null;
+
+    constructor(mpInstance: IMParticleWebSDKInstance) {
+        this.mpInstance = mpInstance;
+        const win = window as WindowWithTracker;
+        const count = (win[WIN_TRACKER_COUNTER_KEY] || 0) + 1;
+        win[WIN_TRACKER_COUNTER_KEY] = count;
+        this._id = `#${count}`;
```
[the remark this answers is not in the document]
alexs-mparticle: Fixed in an earlier commit — `WIN_TRACKER_COUNTER_KEY`, `_id`, and all counter logic were removed from `PageViewTracker`. No `__mpApvTrackerCount__` writes to `window` in the current code.

src/pageViewTracker.ts
```diff
@@ -0,0 +1,307 @@
...
+        this.mpInstance._Events.logEvent({
+            messageType: MessageType.PageView,
+            name: 'PageView',
+            data: {
+                hostname: window.location.hostname,
+                title,
+                path,
+            },
+            eventType: EventType.Unknown,
+        });
```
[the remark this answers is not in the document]
alexs-mparticle: False positive. `PageUrl` is stamped synchronously by `serverModel.ts` (via `getHref()`) at the exact moment `logEvent` is called — there is no deferred serialisation. `firePageView` runs inside `setTimeout(0)`, so each flush fires in its own macrotask. By that point `window.location` reflects the settled URL, same as `title` (which we deliberately read late so the router can update it). The correct per-navigation path is captured in `data.path`. The `PageUrl` behaviour for rapid same-tick navigations is a pre-existing constraint of any deferred-fire strategy, not a regression from this PR.

src/mp-instance.ts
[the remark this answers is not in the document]
alexs-mparticle: False positive for the same-instance re-init case. When `_PageViewTracker` already exists and `init()` is called again, `teardown()` sets `isActive = false` but `init()` immediately sets `isActive = true` before returning. The deferred `firePageView` callback runs after `init()` completes, sees `isActive = true`, and fires normally. The one genuine edge case (a navigation queued by a *stale module-load* tracker being dropped on replacement) is a known limitation documented in code comments and covered by the regression test `should drop a pending page view from the stale tracker on module re-evaluation`.

src/mp-instance.ts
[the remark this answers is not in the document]
alexs-mparticle: Acknowledged and documented. Multiple-instance support is explicitly out of scope for this Rokt/legacy APV feature — the window singletons (`__mpApvTracker__`, `__mpApvInitPVLogged__`) assume a single active SDK instance. This is noted in a code comment added to `completeSDKInitialization`.

src/pageViewTracker.ts
alexs-mparticle: Addressed in this test: https://github.com/mParticle/mparticle-web-sdk/pull/1331/changes#diff-95766f37b2456e31ff6abdb9ed6c2aa543ecdb870ed563dfa4bd63fb20cda6fcR554

test/src/tests-event-logging.ts
alexs-mparticle: Added this test: https://github.com/mParticle/mparticle-web-sdk/pull/1331/changes#diff-9312ed43259c225c470943ef298f0e49ae5cbe99dc998f4ebc98d9d7747b7192R646

src/pageViewTracker.ts
rmi22186: Thanks for adding the regression coverage. The new test currently documents the loss (`expect(logEvent).not.toHaveBeenCalled()`) rather than resolving it, and I do not think we can assume this is only a sub-ms edge: with synchronous SPA rendering and `requestConfig: false`, replacement can occur before the deferred timer on each navigation. Could we transfer pending paths during handoff? Concretely, have each tracker retain `{ path, timeoutId }` entries; expose `takePendingNavigations()` to cancel the stale tracker’s timers and return their paths; call it before `prev.teardown()`; then schedule those paths on the replacement after it becomes active. The test should then expect `/route-a` exactly once, followed by `/route-b` exactly once. This preserves deferred title capture while preventing both duplicates and dropped views.

src/mp-instance.ts
mattbodle, line 286: Why do we have a test helper in our code?
Is this stripped out during bundling?

alexs-mparticle: No it's not stripped out but it probably shouldn't be in our code any longer.

src/pageViewTracker.ts
```diff
@@ -9,12 +9,55 @@ type MarkedHistoryMethod = HistoryStateMethod & {
...
+    }
+
+    static markInitialPageViewFired(): void {
+        (window as WindowWithApvFlags)[WIN_INIT_PV_KEY] = true;
+    }
+
+    // Tears down any window-registered tracker that differs from instanceTracker
+    // (e.g. left by a previous module load), then clears both APV window flags.
+    // Call from _resetForTests after tearing down the instance-level tracker.
+    static resetWindowState(instanceTracker?: PageViewTracker): void {
```
crisryantan, line 41: what do you think about pulling the window access behind a couple of module-level helpers?
```

static _resetWindowStateForTests(): void {
    const windowTracker = getTracker();

    try {
        if (windowTracker) {
            windowTracker.teardown();
        }
    } finally {
        setInitialPageViewFired(false);
        setTracker(undefined);
    }
}

function getTracker(): PageViewTracker | undefined {
    if (typeof window === 'undefined') return undefined;
    return (window as WindowWithApvFlags)[WIN_TRACKER_KEY];
}

function setTracker(tracker: PageViewTracker | undefined): void {
    if (typeof window === 'undefined') return;
    const win = window as WindowWithApvFlags;
    if (tracker) {
        win[WIN_TRACKER_KEY] = tracker;
    } else {
        delete win[WIN_TRACKER_KEY];
    }
}
```

gets the typeof guard everywhere for free instead of only some call sites, and keeps the window access in one place.

src/mp-instance.ts
mattbodle: Ok let's proceed, but I'll raise a PR to clean this up

src/pageViewTracker.ts
alexs-mparticle: Done in 0f51670 — added module-level `getWindowTracker()` / `setWindowTracker()` functions at the bottom of the module. All four `window[WIN_TRACKER_KEY]` call sites (init, setWindowTracker registration, teardown, and resetWindowState) now go through the helpers, which pick up the `typeof window === 'undefined'` guard for free.

crisryantan (approved):

PR #1333, opened 2026-08-19
refactor: migrate forwardingStatsUploader module to TS

jaissica12:
## Background
- Continuing the effort to migrate remaining JavaScript modules in the mParticle Web SDK to TypeScript for improved type safety and developer experience. `forwardingStatsUploader.js` was one of the remaining JS modules.

## What Has Changed
- Converted `src/forwardingStatsUploader.js` to `src/forwardingStatsUploader.ts` with type annotations
- Added `IForwardingStatsUploader` interface and wired it in `src/mp-instance.ts`
- No intentional runtime behavior, public API, or functional changes

## Screenshots/Video
- N/A

## Checklist

- Closes https://go/j/SDKE-1194

### DIFF
--- a/src/forwardingStatsUploader.ts
+++ b/src/forwardingStatsUploader.ts
@@ -1,12 +1,25 @@
-export default function forwardingStatsUploader(mpInstance) {
-    this.startForwardingStatsTimer = function() {
-        mParticle._forwardingStatsTimer = setInterval(function() {
-            prepareAndSendForwardingStatsBatch();
-        }, mpInstance._Store.SDKConfig.forwarderStatsTimeout);
+import { IMParticleWebSDKInstance } from './mp-instance';
+import { Dictionary } from './utils';
+
+export interface IForwardingStatsUploader {
+    startForwardingStatsTimer(): void;
+}
+
+export default function forwardingStatsUploader(
+    this: IForwardingStatsUploader,
+    mpInstance: IMParticleWebSDKInstance
+): void {
+    this.startForwardingStatsTimer = function(): void {
+        (window.mParticle as Dictionary)._forwardingStatsTimer = setInterval(
+            function() {
+                prepareAndSendForwardingStatsBatch();
+            },
+            (mpInstance._Store.SDKConfig as Dictionary).forwarderStatsTimeout
+        );
     };
 
-    function prepareAndSendForwardingStatsBatch() {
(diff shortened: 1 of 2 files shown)


rmi22186:
**Tests:** unchanged (no `test/` files). Core/Jest green.

Migration looks faithful — `IForwardingStatsUploader` wiring in `mp-instance.ts` is good, and `window.mParticle` is a clearer stand-in for the old bare global.

**Nit (same theme as #1262):** prefer typing the surface instead of casting:
- add `forwarderStatsTimeout` to `SDKConfig` (it’s already in `Constants` defaults) so you can drop `(mpInstance._Store.SDKConfig as Dictionary).forwarderStatsTimeout`
- type `_forwardingStatsTimer` on the manager/`mParticle` surface instead of `(window.mParticle as Dictionary)`

Not blocking if you want to land the rename first and follow up on types. Bugbot hits on `mp-instance.ts` look unrelated to this diff (import/type swap only).

src/forwardingStatsUploader.ts
```diff
@@ -1,12 +1,25 @@
...
+export interface IForwardingStatsUploader {
+    startForwardingStatsTimer(): void;
+}
+
+export default function forwardingStatsUploader(
+    this: IForwardingStatsUploader,
+    mpInstance: IMParticleWebSDKInstance
+): void {
+    this.startForwardingStatsTimer = function(): void {
+        (window.mParticle as Dictionary)._forwardingStatsTimer = setInterval(
```
alexs-mparticle, line 13: Can we use `IMParticleWebSDKInstance` here or does that cause too many cascading issues?

alexs-mparticle (commented):
Looks good overall. Only one question that isn't a blocker IMO

src/forwardingStatsUploader.ts
jaissica12: window.mParticle is the instance manager (IMParticleInstanceManager), not IMParticleWebSDKInstance — that’s already mpInstance. The timer lives on the global manager (same as JS / test reset). Using IMParticleWebSDKInstance here would be the wrong type; moving the timer onto mpInstance would change multi-instance behavior. If we drop the Dictionary cast, we should add _forwardingStatsTimer on IMParticleInstanceManager.

crisryantan (approved):

PR #1334, opened 2026-08-19
refactor: migrate filteredMparticleUser module to TS

jaissica12:
## Background
- Continuing the effort to migrate remaining JavaScript modules in the mParticle Web SDK to TypeScript for improved type safety and developer experience. `filteredMparticleUser.js` was one of the remaining JS modules used by forwarders and Rokt filtered-user construction.

## What Has Changed
- Converted `src/filteredMparticleUser.js` to `src/filteredMparticleUser.ts` with type annotations
- Added `IFilteredMparticleUser` interface
- Filtering logic for identities/attributes is preserved
- No intentional public API or functional changes

## Screenshots/Video
- N/A

## Checklist

## Additional Notes
- No dedicated unit/integration tests exist specifically for this module
- `getAllUserAttributes` is referenced via a local function so the module typechecks when invoked without `new` (same call pattern as production call sites)

- Closes https://go/j/SDKE-1193


rmi22186:
**Tests:** unchanged (no `test/` files).

**Request:** add `filterUserAttributes` / `filterUserIdentitiesForForwarders` to `SDKHelpersApi` and call them typed, instead of:
```ts
((mpInstance._Helpers as Dictionary).filterUserAttributes as Function)(...)
```
Those casts undo most of the typing win from the migration.

Union `MPForwarder | { userAttributeFilters: number[] }` is fine for the Rokt path; a tiny shared filter-config type would be cleaner but optional.

Merge this before #1336 (already approved).

src/filteredMparticleUser.ts
```diff
@@ -0,0 +1,119 @@
...
+                            ] as string[]).slice();
+                        } else {
+                            userAttributesCopy[prop] = userAttributes[prop];
+                        }
+                    }
+                }
+            }
+        }
+
+        userAttributesCopy = ((mpInstance._Helpers as Dictionary)
```
alexs-mparticle: We should have an `IHelpersAPI` interface or something similar that can be used here.

jaissica12: `SDKHelpersApi` is the helpers interface (`_Helpers`). The casts were there because `filterUserAttributes` / `filterUserIdentitiesForForwarders` were optional. They’re required now, so this module uses `_Helpers` with no Dictionary/Function casts. No separate IHelpersAPI needed.

src/filteredMparticleUser.ts
```diff
@@ -0,0 +1,116 @@
...
+            );
+
+            return {
+                userIdentities: currentUserIdentities,
+            };
+        },
+        getMPID: function(): MPID {
+            return mpid;
+        },
+        getUserAttributesLists: function(
```
crisryantan, line 86: hey, quick question on this param, it shadows the closure forwarder, but getAllUserAttributes() on the next line still uses the closure one. so attributes get filtered by the closure forwarder's filters first, then by this param's, and the result is the intersection of both.

is this always called with the same forwarder the factory was built with? if so we could drop the param and use the closure one. if not, I think the double filter is a bug.

src/filteredMparticleUser.ts
```diff
@@ -0,0 +1,116 @@
...
+            let userAttributesLists: Dictionary<string[]> = {};
+
+            userAttributes = getAllUserAttributes();
+            for (const key in userAttributes) {
+                if (
+                    userAttributes.hasOwnProperty(key) &&
+                    Array.isArray(userAttributes[key])
+                ) {
+                    if (
+                        !kitBlocker ||
```
crisryantan, line 99: hmm. getAllUserAttributes() already excludes blocked keys, so this check can't ever be true here. same for the .slice() just below those arrays were already copied on the way out.

src/filteredMparticleUser.ts
```diff
@@ -0,0 +1,116 @@
...
+): IFilteredMparticleUser {
+    function getAllUserAttributes(): Dictionary {
+        let userAttributesCopy: Dictionary = {};
+        const userAttributes = mpInstance._Store.getUserAttributes(mpid);
+
+        if (userAttributes) {
+            for (const prop in userAttributes) {
+                if (userAttributes.hasOwnProperty(prop)) {
+                    if (
+                        !kitBlocker ||
```
crisryantan, line 29: nit: the kitBlocker && is redundant since || short-circuits — right side only runs when it's truthy.

src/filteredMparticleUser.ts
```diff
@@ -0,0 +1,116 @@
...
+    getUserIdentities(): { userIdentities: Dictionary<string> };
+    getMPID(): MPID;
+    getUserAttributesLists(forwarder: MPForwarder): Dictionary<string[]>;
+    getAllUserAttributes(): Dictionary;
+}
+
+export default function filteredMparticleUser(
+    mpid: MPID,
+    forwarder: MPForwarder | { userAttributeFilters: number[] },
+    mpInstance: IMParticleWebSDKInstance,
```
crisryantan, line 18: question: the union allows a second shape without userIdentityFilters, but getUserIdentities casts to MPForwarder and reads it so that'd be undefined at runtime with the cast hiding it. is the object-literal variant only used for paths that don't touch identities?

src/filteredMparticleUser.ts
```diff
@@ -0,0 +1,116 @@
...
+                userIdentities: currentUserIdentities,
+            };
+        },
+        getMPID: function(): MPID {
+            return mpid;
+        },
+        getUserAttributesLists: function(
+            forwarder: MPForwarder
+        ): Dictionary<string[]> {
+            let userAttributes: Dictionary;
```
crisryantan, line 89: nit: could be const userAttributes = getAllUserAttributes(); since it's assigned right after.

src/filteredMparticleUser.ts
alexs-mparticle: I think this gets refactored in this pr https://github.com/mParticle/mparticle-web-sdk/pull/1336/changes#diff-54e89dac241c4da52b50178483a17a7125d3486cbf613a9f712c9fa9896c9d61R33

crisryantan (changes requested):
added a couple of comments

src/filteredMparticleUser.ts
jaissica12: Good catch, that’s the intersection of factory filters then the method arg. Same as the old JS, not introduced here. The SDK doesn’t call this method, kits do, and the param is part of that surface. Dropping it or using only the closure would be a behavior change, so leaving it on this migration. Happy to ticket a follow-up if we want to audit kits and collapse to one forwarder.

src/filteredMparticleUser.ts
jaissica12: Agreed, after getAllUserAttributes() the kit-block check can’t fire, and those arrays are already copies. Left as-is here to keep the JS→TS port 1:1. The complexity follow-up (#1336) still has the same pass in buildUserAttributeLists, we can drop it there.

src/filteredMparticleUser.ts
jaissica12: The object-literal shape is Rokt init ({ userAttributeFilters } only). Rokt does call getUserIdentities() on that user. `userIdentityFilters` is then undefined, which means no identity stripping (same as JS). The as MPForwarder just hides that. Kits from `forwarders.js` still pass a full forwarder with both filter lists. Not changing the union on this migration.

src/filteredMparticleUser.ts
jaissica12: Agreed, after getAllUserAttributes() the kit-block check can’t fire, and those arrays are already copies. Left as-is here to keep the JS→TS 1:1. Same redundant pass exists in #1336’s buildUserAttributeLists, it’s a no-op, not a behavior bug, so we’re leaving it there too rather than changing filtering in a complexity-only PR.

crisryantan (approved):

PR #1349, opened 2026-08-20
refactor: extract persistence helpers to reduce cognitive complexity

jaissica12:
## Background
- Sonar reports cognitive complexity failures on six methods in `persistence.ts`. The control flow is pre-existing from the JavaScript version; it is only being scored now because the persistence migration introduces the file as TypeScript. Keeping these refactors out of the migration PR keeps that diff limited to types.

## What Has Changed
- Extracted helpers so each of the following methods is at or under the allowed complexity of 15:
  - `initializeStorage` (was 38): first-run detection, localStorage/cookie merging, non-current-user MPID copying, and corrupt-storage cleanup are now separate functions.
  - `storeDataInMemory` (was 16): split into empty-persistence defaults and store hydration.
  - `reduceAndEncodePersistence` (was 59): the three MPID eviction strategies (no current session MPIDs, MPIDs outside the current session, then current session MPIDs by login order) are now separate functions.
  - `findPrevCookiesBasedOnUI` (was 31): identity matching split into per-MPID and per-identity-type helpers.
  - `encodePersistence` (was 55) and `decodePersistence` (was 61): global settings and per-MPID Base64 handling extracted into paired helpers.
- Cookie and localStorage merge order, MPID eviction priority, Base64 encode/decode, and identity cookie lookup results are unchanged.

## Screenshots/Video
- N/A. No user-facing or UI changes.

## Checklist

## Additional Notes
- Stacked on the persistence migration PR. Please confirm the base branch is `refactor/SDKE-1107-update-persistence-to-TS` and not `development` before review.
- This is a structural refactor with no intended behavior change, so it relies on the existing persistence test coverage rather than adding tests.
- Verified locally with `npx tsc --noEmit --skipLibCheck` and the persistence Jest specs.

- Closes https://go/j/SDKE-1107

### DIFF
--- a/src/persistence.ts
+++ b/src/persistence.ts
@@ -27,21 +27,114 @@ export default function _Persistence(
         );
     };
 
+    function setFirstRunFromExistingData(
+        localStorageData: IPersistenceMinified | null,
+        cookies: IPersistenceMinified | null
+    ): void {
+        if (!localStorageData && !cookies) {
+            mpInstance._Store.isFirstRun = true;
+            (mpInstance._Store as Dictionary).mpid = 0;
+            return;
+        }
+        mpInstance._Store.isFirstRun = false;
+    }
+
+    function mergeStorageSources(
+        localStorageData: IPersistenceMinified | null,
+        cookies: IPersistenceMinified | null
+    ) {
+        if (localStorageData && cookies) {
+            // https://go.mparticle.com/work/SQDSDKS-6047
+            return Utils.extend(false, localStorageData, cookies);
+        }
+        return localStorageData || cookies;
+    }
+
+    // For migrating from localStorage to cookies -- If an instance switches from
+    // localStorage to cookies, then no mParticle cookie exists yet and there is
+    // localStorage. Get the localStorage, set them to cookies, then delete the
+    // localStorage item.
+    function migrateLocalStorageToCookies(
+        storage: Storage,
+        localStorageData: IPersistenceMinified | null,
+        cookies: IPersistenceMinified | null
+    ) {
+        let allData;
+        if (localStorageData) {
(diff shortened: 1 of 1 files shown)


src/persistence.ts
```diff
@@ -13,6 +13,10 @@ const Base64 = Polyfill.Base64,
     SDKv2NonMPIDCookieKeys = Constants.SDKv2NonMPIDCookieKeys,
     StorageNames = Constants.StorageNames;
 
+function removeLocalStorage(localStorageName: string): void {
```
crisryantan: Why not call localStorage.removeItem directly?

src/persistence.ts
```diff
@@ -27,95 +31,131 @@ export default function _Persistence(
         );
     };
 
+    function setFirstRunFromExistingData(
+        localStorageData: IPersistenceMinified | null,
+        cookies: IPersistenceMinified | null
+    ): void {
+        if (!localStorageData && !cookies) {
+            mpInstance._Store.isFirstRun = true;
+            (mpInstance._Store as Dictionary).mpid = 0;
```
crisryantan, line 36: why did we need to cast here?

src/persistence.ts
```diff
@@ -27,95 +31,131 @@ export default function _Persistence(
...
+            return;
+        }
+        mpInstance._Store.isFirstRun = false;
+    }
+
+    function mergeStorageSources(
```
crisryantan, line 42: yeah i think it makes sense

crisryantan (changes requested):
added a few comments

src/persistence.ts
jaissica12: done

crisryantan (approved):

PR #1359, opened 2026-08-21
refactor(apv): encapsulate PageViewTracker so tests stop reaching into internals

mattbodle:
## Summary

Follow-up to the review thread on #1331 (["Why do we have a test helper in our code?"](https://github.com/mParticle/mparticle-web-sdk/pull/1331#discussion_r-mp-instance-286) and @crisryantan's suggestion to pull window access behind module-level helpers).

**Behaviour-preserving, with one deliberate widening.** This restructures `pageViewTracker.ts` so its tests can assert the contract instead of the implementation. #1331 has since merged, so this is rebased onto `development` and sits directly on top of it. The one behavioural delta against `development`: #1331's orphaned-wrapper guard moves from the two history wrappers into `safeHandleNavigation`, so it now also covers the popstate path.

The tracker had no observable surface, so the spec reached into privates in 21 places — `tracker['isActive']`, `tracker['lastPath']`, `tracker['originalPushState']`, `tracker['pushStateWrapper']`, `(window as any)[WIN_TRACKER_KEY]`, and a `jest.spyOn` on the private `handleNavigation`. Those assertions pin how the tracker is built, not what it guarantees, so the file could not be refactored without rewriting its tests.

## What changed

**Pure rules extracted as module functions** — `isNewPage`, `supportsHistoryTracking`, `buildPageViewEvent`. No `this`, no globals, no side effects. The dedup rule and the event shape are now assertable without a DOM, a timer, or a tracker instance.

**#1331's window helpers generalised into a full registry** — `getActiveTracker`, `hasInitialPageViewFired`, `markInitialPageViewFired`, `resetPageViewTracking`, superseding `getWindowTracker`/`setWindowTracker`. `typeof window` is checked in exactly one place, and neither the tracker nor `mp-instance` touches the flags directly.

**`PageViewTracker.resetWindowState(instanceTracker)` is gone.** It was a static whose docstring said "call from `_resetForTests`", and it required the caller to know the teardown ordering. `resetPageViewTracking()` takes no argument and resolves the tracker through the registry, which is the authoritative handle after a module re-evaluation. That also fixes the hazard raised in review: `_resetForTests` could clear `instance._PageViewTracker` while a *different*, still-patched tracker was the registered one, leaving its wrapper installed and unreachable.

```diff
-        const instanceTracker = instance._PageViewTracker;
-        if (instanceTracker) {
-            instanceTracker.teardown();
-            instance._PageViewTracker = undefined;
-        }
-        PageViewTracker.resetWindowState(instanceTracker);
+        resetPageViewTracking();
+        instance._PageViewTracker = undefined;
```

**#1331's orphaned-wrapper guard kept and tested.** Moved from the two history wrappers into `safeHandleNavigation`, so it also covers the popstate path. It had no test; it has one now. Worth noting what that test asserts: no event fires either way, because the deferred flush already aborts when inactive — so the guard's real effect is that a dead tracker does *no work* (no dedup walk, no queued timer), and that is what the test pins.

**Four history fields collapsed into one undo closure.** `patchHistory(onNavigate, log)` installs both wrappers and returns the function that undoes them, or `null` when it installed nothing (already wrapped, or a frozen `History` rejected the assignment). The tracker holds `undoHistoryPatch` and nothing else, so there is no half-patched state to inspect and no way to restore the wrong pair. The push/replace duplication (two near-identical wrapper definitions, two restore calls) goes with it.

**Deferred-fire duplication removed.** `init()` and `handleNavigation()` had two copies of the same `setTimeout` block; they now share `scheduleFire()`. The `init()` copy was missing the "aborted, tracker inactive" log, so inherited paths lost that diagnostic.

**Class surface narrowed** to `constructor` / `init()` / `teardown()` / `isActive`:

```diff
 export declare class PageViewTracker {
-    mpInstance: IMParticleWebSDKInstance;
-    static hasInitialPageViewFired(): boolean;
-    static markInitialPageViewFired(): void;
-    static resetWindowState(instanceTracker?: PageViewTracker): void;
+    private readonly mpInstance;
+    get isActive(): boolean;
     init(): void;
     teardown(): void;
-    takePendingNavigations(): string[];
+    private takePendingNavigations;
 }
```

`takePendingNavigations` is private but still reachable across instances (TypeScript scopes `private` to the class, not the instance), so the handoff protocol stays internal. `isActive` replaces the private field the integration test in #1331 was already reading.

**One redundant condition dropped in `completeSDKInitialization`.** The initial page view was guarded by *both* `!mpInstance._PageViewTracker` and the window flag. A tracker can only exist because a previous `init()` created it, and that `init()` either fired the page view or skipped because the flag was already set — so the flag is set in every path where a tracker exists. The nested branch could not change the outcome; the flag is the only real guard.

## Why this is better

- **The tests describe the contract.** 21 private reach-ins → 0. Every assertion is now something a consumer can observe: an event logged, a history method restored, `isActive`, or a registry helper. The next change to this file does not require rewriting its spec.
- **The interesting logic is reachable without a harness.** The dedup rule, the environment check, the event shape, and the history patch are each callable on their own. `patchHistory` has its own describe block with no tracker in sight.
- **Side effects are cornered.** `window` reads/writes live in five short functions at the top of the file; the class touches `window` only to add/remove its listener and to read `location`/`title` at flush time. Previously `window` was reached from eight places across the class and two more from `mp-instance`.
- **There is no test-only API on the tracker.** `resetPageViewTracking()` is a real lifecycle operation — it is what you call to return a page to its pre-APV state — that `_resetForTests` happens to use.
- **Less code doing the same thing.** Two wrapper definitions → one loop. Two deferred-fire blocks → one. Four state fields → one closure. Nine `'mParticle APV: '` prefixes → one `log()`.

## Test plan

- [x] `npx jest` — **658 passed, 0 failed**. `pageViewTracker.spec.ts` goes 33 → 57 tests.
- [x] `karma start test/karma.config.js --browsers ChromeHeadless` — **1089 passed, 0 failed** (`FirefoxHeadless` is not installed locally, so it was excluded; CI covers it).
- [x] `tsc -p .` — clean.
- [x] `eslint src/pageViewTracker.ts` — no new findings (the file was prettier-parseable before and still is; the pre-existing repo-wide `no-undef`/`no-unused-vars` noise from the non-TS-aware config is unchanged).
- [x] Mutation-checked the new spec rather than trusting the green: dropping the pending-navigation transfer, inverting the dedup rule, inverting the `isActive` guard on the deferred flush, reading the path live at flush time, and skipping the rollback on a rejected history patch each fail 1–12 tests. The last of those is caught by exactly one test, so the added rollback case is not redundant coverage.

## Not in this PR

`_resetForTests` itself still ships in `mp-instance.ts`. Removing it is a separate job — **104 call sites across 38 test files**, plus 7 references in `src/` — and it is the karma suite's only reset mechanism. This PR removes the tracker's *participation* in it; the helper's own removal needs its own change.

### DIFF
--- a/src/pageViewTracker.ts
+++ b/src/pageViewTracker.ts
@@ -1,7 +1,12 @@
 import { IMParticleWebSDKInstance } from './mp-instance';
+import { BaseEvent } from './sdkRuntimeModels';
 import { EventType, MessageType } from './types';
 
 type HistoryStateMethod = History['pushState'];
+type HistoryMethodName = 'pushState' | 'replaceState';
+type NavigationSource = HistoryMethodName | 'popstate';
+
+const HISTORY_METHODS: HistoryMethodName[] = ['pushState', 'replaceState'];
 
 const WRAPPED_MARKER = '__mpApvWrapped__';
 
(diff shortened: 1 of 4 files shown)


src/pageViewTracker.ts
```diff
@@ -1,382 +1,413 @@
...
-        setWindowTracker(undefined);
+    const active = win[WIN_TRACKER_KEY];
+    if (active) {
+        active.teardown();
     }
 
-    mpInstance: IMParticleWebSDKInstance;
+    // teardown() clears WIN_TRACKER_KEY itself; delete unconditionally in case a
+    // tracker was registered by an older build whose teardown did not.
+    delete win[WIN_TRACKER_KEY];
```
crisryantan: I cringe a bit every time I see `delete` used. It's not a perf issue on window, but the deeper thing is that we're managing two loose window keys, which is what forces the unconditional delete plus the "in case an older build didn't" caveat.

Could we collapse them into one __mpApv__ state object ({ tracker?, initialPageViewFired })? Reset then becomes a single reassignment to a fresh state, the delete goes away, and the older-build case is handled by construction rather than by comment.

crisryantan (dismissed):
added a suggestion to replace the usage of `delete` in this file

src/pageViewTracker.ts
mattbodle: Done in 65177e8f — collapsed to a single `__mpApv__` key holding `{ tracker?, initialPageViewFired }`.

Reset is now one reassignment to fresh state, so the `delete` is gone and there is no ordering in which half the state survives:

```ts
export const resetPageViewTracking = (): void => {
    const win = apvWindow();
    if (!win) return;

    const active = getActiveTracker();
    if (active) active.teardown();

    win[WIN_APV_KEY] = freshState();
};
```

Two notes on top of your suggestion:

- **The older-build caveat is not just handled by construction, it was moot.** I checked before renaming: #1331 is in no tagged release (`v2.79.1` predates it and does not even contain this file), so `__mpApvTracker__` / `__mpApvInitPVLogged__` have never shipped. No back-compat surface, so the rename is free.
- **Reads no longer create the object.** `readState()` tolerates absent state and only the two writers (`markInitialPageViewFired`, `setActiveTracker`) construct it, so a read never mutates `window` — otherwise `hasInitialPageViewFired()` would have become a lazy-init side effect, which is the sort of thing the rest of this PR is trying to remove.

Two specs added: one pins the single-key contract and shape, one asserts reset swaps the object rather than mutating it (`expect(after).not.toBe(before)`).

mattbodle:
## Review feedback addressed (65177e8f)

**@crisryantan — the `delete` / two loose window keys.** Done. Collapsed into a single `__mpApv__` object; reset is one reassignment. Details in [the thread](https://github.com/mParticle/mparticle-web-sdk/pull/1359#discussion_r3833326208).

**SonarCloud (3 new MINOR smells).**

- `S7741` (`typeof` vs direct `undefined` comparison, `supportsHistoryTracking`) — **fixed**.
- `S6582` ×2 (prefer optional chaining) — **not adopted, deliberately.** The pinned prettier in this repo cannot parse `?.`; introducing one produces `Parsing error: Expression expected` from `prettier/prettier`, which silently stops it formatting the entire file. I verified this by adding a single `?.` and re-running eslint. This is not hypothetical — `src/mp-instance.ts` is already in that state because of the existing `self._Store?.isInitialized` on line 166, so it currently gets no prettier formatting at all.

  Worth a separate issue: bumping prettier would clear both the Sonar smells and un-break `mp-instance.ts`. Happy to raise it if useful — it felt out of scope for this PR.

Verification unchanged: **1089 karma + 659 jest passing**, `tsc -p .` clean, `dist/` untouched (CI regenerates).

src/pageViewTracker.ts
```diff
@@ -1,382 +1,442 @@
...
+};
+
+export const getActiveTracker = (): PageViewTracker | undefined => {
+    const state = readState();
+    return state ? state.tracker : undefined;
+};
+
+export const hasInitialPageViewFired = (): boolean => {
+    const state = readState();
+    return !!(state && state.initialPageViewFired);
```
alexs-mparticle, line 120: Agreed. I think optional chaining reads cleaner.

alexs-mparticle (approved):
LGTM

PR #1360, opened 2026-08-24
refactor: replace persistence any with accurate cookie types

jaissica12:
## Background
- Follow-up to the persistence JS → TS migration and the cognitive-complexity extract.
- Review asked us to drop `any` now that this module is TypeScript (`decodePersistence` / cookie expiry / logged-in flag / instance manager), instead of leaving Dictionary/`any` casts from the 1:1 JS port.

## What Has Changed
- Typed on-disk logged-in as `boolean | 0 | 1` on `IPersistenceMinified` (`l` is written as `0|1`, coerced to boolean on decode).
- Parse/decode cookie and localStorage blobs as `IPersistenceMinified` instead of `any`.
- Cookie expiry uses `toUTCString()` (same GMT date string as `expireCookies`) and `SDKConfig.cookieExpiration` instead of `as any` / `toGMTString()`.
- Typed `_forceNoLocalStorage` and `_isTestEnv` on the instance manager; no `window as any` / `Dictionary` for those reads.
- `decodePersistence` accepts `string | null` (localStorage `getItem`).
- `JSON.stringify(mpidRecord, null, 2)` instead of `0 as any` as the replacer.

## Screenshots/Video
- N/A

## Checklist

## Additional Notes
- Stacked on `refactor/SDKE-1107-persistence-complexity-followup`. Merge migrate → complexity follow-up → this PR.
- No partner/public API or cookie name/schema change. `toUTCString()` matches the existing expire-cookie path.
- Persistence Jest (`test/jest/persistence.spec.ts`) passing locally. No new tests; this is types-only.

- Relates to https://go/j/SDKE-1107

### DIFF
--- a/src/persistence.interfaces.ts
+++ b/src/persistence.interfaces.ts
@@ -43,8 +43,7 @@ export interface IGlobalStoreV2MinifiedKeys {
 export interface IPersistenceMinified extends Dictionary {
     cu: MPID; // Current User MPID
     gs: IGlobalStoreV2MinifiedKeys;
-    // Written as 0/1 on disk, coerced to boolean on decode
-    l: boolean | 0 | 1;
+    l: boolean | 0 | 1; // IsLoggedIn
 
     // Persistence Minified can also store optional dictionaries with
     // an idex of MPID
--- a/src/persistence.interfaces.ts
+++ b/src/persistence.interfaces.ts
@@ -101,7 +101,7 @@ export interface IPersistence {
     initializeStorage(): void;
     update(): void;
     storeDataInMemory(obj: IPersistenceMinified, currentMPID?: MPID): void;
-            determineLocalStorageAvailability(storage?: Storage): boolean;
+    determineLocalStorageAvailability(storage?: Storage): boolean;
     setLocalStorage(): void;
     getLocalStorage(): IPersistenceMinified | null;
     expireCookies(cookieName: string): void;
--- a/src/persistence.interfaces.ts
+++ b/src/persistence.interfaces.ts
@@ -43,7 +43,8 @@ export interface IGlobalStoreV2MinifiedKeys {
 export interface IPersistenceMinified extends Dictionary {
     cu: MPID; // Current User MPID
     gs: IGlobalStoreV2MinifiedKeys;
-    l: boolean; // IsLoggedIn
+    // Written as 0/1 on disk, coerced to boolean on decode
+    l: boolean | 0 | 1;
 
     // Persistence Minified can also store optional dictionaries with
     // an idex of MPID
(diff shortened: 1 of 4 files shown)


src/persistence.interfaces.ts
```diff
@@ -43,7 +43,8 @@ export interface IGlobalStoreV2MinifiedKeys {
 export interface IPersistenceMinified extends Dictionary {
     cu: MPID; // Current User MPID
     gs: IGlobalStoreV2MinifiedKeys;
-    l: boolean; // IsLoggedIn
+    // Written as 0/1 on disk, coerced to boolean on decode
```
crisryantan: nit, we can cleanup

src/persistence.ts
```diff
@@ -259,18 +264,18 @@ export default function _Persistence(
...
     // https://go.mparticle.com/work/SQDSDKS-5022
-    this.determineLocalStorageAvailability = function(storage: Storage): boolean {
-        let result;
-
-        if (window.mParticle && (window.mParticle as Dictionary)._forceNoLocalStorage) {
-            storage = undefined as any;
+    this.determineLocalStorageAvailability = function(
+        storage?: Storage
+    ): boolean {
+        if (getMParticleManager()?._forceNoLocalStorage) {
```
crisryantan:
```suggestion
        const mparticleManager = getMParticleManager()
        if (getMParticleManager()?._forceNoLocalStorage) {
```
From a perf standpoint, I think it would be better if we assign a const rather than calling the `getMParticleManager()` multiple times.

crisryantan: as a result you can also update this too https://github.com/mParticle/mparticle-web-sdk/pull/1360/changes#diff-d74addbec17763255f65ef752a33e2f0355f77d42534c05a8518ba3b3fa3bfbfL1185

src/persistence.ts
```diff
@@ -358,19 +363,24 @@ export default function _Persistence(
...
-                }
+        if (!decodedPersistence) {
+            return null;
+        }
+
+        const parsedPersistence = JSON.parse(
+            decodedPersistence
+        ) as IPersistenceMinified;
+        const obj: IPersistenceMinified = {} as IPersistenceMinified;
+        let j: string;
```
crisryantan: we probably want a more descriptive variable name for readability

src/persistence.interfaces.ts
jaissica12: done

src/persistence.ts
jaissica12: done

src/persistence.ts
jaissica12: done

crisryantan (approved):

PR #1363, opened 2026-08-24
refactor: reduce forwarders cognitive complexity for Sonar

jaissica12:
## Background
- Stacked on top of [#1260](https://github.com/mParticle/mparticle-web-sdk/pull/1260) (`refactor/SDKE-1109-update-forwarders-to-TS`).
  - `isEnabledForUserAttributes` — 18
  - `sendEventToForwarders` — 17
  - `setForwarderOnIdentityComplete` — 24

## What Has Changed
- `isEnabledForUserAttributes`: extracted the hash-and-compare loop into a module-level `userAttributesMatchFilter` helper that returns on the first match, instead of setting an `isMatch` flag and breaking. Also dropped the unreachable `if (filterObject) … else return true` tail, since `filterObject` is already guaranteed by the guard at the top of the function. `user.getAllUserAttributes()` still runs outside the `try`, so a throw there propagates exactly as before rather than being swallowed into `return true`.
- `sendEventToForwarders`: inverted the outer `if` into an early return and moved the loop body into a `forwardEvent` closure, so each `continue` becomes a `return`. The indexed `for` loop is unchanged (not converted to `forEach`), so iteration semantics stay identical if a kit mutates `activeForwarders` mid-send.
- `setForwarderOnIdentityComplete`: replaced the four-branch `else if` chain with a module-level map from identity method to kit callback name (`identify` → `onIdentifyComplete`, etc.), reducing the body to one lookup and one null check.

## Screenshots/Video
- N/A — internal refactor, no user-facing behavior change.

## Checklist

- Closes https://go/j/SDKE-1109

### DIFF
--- a/src/forwarders.ts
+++ b/src/forwarders.ts
@@ -301,9 +301,7 @@ export default function Forwarders(
             }
         };
 
-        for (let i = 0; i < mpInstance._Store.activeForwarders.length; i++) {
-            forwardEvent(mpInstance._Store.activeForwarders[i]);
-        }
+        mpInstance._Store.activeForwarders.forEach(forwardEvent);
     };
 
     this.handleForwarderUserAttributes = function(
--- a/src/forwarders.ts
+++ b/src/forwarders.ts
@@ -301,7 +301,9 @@ export default function Forwarders(
             }
         };
 
-        mpInstance._Store.activeForwarders.forEach(forwardEvent);
+        for (const forwarder of mpInstance._Store.activeForwarders) {
+            forwardEvent(forwarder);
+        }
     };
 
     this.handleForwarderUserAttributes = function(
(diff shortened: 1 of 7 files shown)


src/forwarders.ts
```diff
@@ -211,84 +227,82 @@ export default function Forwarders(
...
+                'Sending message to forwarder: ' + forwarder.name
+            );
+            const result = forwarder.process(clonedEvent);
+
+            if (result) {
+                mpInstance.Logger.verbose(result);
+            }
+        };
+
+        for (let i = 0; i < mpInstance._Store.activeForwarders.length; i++) {
```
mattbodle: As per sonarcloud, why not for-of or forEach?

mattbodle (approved):

src/forwarders.ts
crisryantan: Would appreciate explicit variables for readability tbh. An agents.md improvement soon 👀

crisryantan (approved):
Had 1 minor comment so I approved as I don't see it blocking

src/forwarders.ts
jaissica12: done

jaissica12: done

PR #1366, opened 2026-08-24
feat(apv): attach allowlisted query params to auto page views

mattbodle:
## Summary

The auto page view carried `hostname`, `title` and — for SPA navigations — `path`. The query string was never read, so campaign attribution and pagination context were absent from every APV event.

Both APV emitters now attach an allowlisted set of query params as flat event attributes.

```js
// before
data: { hostname: 'shop.example.com', title: 'Shoes', path: '/search' }

// after
data: {
  hostname: 'shop.example.com',
  title:    'Shoes',
  path:     '/search',
  utm_source: 'google',
  utm_medium: 'cpc',
  gclid:      'Cj0KC...',
}
```

## There are two APV emitters, and only one is the tracker

This is the part worth reviewing carefully. The APV feature fires from two places:

| | emitter | fires on |
| :-- | :-- | :-- |
| Landing page view | `_Events.logPageView()` (`src/events.ts`) | `mParticle.init()` |
| SPA navigations | `PageViewTracker.firePageView()` | `pushState` / `replaceState` / `popstate` |

The **landing** view is the one that matters for campaign attribution, because `utm_*` and `gclid` live on the entry URL by definition. Patching only the tracker would have shipped a change that captured attribution params approximately never. Both are changed.

## Filtering reuses the existing parser

`utils.queryStringParser(url, keys)` already does allowlist matching, lowercases keys (so `?UTM_Source=` and `?utm_source=` land on one attribute), drops empty values, and carries the fallback for browsers without `URLSearchParams`. `integrationCapture.ts` already consumes it the same way. The new helper is one line:

```ts
export const allowedQueryParams = (href: string): Dictionary<string> =>
    queryStringParser(href, ALLOWED_QUERY_PARAMS);
```

It takes `getHref()`, which yields `''` under SSR, so that path returns no params rather than throwing.

## An allowlist, not a denylist

Partner URLs routinely hold order ids, email addresses and session tokens. None of those should reach the event stream because nobody remembered to exclude them, so anything absent from `ALLOWED_QUERY_PARAMS` is dropped. 31 params, grouped:

- **Campaign attribution** — `utm_source` `utm_medium` `utm_campaign` `utm_term` `utm_content` `utm_id`
- **Ad-network click ids** — `gclid` `gbraid` `wbraid` `fbclid` `msclkid` `ttclid` `twclid` `li_fat_id` `dclid`
- **OAuth / OIDC** — `client_id` `redirect_uri` `response_type` `scope` `state` `code` `nonce`
- **Pagination and search** — `page` `limit` `offset` `cursor` `per_page` `q` `search`
- **Referral** — `ref` `referrer`

Params are spread **before** the core fields, so a future allowlist addition cannot silently overwrite `hostname` / `title` / `path`. No current entry collides; the ordering and its test are what keep it that way.

## ⚠️ SECURITY: `code`, `state` and `nonce` are on the list by explicit product decision

These are the OAuth 2.0 / OIDC authorization code and the CSRF/replay tokens. They are credentials until redeemed. Attaching them means they are persisted in the event store and forwarded to every configured kit.

This was raised and the decision was to include them. A `SECURITY:` comment above `ALLOWED_QUERY_PARAMS` records that, so it reads as a decision rather than an accident, and removing those three entries is the whole of the rollback. Flagging here so reviewers see it rather than discover it.

`q` carries the same shape of risk at lower severity — free-text consumer search queries routinely contain PII.

## Behaviour change: dedup now keys on the params too

Dedup was `pathname` only. It is now `pathname` + the allowlisted params, sorted so that reordering the query string is not a new page.

- `?page=2` → `?page=3` **now fires** a page view. Without this, `page` / `limit` / `offset` / `cursor` / `per_page` / `q` would be captured only on navigations that also changed the path — i.e. almost never.
- `?tab=1` → `?tab=2` (unallowlisted) still fires nothing.
- A hash-only change still fires nothing.

**This raises APV event volume**, and reviewers should size that against their own traffic. The sharp edge is search: an SPA that pushes `?q=` per keystroke will emit a page view per keystroke. If that is unacceptable, dropping `q` from the allowlist (or reverting dedup to path-only) are both one-line changes.

## Verbose logs print param names, never values

Not requested, but console output is captured and shipped off-domain by session-replay tooling, which is not a place for an OAuth code or a consumer's search terms. `describePage()` emits `/callback (params: code)`, never the value. Asserted by a test.

## Verification

- `jest` — **692 pass / 0 fail** (was 679; +13)
- `karma` ChromeHeadless — **1089 pass / 0 fail**
- `tsc -p .` clean, `eslint src/ test/src/` clean
- **Mutation-checked.** Seven deliberate breakages, each caught by a named test: unsorted dedup key, params spread after the core fields, allowlist removed, `describePage` logging values, params re-read at flush time instead of navigation time, dedup reverted to path-only, and the spread dropped from `events.ts`.

New coverage: `#allowedQueryParams` (6), `#pageKey` (4), `#buildPageViewEvent` params + collision + aliasing, allowlisted-vs-unallowlisted dedup, capture-at-navigation-time, the log-safety assertion, and a new `test/jest/events-page-view.spec.ts` (3) for the landing-page emitter.

> The landing-page assertions are in a new jest spec rather than the karma suite deliberately. Asserting it in karma requires mutating `window.location` while a live `PageViewTracker` is patched into `history`, which queues deferred page views that land in the *next* test and broke two sibling APV tests. The jest spec drives `Events` against a stub instance instead — isolated, deterministic, no bundle rebuild.

## Out of scope, flagged for follow-up

1. **`serverModel.ts:357` sends `window.location.href` as `LaunchReferral`** on session start — the full URL, every param, unfiltered. That predates this PR and is a wider exposure than anything the allowlist governs. This PR does not touch it.
2. **The landing page view still has no `path` attribute** while SPA views do, so downstream cannot tell which page the entry view was for. One line to fix, but it changes the shape of an existing event, so it wants a data-plan decision first.
3. **New attribute keys need adding to consumers' mParticle data plans** before they will validate.

### DIFF
--- a/src/pageViewTracker.ts
+++ b/src/pageViewTracker.ts
@@ -165,17 +165,22 @@ export const supportsHistoryTracking = (win: Window | null): boolean =>
 // Mirrors the event shape of the public mParticle.logPageView(), but carries the
 // path and query params captured when the navigation was accepted rather than the
 // live location.
-export const buildPageViewEvent = (data: IPageViewData): BaseEvent => ({
+export const buildPageViewEvent = ({
+    params,
+    hostname,
+    title,
+    path,
+}: IPageViewData): BaseEvent => ({
     messageType: MessageType.PageView,
     name: 'PageView',
     // Params spread first, then the core fields by name, so a core field always
     // wins. No allowlist entry collides with hostname/title/path today; naming
     // them here is what keeps a later addition from silently overwriting one.
     data: {
-        ...data.params,
-        hostname: data.hostname,
-        title: data.title,
-        path: data.path,
+        ...params,
+        hostname,
+        title,
+        path,
     },
     eventType: EventType.Unknown,
 });
--- a/src/pageViewTracker.ts
+++ b/src/pageViewTracker.ts
@@ -123,12 +123,18 @@ interface IPendingNavigation {
 export const allowedQueryParams = (href: string): Dictionary<string> =>
     queryStringParser(href, ALLOWED_QUERY_PARAMS);
 
-// The dedup key: pathname plus the allowlisted params, sorted so that reordering
-// the query string is not a new page. Params outside the allowlist never make it
-// into `page.params` and so cannot key a view — nor can the hash.
+// The captured params, in allowlist order. Ordering comes from the constant
+// rather than a sort: it is deterministic without needing a comparator, and it
+// does not depend on the object's insertion order, so reordering the query string
+// cannot produce a different key.
+const capturedNames = (params: Dictionary<string>): string[] =>
+    ALLOWED_QUERY_PARAMS.filter(name => name in params);
+
+// The dedup key: pathname plus the allowlisted params in a fixed order, so that
+// reordering the query string is not a new page. Params outside the allowlist
+// never make it into `page.params` and so cannot key a view — nor can the hash.
 export const pageKey = (page: IPageSnapshot): string => {
-    const query = Object.keys(page.params)
-        .sort()
+    const query = capturedNames(page.params)
         .map(name => `${name}=${page.params[name]}`)
         .join('&');
 
@@ -260,7 +266,7 @@ const describePage = (page: IPageSnapshot | null): string => {
         return 'null';
     }
 
-    const names = Object.keys(page.params).sort();
+    const names = capturedNames(page.params);
     return names.length ? `${page.path} (params: ${names.join()})` : page.path;
 };
 
--- a/src/pageViewTracker.ts
+++ b/src/pageViewTracker.ts
@@ -1,6 +1,7 @@
 import { IMParticleWebSDKInstance } from './mp-instance';
 import { BaseEvent } from './sdkRuntimeModels';
 import { EventType, MessageType } from './types';
+import { Dictionary, getHref, queryStringParser } from './utils';
 
 type HistoryStateMethod = History['pushState'];
 type HistoryMethodName = 'pushState' | 'replaceState';
@@ -41,14 +42,71 @@ type WindowWithApv = Window & {
     [WIN_APV_KEY]?: IApvState;
 };
 
-interface IPageViewData {
+// The query params an auto page view may carry. An allowlist, not a denylist:
+// partner URLs routinely hold order ids, email addresses and session tokens, and
+// none of those should reach the event stream because someone forgot to exclude
+// them. Anything absent from this list is dropped.
+//
+// SECURITY: `code`, `state` and `nonce` are the OAuth 2.0 / OIDC authorization
+// code and the CSRF/replay tokens. They are credentials until redeemed, and
+// attaching them here persists them in the event store and forwards them to
+// every configured kit. They are on the list by explicit product decision —
+// deleting that line is the whole of the fix if that decision is revisited.
+export const ALLOWED_QUERY_PARAMS: string[] = [
+    // Campaign attribution
+    'utm_source',
+    'utm_medium',
+    'utm_campaign',
+    'utm_term',
+    'utm_content',
+    'utm_id',
+
+    // Ad-network click ids
+    'gclid',
+    'gbraid',
+    'wbraid',
+    'fbclid',
+    'msclkid',
+    'ttclid',
+    'twclid',
+    'li_fat_id',
+    'dclid',
+
+    // OAuth / OIDC — see the SECURITY note above
+    'client_id',
+    'redirect_uri',
+    'response_type',
+    'scope',
+    'state',
+    'code',
+    'nonce',
+
+    // Pagination and search
+    'page',
+    'limit',
+    'offset',
+    'cursor',
+    'per_page',
+    'q',
+    'search',
+
+    // Referral
+    'ref',
+    'referrer',
+];
+
+interface IPageSnapshot {
+    path: string;
+    params: Dictionary<string>;
+}
+
+interface IPageViewData extends IPageSnapshot {
     hostname: string;
     title: string;
-    path: string;
 }
 
 interface IPendingNavigation {
-    path: string;
+    page: IPageSnapshot;
     timeoutId: ReturnType<typeof setTimeout>;
 }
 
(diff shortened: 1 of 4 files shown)


src/pageViewTracker.ts
```diff
@@ -71,11 +150,18 @@ export const supportsHistoryTracking = (win: Window | null): boolean =>
     typeof win.addEventListener === 'function';
 
 // Mirrors the event shape of the public mParticle.logPageView(), but carries the
-// path captured when the navigation was accepted rather than the live location.
-export const buildPageViewEvent = (data: IPageViewData): BaseEvent => ({
+// path and query params captured when the navigation was accepted rather than the
+// live location.
+export const buildPageViewEvent = ({
+    params,
+    ...page
```
crisryantan: Why do a spread here and then another [here](https://github.com/mParticle/mparticle-web-sdk/pull/1366/changes#diff-b44444f65ddf989f69eee9d04ff5864b7c5508e0a6ab85d7635f303dfccefe74R164)

src/pageViewTracker.ts
```diff
@@ -160,7 +252,23 @@ const clearActiveTracker = (tracker: PageViewTracker): void => {
...
+    params: allowedQueryParams(getHref()),
+});
+
+// Log-safe description of a page: the path, plus the NAMES of the captured
+// params. Values are deliberately omitted — verbose logging goes to the console,
+// and session-replay tooling ships console output off-domain, which is not a
+// place for an OAuth code or a consumer's search terms.
+const describePage = (page: IPageSnapshot | null): string => {
+    if (!page) {
+        return 'null';
```
crisryantan: why not just return an empty string?

src/pageViewTracker.ts
```diff
@@ -57,12 +115,33 @@ interface IPendingNavigation {
...
+// the query string is not a new page. Params outside the allowlist never make it
+// into `page.params` and so cannot key a view — nor can the hash.
+export const pageKey = (page: IPageSnapshot): string => {
+    const query = Object.keys(page.params)
+        .sort()
+        .map(name => `${name}=${page.params[name]}`)
+        .join('&');
+
+    return query ? `${page.path}?${query}` : page.path;
+};
```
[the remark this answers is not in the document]
mattbodle: Good catch — real, and it drops page views rather than just producing an odd key. Fixed in 69446607.

`queryStringParser` returns **decoded** values, so a value carrying the pair delimiters serialized exactly like two separate params:

- `{q: "a&search=b"}` → `q=a&search=b`
- `{q: "a", search: "b"}` → `q=a&search=b`

Identical, so `isNewPage` treated a genuine navigation between the two as the same page and skipped the view. Worth noting the collision needs the injected `name=value` to sit *after* its host param in `ALLOWED_QUERY_PARAMS` order — `search` follows `q`, and `response_type` follows `redirect_uri`, so an OAuth `redirect_uri` (which routinely carries encoded `&`/`=`) or a search query reaches it.

Values are now re-encoded when the key is built. Two regression tests: one states the case by hand, one drives it through the real parser from a URL so it covers `allowedQueryParams` actually producing the colliding shape. Both fail if the `encodeURIComponent` is removed — verified.

src/pageViewTracker.ts
mattbodle: Fair — it was needlessly clever. The rest-destructure existed only to get "everything except params" in one token, and it made the precedence rule invisible: you had to know how object rest interacts with spread order to see that a core field wins.

Dropped it in 69446607 — one spread, then the core fields named explicitly:

```ts
data: {
    ...data.params,
    hostname: data.hostname,
    title: data.title,
    path: data.path,
},
```

Same behaviour, and the reason for the ordering is now readable at the call site. The test that a param cannot overwrite a core field still fails if the two are swapped.

src/pageViewTracker.ts
mattbodle: Done in 69446607 — returns `""` now.

You are right that the literal string `"null"` reads like an accidentally stringified value. I had it that way to keep the log line unambiguous (`from: null, to: /products`), but the branch is unreachable: `lastPage` is assigned in `init()` before the history patch and the popstate listener are installed, so no navigation handler can observe it as null. The sentinel was paying for a case that cannot happen. Comment above the function now says so.

src/pageViewTracker.ts
```diff
@@ -71,11 +163,20 @@ export const supportsHistoryTracking = (win: Window | null): boolean =>
     typeof win.addEventListener === 'function';
 
 // Mirrors the event shape of the public mParticle.logPageView(), but carries the
-// path captured when the navigation was accepted rather than the live location.
+// path and query params captured when the navigation was accepted rather than the
+// live location.
 export const buildPageViewEvent = (data: IPageViewData): BaseEvent => ({
```
crisryantan: Probably cleaner to do something like this?
```
export const buildPageViewEvent = ({ params, hostname, title, path }: IPageViewData): BaseEvent => ({
    messageType: MessageType.PageView,
    name: 'PageView',
    // Params spread first, then the core fields by name, so a core field always
    // wins. No allowlist entry collides with hostname/title/path today; naming
    // them here is what keeps a later addition from silently overwriting one.
    data: {
        ...params,
        hostname,
        title,
        path,
    },
    eventType: EventType.Unknown,
});
```

mattbodle: Yes, cleaner — taken as written in d945c6fc, just wrapped to the file's 80-col style:

```ts
export const buildPageViewEvent = ({
    params,
    hostname,
    title,
    path,
}: IPageViewData): BaseEvent => ({
```

Destructuring the named fields beats repeating `data.` at each use, and unlike the rest-destructure you flagged earlier it keeps the precedence explicit. Re-checked that the "param cannot overwrite a core field" test still fails if `...params` is moved after the three.

src/pageViewTracker.ts
```diff
@@ -344,17 +468,17 @@ export class PageViewTracker {
...
 
-    // Cancels this tracker's deferred fires and hands back their captured paths.
-    private takePendingNavigations(): string[] {
+    // Cancels this tracker's deferred fires and hands back their captured pages.
+    private takePendingNavigations(): IPageSnapshot[] {
         const pending = this.pendingNavigations;
         this.pendingNavigations = [];
         pending.forEach(p => clearTimeout(p.timeoutId));
-        return pending.map(p => p.path);
+        return pending.map(p => p.page);
```
crisryantan:
```suggestion
        return pending.map({ page } => page);
```

mattbodle: Taken in 696d5175, with the parens added — `{ page } => page` does not parse, arrow-function parameter destructuring needs them:

```ts
pending.forEach(({ timeoutId }) => clearTimeout(timeoutId));
return pending.map(({ page }) => page);
```

Did the `clearTimeout` line on the way past so the two read the same.

One I left alone: the `filter` in `scheduleFire()` is `p => p.timeoutId !== timeoutId`, comparing against the outer `timeoutId` in scope. Destructuring there shadows it and would need a rename to stay correct, which costs more than it reads.

crisryantan (approved):

PR #1374, opened 2026-08-26
feat(apv): narrow the default query-param allowlist to page-identity params

mattbodle:
> **Merge LAST.** Depends on #1375 → #1373. This branch is stacked on both, so the diff shown here includes them until they land.
>
> The ordering is the point: these removals are only safe *after* customers can add their own params. See "Sequence" — steps 3 and 4 are a **hard** prerequisite, not just a practical one.

## Summary

Narrows `ALLOWED_QUERY_PARAMS` from **31 → 9**, in two commits with two unrelated reasons. What is left is the set that identifies **which page you are on**, which is also what the dedup key needs:

```
page  limit  offset  cursor  per_page      (pagination)
q     search                               (search)
ref   referrer                             (referral)
```

| commit | removed | why |
| :-- | :-- | :-- |
| `feat(apv): drop the OAuth/OIDC params…` | `code` `state` `nonce` `client_id` `redirect_uri` `response_type` `scope` | credentials, and a nested-query hazard |
| `feat(apv): drop the attribution params…` | `utm_source` `utm_medium` `utm_campaign` `utm_term` `utm_content` `utm_id` `gclid` `gbraid` `wbraid` `fbclid` `msclkid` `ttclid` `twclid` `li_fat_id` `dclid` | already carried on planes built for them |

---

# Part 1 — OAuth/OIDC (7 params)

## Two rationales, not one

They are not equally strong, so they are stated separately rather than blended.

**Credential exposure — `code`, `state`, `nonce`.** `code` is an OAuth 2.0 authorization code; `state` and `nonce` are CSRF and replay tokens. **They are credentials until redeemed.** Capturing them by default persisted them in the event store and forwarded them to every configured kit — for every customer with APV enabled, whether or not they had any use for them, by a product decision no individual customer opted into.

**No analytics use for this SDK, plus a nested-query hazard — `client_id`, `redirect_uri`, `response_type`, `scope`.** These are not secrets. This half is a judgement rather than a measurement.

> `client_id` is not an inert name. It is a real identifier in the GA4 world, and it already appears as an integration attribute elsewhere in this repo (`test/src/tests-integration-capture.ts` — a different mechanism, but the name is not meaningless). The honest claim is narrower: **I have no evidence anyone reads these four off page-view events, and `redirect_uri` routinely embeds a nested query string**, which is the concrete hazard.

> **Scope note, and it weakens this section.** This is not a data-egress change. `createEventObject` sets `PageUrl: getHref()` on **every** event (`src/serverModel.ts:315`, `src/batchUploader.ts:201`) and the converter emits it as `page_url` (`src/sdkToEventsApiConverter.ts:668`) — added in #1276, with a test asserting the full URL survives verbatim. **The raw `?code=` value already leaves the browser on every event and still will after this PR.** What changes is that it stops becoming an individually-keyed, queryable event attribute forwarded to every kit. That is a real reduction in blast radius, but it is narrower than "we stopped sending it", and the earlier revision of this description implied the stronger claim.

---

# Part 2 — attribution (15 params)

## The overlap, confirmed against source

**Click ids — 5 of 9 are captured by `IntegrationCapture` already**, as per-network custom flags:

| param | `integrationMappingExternal` | mapped to |
| :-- | :-- | :-- |
| `gclid` | yes | `GoogleEnhancedConversions.Gclid` |
| `gbraid` | yes | `GoogleEnhancedConversions.Gbraid` |
| `wbraid` | yes | `GoogleEnhancedConversions.Wbraid` |
| `fbclid` | yes | `Facebook.ClickId` — **transformed** to `fb.N.<ts>.<id>` |
| `ttclid` | yes | `TikTok.Callback` |
| `msclkid` `twclid` `li_fat_id` `dclid` | **no** | — |

It also captures `ScCid` (Snapchat), which APV never did, plus the `_fbp` / `_fbc` / `_ttp` / `_scid` cookies.

Two honest caveats, because they cut against a pure "duplicate" argument:

- **It is flag-gated.** `captureIntegrationSpecificIds.V2` defaults to `'none'` (`src/store.ts:814`); `'roktonly'` captures Rokt ids only. The overlap exists **only on `'all'`**.
- **Different data plane.** Custom flags are per-kit forwarding metadata; event attributes are queryable analytics. You cannot segment an audience on a custom flag, and `fbclid`'s IC value is not even the raw one.

**UTMs — no web-side duplicate at all.** `IntegrationCapture` has no `utm_*` mapping; neither does the Rokt kit or the Rokt web SDK. mPServer *does* parse `utm_*` (Adobe, Omniture, Kochava) but only from `AppInfo.InstallReferrer` — **Android/FireTV only, never web**.

The real overlap is a *convention*: mParticle has reserved user attributes `$utm_source` / `$utm_medium` / `$utm_term` / `$utm_content` / `$utm_campaign` / `$gclid` / `$campaign_id` (`GoogleAnalyticsRequestContext.cs:59-67`), which the GA forwarder maps to `cs` / `cm` / `ck` / `cc` / `cn` / `gclid` / `ci`. The web SDK never populates them. **So UTMs as page-view event attributes were a parallel copy no forwarder reads** — captured in the wrong place rather than captured twice.

## The independent reason: 15 attributes on every page view

There is **no per-event attribute cap in the SDK** — `sanitizeAttributes` validates types only. Forwarders impose their own, and they truncate rather than reject: **Flurry keeps the first 10 event attributes and silently drops the rest** (`FlurryMessageGeneratorBase.cs:53,362`). So a fat default does not add to the customer's data, it **displaces** it.

This reason holds regardless of whether you accept the overlap argument, and it is why `MAX_CUSTOM_QUERY_PARAMS`'s comment is rewritten: it previously claimed "mParticle caps attributes per event", which I could not substantiate anywhere in either repo.

---

## This is a change of default, not of capability

Since #1373 a customer adds query param names in the Web input's Advanced Settings, and those are unioned with this list.

```js
// default: dropped
allowedQueryParams('https://x.com/landing?gclid=abc')             // → []

// configured by the customer: captured
const { allowed } = parseQueryParamAllowlist('gclid');            // → ['gclid']
allowedQueryParams('https://x.com/landing?gclid=abc', allowed)    // → [{ name: 'gclid', value: 'abc' }]
```

The **second form is the one that matters**, and it is the one under test. While these were built-ins, `parseQueryParamAllowlist('gclid')` dropped it as a duplicate and returned `[]` — so a customer typing `gclid` into Advanced Settings would have received an empty list and the escape hatch would have silently done nothing. Removing them from the defaults is precisely what makes the string path work.

## Breaking-change note

**This is a breaking change for anyone reading these attributes today** — `gclid` and the UTMs shipped in `2.80.0`. A customer relying on them stops receiving them until they add them back in the UI.

### Release note — lead with promo codes and campaigns, not OAuth

`?code=` is heavily used for **coupon, promo and referral codes**, and `?state=` for **US state on checkout pages**. UTMs are near-universal. The customer-facing note should open with:

> If you read `utm_*`, `gclid` or `code` off auto page view events, add them under the Web input's Advanced Settings. For campaign reporting, prefer the reserved `$utm_*` user attributes, which forwarders already consume.

### It ships as a MINOR, not a major — and that is deliberate

Both commits are `feat(apv): …` with **no `!`**, plus a real `BREAKING CHANGE:` footer. Verified against the installed packages:

- `conventional-changelog-angular`'s `headerPattern` is `/^(\w*)(?:\((.*)\))?: (.*)$/`, which **does not match** `feat(apv)!:`. Parsed with the repo's own preset, a `!` commit comes back `type: undefined, notes: []` and `@semantic-release/commit-analyzer` returns **`null`** — no release at all.
- There is no `breakingHeaderPattern` in that preset, so `!` is never converted into a note. Only `noteKeywords: ['BREAKING CHANGE']` is read — hence the footer.

It returns `minor` rather than `major` because `release.config.js` lists `{ type: 'feat', release: 'minor' }` in `releaseRules`, consulted **before** semantic-release's default breaking rule.

**I have deliberately not touched `release.config.js`.** **Question for the release owners:** is `minor` the intended outcome, or should `releaseRules` grow a `{ breaking: true, release: 'major' }` entry first? If the latter, that should land as its own PR before this one.

## Verification

Measured on this branch's tip:

- `jest` — **795 pass / 0 fail**, up from **778** on #1373 (+17)
- `karma` ChromeHeadless — **1089 pass / 0 fail**, 10 skipped, after a fresh `npm run build && npm run build:test-bundle`
- `tsc -p . --noEmit` clean, `eslint src/ test/src/` clean

Mutation-checked:

| mutation | test(s) that fail |
| :-- | :-- |
| `'code'` restored to `ALLOWED_QUERY_PARAMS` | `should not capture the OAuth/OIDC param code by default` **and** `should accept a configured OAuth param through the string config path` |
| `'gclid'` restored | `should not capture the attribution param gclid by default` **and** `should capture the attribution param gclid once configured` — the second fails because a restored built-in is dropped from `allowed` as a duplicate |

The tests that used `utm_source` as their stand-in for "a built-in param" now use `ref`, and the long-value case uses `search` rather than `utm_content`. Substitutions, not deletions — the assertions are unchanged.

## Sequence

1. #1375 — hardening
2. #1373 — flag + union
5. **this PR** — narrow the defaults
6. follow-up — `-` prefix so a customer can also *remove* a default

> `src/mp-instance.ts:~250` merges the server config over the init config with a **shallow** `extend({}, config, result)`, and `src/store.ts:~734-738` then replaces `SDKConfig.flags` **wholesale** whenever the merged config has a `flags` key. So for any customer who has not set `requestConfig: false`, an init-time `config.flags.autoLogPageViewQueryParams` is discarded in favour of the server blob. The mPServer `SettingTemplate` row is a **hard** prerequisite for the escape hatch, not merely a practical one.

## Two follow-ups this PR does not do

- **`msclkid`, `twclid`, `li_fat_id` and `dclid` are in no capture path now.** They were never in `integrationMappingExternal`, so removing them here leaves a genuine gap. The right fix is to add them to `IntegrationCapture` alongside the other five, which is a separate change in a separate module.
- **mPServer's `BUILT_IN_QUERY_PARAMS` is a hand-copied duplicate of this list** and silently *drops* anything a customer types that is already a built-in. It must shrink with this PR or a customer who types `gclid` has it discarded as "you already have this" when they no longer do. Tracked on the mPServer side.

## Summary by CodeRabbit

* **New Features**
  * Added configurable query-parameter capture for automatically logged page views.
  * Supports custom parameter allowlists with validation, deduplication, size limits, and warnings for rejected entries.
  * Excludes attribution and OAuth/OIDC parameters by default unless explicitly configured.
  * Captured parameters are included in page-view event attributes.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -190,6 +190,10 @@ const Constants = {
         CaptureIntegrationSpecificIdsV2: 'captureIntegrationSpecificIds.V2',
         AstBackgroundEvents: 'astBackgroundEvents',
         AutoLogPageView: 'autoLogPageView',
+        // Comma-separated query param names the customer adds to the APV
+        // allowlist, from the Web input's Advanced Settings. Unioned with
+        // ALLOWED_QUERY_PARAMS; absent or empty means built-ins only.
+        AutoLogPageViewQueryParams: 'autoLogPageViewQueryParams',
     },
     DefaultInstance: 'default_instance',
     CCPAPurpose: 'data_sale_opt_out',
--- a/src/helpers.ts
+++ b/src/helpers.ts
@@ -6,6 +6,7 @@ import Validators from './validators';
 import KitFilterHelper from './kitFilterHelper';
 import { IMParticleWebSDKInstance } from './mp-instance';
 import { SDKHelpersApi } from './sdkRuntimeModels';
+import { IQueryParamAllowlist } from './pageViewTracker';
 import { IMParticleUser, IdentityCallback, IdentityResult, ISDKUserIdentity } from './identity-user-interfaces';
 import { IAliasResult } from './identity.interfaces';
 import { AliasUsersCallback, MPID, UserIdentities } from '@mparticle/web-sdk';
@@ -82,7 +83,12 @@ export default function Helpers(
         return false;
     };
 
-    this.getFeatureFlag = function(feature: string): boolean | string {
+    // The union is every shape processFlags can put in SDKConfig.flags. It has to
+    // include IQueryParamAllowlist or the APV callers need an `as unknown as` cast
+    // to read a value this function genuinely returns.
+    this.getFeatureFlag = function(
+        feature: string
+    ): boolean | string | IQueryParamAllowlist {
         if (mpInstance._Store.SDKConfig.flags.hasOwnProperty(feature)) {
             return mpInstance._Store.SDKConfig.flags[feature];
         }
--- a/kits/braze/braze-3/README.md
+++ b/kits/braze/braze-3/README.md
@@ -1,52 +1,116 @@
-![Braze Logo](https://github.com/mparticle-integrations/mparticle-javascript-integration-appboy/blob/master/braze-logo.png) 
+![Braze Logo](https://github.com/mparticle-integrations/mparticle-javascript-integration-appboy/blob/master/braze-logo.png)
 
-⚠️⚠️⚠️
-# Notice! Opt in is now available for Braze Web SDK V4 - Action Required
+# mParticle Braze Kit — Braze Web SDK V3
 
-You can now select what version of the Braze SDK you want to use when setting up a Braze connection in the mParticle UI.  Braze occasionally makes breaking changes to their SDK, so if you call `appboy` directly in your code, you will have to update your code to ensure your website performs as expected when updating versions of Braze.
+This kit bundles **Braze Web SDK V3** (`@braze/web-sdk@^3.5.0`) and exposes it on the page as `window.appboy`.
 
-Please review the [Braze Changelog](https://www.braze.com/docs/developer_guide/platform_integration_guides/web/changelog#400) and [V4 migration guide](https://github.com/braze-inc/braze-web-sdk/blob/master/UPGRADE_GUIDE.md) to learn about the differences between V3 and V4 and what changes you will need to make in your code. The most significant breaking changes are the replacement of the `appboy` class name with `braze`, in addition to the removal and renaming of several APIs.
+---
 
-You can opt into the latest major version of the Braze Web SDK whether you implement mParticle's Web SDK using npm or our snippet/CDN.
-* Customers who self-host mParticle via npm - You should add @mparticle/web-braze-kit version 4.0.0 or greater in your package.json now! You must also select `Version 4` under `Braze Web SDK Version`  in the Braze connection settings.
-* Customers who load mParticle via snippet/CDN - Opting in is available via the mParticle UI in your Braze connection settings.
+## ⚠️ We recommend upgrading to Braze Web SDK V6
 
-Note that the following is only one example.  Everywhere you manually call `appboy` needs to be updated similar to the below. If you are using NPM, you can skip to step 3.  Please be sure to test your site fully in development prior to releasing.
+Braze is now on **V6**, and V6 is the version mParticle recommends for all customers. We recommend updating straight to V6 for latest support of all Braze features.
 
+See the [Braze Event Integration](https://docs.mparticle.com/integrations/braze/event/) docs.
 
-* Step 1: Legacy code sample. Find all the places where your code references the `appboy.display` namespace.  Braze has removed all instances of the `display` namespace:
+### How to upgrade
+
+1. **Audit the code you own.** If you call Braze directly, find every `appboy` reference and compare it with the table below and Braze's upgrade documentation.
+2. **Move to the V6 kit.**
+   - Self-hosting via npm with **core Web SDK v3 (latest)**: `npm install @mparticle/web-braze-kit-6`
+   - Self-hosting via npm with **core Web SDK v2 (legacy)**: `npm install @mparticle/web-braze-kit@^6`
+   - Loading mParticle via snippet/CDN: nothing to install; the kit is delivered for you.
+3. **Add defensive code** if you load mParticle via the snippet and call Braze directly. Skip this if you self-host via npm. See [Write version-tolerant code](#write-version-tolerant-code).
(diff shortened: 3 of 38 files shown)


src/constants.ts
```diff
@@ -196,6 +196,10 @@ const Constants = {
         CaptureIntegrationSpecificIdsV2: 'captureIntegrationSpecificIds.V2',
         AstBackgroundEvents: 'astBackgroundEvents',
         AutoLogPageView: 'autoLogPageView',
+        // Comma-separated query param names the customer adds to the APV
```
alexs-mparticle, line 193: Remove unnecessary comments

mattbodle: Thanks, Alex — done. That comment is gone, and I took the same pass over the rest of the stack: the source comments these changes add went from 116 lines to 25, keeping only the ones that explain a non-obvious constraint, and the stack now removes 40 older ones along the way.

This work has moved to #1475, #1476 and #1477, rebuilt on the current `v3-development`, so I'm closing this PR in favour of #1476.

src/helpers.ts
```diff
@@ -82,7 +83,12 @@ export default function Helpers(
         return false;
     };
 
-    this.getFeatureFlag = function(feature: string): boolean | string {
+    // The union is every shape processFlags can put in SDKConfig.flags. It has to
+    // include IQueryParamAllowlist or the APV callers need an `as unknown as` cast
+    // to read a value this function genuinely returns.
+    this.getFeatureFlag = function(
+        feature: string
+    ): boolean | string | IQueryParamAllowlist {
```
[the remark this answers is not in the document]
mattbodle: Agreed — fixed in #1475. Both `getFeatureFlag` in `helpers.ts` and `SDKHelpersApi.getFeatureFlag` now include `null` in the return type.

mattbodle:
Superseded by #1476: the same change, rebuilt as one signed commit on the current `v3-development` and opened from a branch in this repository rather than a fork. Closing in favour of that one — thanks!

PR #1375, opened 2026-08-26
fix: do not resolve query-param keys that name inherited prototype members

mattbodle:
## Summary

Two instances of the same bug class in `queryStringParser`, one latent and one reachable from a URL alone.

**1. Caller-supplied keys resolve through the prototype chain.** `queryStringParser` takes a caller-supplied list of keys and looks each one up on a plain object. `obj[name]` walks the prototype chain, so a key named `constructor` resolves to `Object.prototype.constructor`, passes the truthiness gate, and is copied into the result as though the URL had supplied it.

```js
queryStringParser('https://x.com?foo=bar', ['constructor'])
// before: { constructor: 'function Object() { [native code] }' }
// after:  {}
```

A real `?constructor=legitimate` in the URL still resolves — the guard is an own-property check, not a name ban. There is a test for that.

**2. `queryStringParserFallback` throws on a URL param named `hasOwnProperty`.** The fallback's returned `forEach` calls `params.hasOwnProperty(key)`, where `params`' keys come straight off the URL. `?hasOwnProperty=1` gives that map an own property shadowing the inherited method with the string `'1'`, and the next iteration invokes it.

```js
queryStringParser('https://x.com?hasOwnProperty=1&foo=bar', ['foo'])
// before: TypeError: params.hasOwnProperty is not a function
// after:  { foo: 'bar' }
```

That throws out of `queryStringParser`, through `allowedQueryParams` and `currentPage()`, and out of `IntegrationCapture.capture()` — **on the init path, from a URL alone, with no configuration required.** It is therefore strictly more reachable than (1), which needs a configurable key list to become input at all. The fallback is live code for browsers without `URL`/`URLSearchParams`, and the spec already exercises it (`describe('without URLSearchParams')`), which had no prototype-name coverage before this PR.

Both are fixed with one `hasOwnProp` helper.

## Reachability, stated plainly

- **(1) is not reachable in production today.** Both callers pass hardcoded key lists — `integrationCapture`'s click IDs and `pageViewTracker`'s `ALLOWED_QUERY_PARAMS` — and neither contains a prototype member name. It becomes reachable *input* the moment either list can be extended from configuration, which is the next PR.
- **(2) is reachable today**, on any browser taking the fallback path, from any URL carrying `?hasOwnProperty=`.

**`Object.prototype` is not polluted by any of this, before or after.** I verified. (1) is a data-hygiene and dedup-integrity bug and (2) is a crash — neither is a prototype-pollution vulnerability, and I would rather label them accurately than escalate.

## Why guard the lookup instead of `Object.create(null)`

`Object.create(null)` on the lookup map looks like the smaller fix and kills the whole class at the root. It is the wrong call here for two concrete reasons:

1. `queryStringParser` returns that same map **directly** when no keys are passed (`if (isEmpty(keys)) return lowerCaseUrlParams`), so the null prototype would escape to callers.
2. `integrationCapture.ts:336` — `applyProcessors` — calls `.hasOwnProperty()` on its `clickIds` argument, and `getQueryParams()` passes the return value of `queryStringParser` straight in. A null-prototype object throws there.

So the returned object stays plain, and a test pins that contract.

> Correction from the first revision of this description, which cited `integrationCapture.ts:289`. That line operates on `this.clickIds`, a fresh spread literal (`{...this.clickIds, ...queryParams, ...}`), so it is a plain object regardless of what `queryStringParser` returns. `:336` is the citation that actually holds. The same wrong claim was repeated in a test comment and is fixed there too.

## The other half, and its missing test

This also switches `pageViewTracker`'s `capturedNames()` from `in` to the same own-property check.

**That half has no test in this PR, deliberately.** `capturedNames()` only ever tests names drawn *from* `ALLOWED_QUERY_PARAMS`; with a hardcoded list containing no prototype member name, `in` and `hasOwnProperty` are unobservable through the public helpers.

> Correction: the first revision said the two "cannot be told apart by any test". That is false. `ALLOWED_QUERY_PARAMS` is a mutable export, so a test could push `'constructor'` onto it, assert `pageKey`, and pop it in a `finally`. The honest reason there is no test here is that the only possible one requires mutating an exported array that every other test in the file reads — shared mutable state in a suite that otherwise has none — and that was not worth the smell for one release. It gains a clean regression test in #1373, where the allowlist becomes configurable and the distinction is observable without touching module state.

## Test-quality corrections

The six-key prototype-member test (`['foo', 'constructor', '__proto__', 'toString', 'valueOf', 'hasOwnProperty']`) is **5/6 inert**, and its comment claimed otherwise. `queryStringParser` lowercases the key before the lookup, so:

| key | resolves to | exercises the guard? |
|---|---|---|
| `constructor` | `Object.prototype.constructor` (truthy) | **yes** |
| `__proto__` | `Object.prototype`, but assigning it back hits the setter rather than creating an own property | no |
| `toString` / `valueOf` / `hasOwnProperty` | `tostring` / `valueof` / `hasownproperty` → `undefined` | no |

The list is kept as documentation of the name class, with a comment that says which name does the work — and notes that the `.toLowerCase()` is therefore load-bearing for the other five, which was previously undocumented.

## Verification

Measured, not estimated:

- `jest` — **700 pass / 0 fail**, up from **694** on `origin/development` (+6: 4 from the original commit, 2 from the fallback fix)
- `karma` ChromeHeadless — **1089 pass / 0 fail**, 10 skipped
- `tsc -p . --noEmit` clean, `eslint src/ test/src/` clean

> Correction: the first revision said "+6 jest tests (from 692)". The original commit adds exactly 4 `it()` blocks and removes 0. The 698 total was right; the baseline was stale. Both numbers above are re-measured.

Mutation-checked — each fix deliberately broken, and the named test that catches it:

| mutation | test that fails |
|---|---|
| fallback guard reverted to `params.hasOwnProperty(key)` | `without URLSearchParams › survives a query param named after a prototype method` (and `› still returns a param named after a prototype method when asked for it`) |
| own-property guard deleted from `queryStringParser` | `with URLSearchParams › does not resolve keys naming inherited Object.prototype members` |
| guard replaced with a name ban (`name in Object.prototype`) | `with URLSearchParams › still resolves a real query param sharing a prototype member name` |

## Follow-ups, not fixed here

**Pre-existing full-harvest bug on `development`, unrelated to this PR.** `src/mp-instance.ts:~1546-1553` enables integration capture for any truthy `captureIntegrationSpecificIds.V2` value that is not `'none'`, but only assigns `captureMode` for `'all'` and `'roktonly'`. An unrecognised value leaves `captureMode` `undefined` → `getActiveIntegrationMapping()` returns `{}` → `queryStringParser(href, [])` takes the no-keys branch → **every query param on the page** is captured and forwarded. Verified structurally, not fixed here; it wants its own PR and its own risk assessment.

**Unsafe `obj.hasOwnProperty(...)` invocation elsewhere.** `src/utils.ts:47`, `:453`, `:502` and `src/integrationCapture.ts:289`/`:315`/`:336` all invoke `hasOwnProperty` as a method on an object whose keys may come from user data. A user attribute literally named `hasOwnProperty` throws at `utils.ts:453`. Same class as (2) above; a follow-up ticket rather than scope creep on this stack.

## Stack

Part 1 of 4 for customer-configurable APV query params:

1. **this PR** — hardening
2. #1373 — flag + union of `ALLOWED_QUERY_PARAMS` with a customer list

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -309,10 +309,8 @@ const mergeObjects = <T extends object>(...objects: T[]): T => {
 const moveElementToEnd = <T>(array: T[], index: number): T[] =>
     array.slice(0, index).concat(array.slice(index + 1), array[index]);
 
-// Own-property check that ignores inherited Object.prototype members. Needed
-// wherever a key is looked up by a name that did not come from the object itself:
-// `obj[name]` and `name in obj` both walk the prototype chain, so a name like
-// `constructor` resolves to something truthy on any plain object.
+// For keys that did not come from the object itself: `obj[name]` and `name in obj`
+// both walk the prototype chain, so `constructor` is truthy on any plain object.
 const hasOwnProp = (obj: object, key: string): boolean =>
     Object.prototype.hasOwnProperty.call(obj, key);
 
@@ -343,11 +341,7 @@ const queryStringParser = (
         keys.forEach(key => {
             const name = key.toLowerCase();
 
-            // The requested key may not have come from this URL — callers pass
-            // their own lists. Without the own-property guard, a key named
-            // `constructor` resolves to Object.prototype.constructor, passes the
-            // truthiness check below, and is copied out as if the URL had
-            // supplied it.
+            // Callers pass their own key list, so `name` may not be from this URL.
             if (!hasOwnProp(lowerCaseUrlParams, name)) {
                 return;
             }
@@ -390,12 +384,8 @@ const queryStringParserFallback = (url: string): URLSearchParamsFallback => {
         },
         forEach: function(callback: (value: string, key: string) => void) {
             for (var key in params) {
-                // Not `params.hasOwnProperty(key)`. The keys here come straight
-                // off the URL, so `?hasOwnProperty=1` gives `params` an own
-                // property shadowing the inherited method with the string `'1'`,
-                // and the next iteration calls it: `TypeError: params
-                // .hasOwnProperty is not a function`. That throws out of
-                // queryStringParser on the init path.
+                // Keys come straight off the URL, so `?hasOwnProperty=1` would
+                // shadow the method and `params.hasOwnProperty(key)` would throw.
                 if (hasOwnProp(params, key)) {
                     callback(params[key], key);
                 }
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -309,6 +309,13 @@ const mergeObjects = <T extends object>(...objects: T[]): T => {
 const moveElementToEnd = <T>(array: T[], index: number): T[] =>
     array.slice(0, index).concat(array.slice(index + 1), array[index]);
 
+// Own-property check that ignores inherited Object.prototype members. Needed
+// wherever a key is looked up by a name that did not come from the object itself:
+// `obj[name]` and `name in obj` both walk the prototype chain, so a name like
+// `constructor` resolves to something truthy on any plain object.
+const hasOwnProp = (obj: object, key: string): boolean =>
+    Object.prototype.hasOwnProperty.call(obj, key);
+
 const queryStringParser = (
     url: string,
     keys: string[] = []
@@ -334,7 +341,18 @@ const queryStringParser = (
         return lowerCaseUrlParams;
     } else {
         keys.forEach(key => {
-            const value = lowerCaseUrlParams[key.toLowerCase()];
+            const name = key.toLowerCase();
+
+            // The requested key may not have come from this URL — callers pass
+            // their own lists. Without the own-property guard, a key named
+            // `constructor` resolves to Object.prototype.constructor, passes the
+            // truthiness check below, and is copied out as if the URL had
+            // supplied it.
+            if (!hasOwnProp(lowerCaseUrlParams, name)) {
+                return;
+            }
+
+            const value = lowerCaseUrlParams[name];
             if (value) {
                 results[key] = value;
             }
@@ -706,6 +724,7 @@ export {
     mergeObjects,
     moveElementToEnd,
     queryStringParser,
+    hasOwnProp,
     getCookies,
     getHref,
     replaceMPID,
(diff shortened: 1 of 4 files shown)


src/utils.ts
```diff
@@ -334,7 +339,14 @@ const queryStringParser = (
         return lowerCaseUrlParams;
     } else {
         keys.forEach(key => {
-            const value = lowerCaseUrlParams[key.toLowerCase()];
+            const name = key.toLowerCase();
+
+            // Callers pass their own key list, so `name` may not be from this URL.
+            if (!hasOwnProp(lowerCaseUrlParams, name)) {
```
alexs-mparticle: Can we invert this guard to make it more readable? So that rather than return if it's negative we just assign if possitive?

alexs-mparticle (changes requested):

src/utils.ts
mattbodle: Done in 47a78ac8 — the own-property check and the truthiness check collapse into one positive condition, so the early return and the intermediate `value` const both go:

```ts
keys.forEach(key => {
    const name = key.toLowerCase();

    // Callers pass their own key list, so `name` may not be from this URL.
    if (hasOwnProp(lowerCaseUrlParams, name) && lowerCaseUrlParams[name]) {
        results[key] = lowerCaseUrlParams[name];
    }
});
```

Behaviour is unchanged — `&&` short-circuits in the same order the two separate checks ran in, so a key that is absent, or present with an empty value, is still skipped. `tsc` clean, jest 701, karma 1089 on ChromeHeadless.

While in here I also added a note above `hasOwnProp` recording why it is not `Object.hasOwn`, since Sonar raises `typescript:S6653` recommending exactly that on this line. It is ES2022: I measured it, and it fails to compile under this project's `lib: ["es5", "es6", "dom"]` with `TS2550`. Widening the lib against `target: es5` would emit a call missing from every browser before 2021 — in the same file that still carries a fallback for browsers with no `URLSearchParams`. Left the MINOR Sonar issue standing rather than take the suggestion; happy to suppress it instead if you'd prefer the gate clean.

rmi22186 (approved):

rmi22186:
updated PR into v3-development

PR #1382, opened 2026-08-28
feat(apv): let a customer remove a default query param, not just add

mattbodle:
> **Merge last of five.** Stacked on #1375 → #1373 → #1374, so the diff shown here includes all three until they land. Nothing in it is safe before #1374, which is what makes the default list small enough for removal to be the missing half rather than a nicety.

## Summary

Advanced Settings took **additions only**, so after #1374 the nine built-in params were a floor no customer could get under. A `-` prefix in the same field now removes one:

```
-ref, promo_code        →  drops `ref`, adds `promo_code`
```

## Why one field and not two

`QUERY_PARAM_NAME` is `/^[a-z0-9_][a-z0-9_.-]{0,63}$/` — it **forbids a leading hyphen**. So before this change every `-name` entry was rejected outright, which means:

- no stored config can already mean something else by it,
- no second `SettingTemplate` id, feature flag, validator or migration is needed,
- the existing "a leading hyphen is invalid" test is the thing that proves it, and it is deliberately flipped here.

## Semantics, and why each choice is the safe one

**Exclusions apply after additions.** `-ref, ref` and `ref, -ref` both remove. The outcome cannot depend on which side of the comma a customer wrote first — pinned by a test that asserts the two orderings are equal.

**Exclusions are not capped.** `MAX_CUSTOM_QUERY_PARAMS` exists to stop the attribute count growing; exclusions only ever shrink the list, so they cannot be the reason an event runs out of headroom.

**An exclusion naming something that is not a default is a no-op, not an error.** Rejecting it would turn every stored `-thatparam` into a warning the next time `ALLOWED_QUERY_PARAMS` changes underneath a config, and SDK versions in the wild lag the dashboard's list by months.

> The dashboard is permissive here too, for the mirror-image reason: its error channel **disables Save for the whole app-config dialog**, so a stored `-name` going stale would block a customer from saving unrelated settings. The cost of both sides being permissive is real and worth a reviewer's opinion — a typo'd `-referer` does nothing, quietly, in both places.

### 1. An exclusion merges pages that used to be distinct

`pageKey` walks the effective allowlist, so removing an entry changes the key of every page that carried it. Exclude `page` and `?page=2` keys the same as `?page=3` — the second view **dedups away and never fires**.

That is the customer's call to make, but it is not a cosmetic setting, so the tracker logs it at **`warning`** rather than verbose. It is the setting most likely to be blamed for missing page views by whoever did not know it was set.

### 2. `-` reaches past the validator, and that opened a log-echo hole

This is the one thing in the PR I would most want a second pair of eyes on.

The established rule for rejected entries is **positions, never names** — a rejected entry may be something pasted into the wrong field, so `password=hunter2` must not reach `Logger.warning`. `-` weakens the validator's reach: `-secret-value-abcdef` has a perfectly legal name behind the prefix, so it no longer fails validation. It arrives as an **exclusion**, and exclusions are named in a log.

Fixed by naming only the exclusions that actually hit the default list:

```ts
const applied = excluded.filter(
    name => ALLOWED_QUERY_PARAMS.indexOf(name) !== -1
);
```

Those names are **our own constants**, so that log echoes our words, not the customer's. An exclusion matching nothing is inert and stays silent. Pinned by `should never name an exclusion that matched no built-in`, which asserts the warning contains `ref` but not `secret` or `abcdef`.

The old rejection test that used `-secret-value-abcdef` as a leaked-prefix case is **moved rather than deleted** — it now covers this path instead, because under the new syntax it is no longer a rejection at all.

## Rename: `IQueryParamAllowlist` → `IQueryParamConfig`

A type that holds exclusions is not an allowlist. Alongside it, `allowed` → `added`, so `added`/`excluded` read as the pair they are, and `parseQueryParamAllowlist` → `parseQueryParamConfig`.

Callers now pass the **whole config object** rather than reaching in for one field:

```diff
-allowedQueryParams(getHref(), (getFeatureFlag(FLAG) as IQueryParamAllowlist)?.allowed)
+allowedQueryParams(getHref(), getFeatureFlag(FLAG) as IQueryParamConfig)
```

That is smaller at both call sites, and it removes the `extras: string[]` parameter that a reviewer flagged as under-named. The helper signatures take `Partial<IQueryParamConfig>`, which is honest: `getFeatureFlag` returns `null` for an absent flag and a self-hosted `config.flags` can pass an incomplete object.

## Verification

Measured on this branch's tip:

- `jest` — **811 pass / 0 fail**, up from **795** on #1374 (+16)
- `karma` ChromeHeadless — **1089 pass / 0 fail**, 10 skipped, after a fresh `npm run build && npm run build:test-bundle`
- `tsc -p . --noEmit` clean, `eslint src/ test/src/` exit 0

New coverage, by the rule it holds:

| rule | test |
| :-- | :-- |
| `-` means exclude, not add | `should read a '-' prefix as an exclusion rather than an addition` |
| one field carries both | `should split additions and exclusions out of one field` |
| exclusion beats addition, either order | `should let an exclusion beat an addition of the same name` |
| removal preserves survivor order | `should keep the remaining built-ins in order after an exclusion` |
| exclusions are uncapped | `should not count exclusions against the addition cap` |
| unknown exclusion is inert | `should accept an exclusion that matches no default` |
| `-` / `--x` are still rejected | `should reject a bare hyphen` / `a doubled hyphen` |
| the dedup-key consequence is real | `should drop an excluded param out of the key` |
| no customer text in the log | `should never name an exclusion that matched no built-in` |
| end to end through the tracker | `should stop capturing an excluded built-in` |

## The mPServer half

mparticle-by-rokt/mPServer#44037 carries it: `validateAutoLogPageViewQueryParams` accepts a `-` prefix, `normalize()` keeps exclusions verbatim, the help text and placeholder document the syntax, and the hand-copied `BUILT_IN_QUERY_PARAMS` shrinks to the same nine names #1374 leaves.

**The two must ship together.** This PR alone is inert — the UI would refuse to save a `-name`. That PR alone is worse than inert: the SDK re-validates remote config, so a stored `-ref` would be rejected as an illegal name and reported as a position warning, and the exclusion would silently do nothing.

That PR also fixes a silent failure this one exposes: the stale 24-name `BUILT_IN_QUERY_PARAMS` *drops* names it thinks are redundant, so after #1374 a customer typing `gclid` would have had it discarded as "you already have this" when the SDK no longer captures it — no error, no attribute.

### DIFF
--- a/kits/braze/braze-3/README.md
+++ b/kits/braze/braze-3/README.md
@@ -1,52 +1,116 @@
-![Braze Logo](https://github.com/mparticle-integrations/mparticle-javascript-integration-appboy/blob/master/braze-logo.png) 
+![Braze Logo](https://github.com/mparticle-integrations/mparticle-javascript-integration-appboy/blob/master/braze-logo.png)
 
-⚠️⚠️⚠️
-# Notice! Opt in is now available for Braze Web SDK V4 - Action Required
+# mParticle Braze Kit — Braze Web SDK V3
 
-You can now select what version of the Braze SDK you want to use when setting up a Braze connection in the mParticle UI.  Braze occasionally makes breaking changes to their SDK, so if you call `appboy` directly in your code, you will have to update your code to ensure your website performs as expected when updating versions of Braze.
+This kit bundles **Braze Web SDK V3** (`@braze/web-sdk@^3.5.0`) and exposes it on the page as `window.appboy`.
 
-Please review the [Braze Changelog](https://www.braze.com/docs/developer_guide/platform_integration_guides/web/changelog#400) and [V4 migration guide](https://github.com/braze-inc/braze-web-sdk/blob/master/UPGRADE_GUIDE.md) to learn about the differences between V3 and V4 and what changes you will need to make in your code. The most significant breaking changes are the replacement of the `appboy` class name with `braze`, in addition to the removal and renaming of several APIs.
+---
 
-You can opt into the latest major version of the Braze Web SDK whether you implement mParticle's Web SDK using npm or our snippet/CDN.
-* Customers who self-host mParticle via npm - You should add @mparticle/web-braze-kit version 4.0.0 or greater in your package.json now! You must also select `Version 4` under `Braze Web SDK Version`  in the Braze connection settings.
-* Customers who load mParticle via snippet/CDN - Opting in is available via the mParticle UI in your Braze connection settings.
+## ⚠️ We recommend upgrading to Braze Web SDK V6
 
-Note that the following is only one example.  Everywhere you manually call `appboy` needs to be updated similar to the below. If you are using NPM, you can skip to step 3.  Please be sure to test your site fully in development prior to releasing.
+Braze is now on **V6**, and V6 is the version mParticle recommends for all customers. We recommend updating straight to V6 for latest support of all Braze features.
 
+See the [Braze Event Integration](https://docs.mparticle.com/integrations/braze/event/) docs.
 
-* Step 1: Legacy code sample. Find all the places where your code references the `appboy.display` namespace.  Braze has removed all instances of the `display` namespace:
+### How to upgrade
+
+1. **Audit the code you own.** If you call Braze directly, find every `appboy` reference and compare it with the table below and Braze's upgrade documentation.
+2. **Move to the V6 kit.**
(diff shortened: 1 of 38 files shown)


mattbodle:
Superseded by #1477: the same change, rebuilt as one signed commit on the current `v3-development` and opened from a branch in this repository rather than a fork. Closing in favour of that one — thanks!

PR #1393, opened 2026-09-01
ci: publish v3 kits from staging step one

rmi22186:
## Summary
Relative to `main`, this PR makes staging Step 1 the single release publisher and extends its v3 path from publishing only `@mparticle/web-sdk` to publishing the core SDK and all 33 leaf packages in `kits/publish-matrix.json` at one lockstep version.

The workflow freezes the selected development branch to one immutable commit before validation. Every build, core test, integration test, kit test, dry-run preview, and release branch uses that exact commit. The legacy `.github/workflows/release.yml` publisher is removed.

The v2 path remains core-only.

## Normal v3 release flow
1. Step 1 validates the selected track and workflow source, resolves `v3-development` to one immutable candidate SHA, and records it in the workflow summary.
2. Core builds and tests, integration tests, matrix loading, and all kit tests run against that candidate SHA.
3. The release branch is created from the same candidate SHA. `v3-staging` must be its ancestor, the merge must fast-forward, and the resulting pre-release `HEAD` must equal the tested candidate.
4. Semantic-release calculates the next core version.
5. `verifyRelease` confirms an established v3 `next` channel is complete: every kit's `next` tag must equal the current core `next` version. The one-time `3.0.0` bootstrap permits all kit tags to be absent so the first synchronized release can establish the baseline. Channel rollback is also rejected.
6. `scripts/prepare-kit-release.js` validates the inventory, cleans every publishable output directory, and applies the target version to all publishable and aggregate manifests plus lockfiles.
7. `scripts/release.sh` builds core and 32 unique kit build roots.
8. Before core publication, all 33 newly built kit artifacts are packed and registry-preflighted. Existing target versions must be byte-identical and use the expected channel; conflicts stop the release.
9. Semantic-release commits release assets, creates the tag, and publishes core.
10. A v3-only job checks out the verified release tag, packs core and all kits, and publishes missing kits sequentially from those exact tarballs with npm OIDC provenance.
11. Byte-identical existing kits are skipped. Core, all kits, their dist-tags, and the release tag are audited after publication.

## Failure and recovery behavior
Core and 33 kits cannot publish atomically. If publication stops partway through, newer v3 releases remain blocked until recovery completes.

Recover by rerunning Step 1 from `v3-staging` with:
- `track=v3`
- `dryRun=false`
- `resumeKitReleaseTag=<exact Step 1 tag>`

Recovery treats the supplied immutable tag as authoritative, requires `v3-staging` to resolve to that commit, confirms core and existing kits match the expected artifacts, and publishes only missing kits. If core is absent, recovery stops with controlled tag/staging reconciliation instructions. Other failures identify kits not yet recorded successful and print the exact recovery inputs.

## Dry-run behavior
For `dryRun=true`:
- the development candidate is frozen to one SHA
- core, integration, and kit tests run against that SHA
- semantic-release predicts the shared version and notes from that candidate
- an isolated worktree versions and builds all 32 kit build roots
- all 33 kits are packed before registry preflight begins
- no branch, commit, tag, core package, or kit package is published

## Release safeguards
- staging must be an ancestor of the frozen candidate, and both preview and real merges are fast-forward-only
- the pre-release checkout must equal the tested candidate SHA before semantic-release begins
- released source derives exclusively from the tested candidate; generated release assets are built and preflighted before publication
- v2 uses npm dist-tag `latest`; v3 uses `next`
- Step 1 is the only repository release publisher
- after the initial kit-channel bootstrap, a newer v3 release is blocked until core and all kits are aligned on the current `next` version
- target kit artifacts are packed and preflighted before core publication
- package names, paths, versions, repository metadata, required files, and stable SemVer are validated
- every publishable output directory is cleaned before builds; nested Adobe packages also use explicit file allowlists
- all artifacts are packed before registry checks or npm mutation
- existing and published artifacts must match local SHA-512 integrity and the expected dist-tag
- kit publication honors workflow cancellation
- release tag SHA is checked before and after kit publication
- partial results are written to the workflow summary

## Rollout prerequisites
- merge paired workflow PR #1394 so `.github/workflows/staging-step-1.yml` remains identical on `main` and `master`
- synchronize the final Step 1 workflow to `staging` and `v3-staging`
- trusted publishing for core and all 33 kits is configured, per maintainer confirmation
- core currently has `next=3.0.0` and no kit has a `next` tag; this recognized bootstrap state allows the first synchronized release to establish all kit tags

## Validation performed
- [x] focused release-script Jest suite: 17\/17 passing
- [x] live npm preflight recognizes the `3.0.0` no-kit baseline and permits the initial synchronized release
- [x] local full previews built 32 roots and packed/preflighted all 33 kits
- [x] candidate and ancestry workflow parses as YAML and passes Prettier, whitespace, and IDE diagnostics
- [x] JavaScript syntax, Prettier, whitespace, and IDE diagnostics pass
- [x] prior SonarCloud quality gate passes; the remaining optional-chaining suggestion is addressed at current head


scripts/release.sh
```diff
@@ -1,17 +1,53 @@
...
+        process.stdout.write(inventory.buildPaths.join('\n'));
+    " > "$BUILD_PATHS_FILE"
+    while IFS= read -r KIT_PATH; do
+        [ -n "$KIT_PATH" ] || continue
+        echo "Installing dependencies for $KIT_PATH"
+        npm ci --prefix "$KIT_PATH"
+        rm -rf "$KIT_PATH/dist"
+        echo "Building $KIT_PATH"
+        npm run build --prefix "$KIT_PATH"
+    done < "$BUILD_PATHS_FILE"
```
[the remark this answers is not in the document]
rmi22186: Resolved in 66fb9f60. Build paths are now serialized with a trailing newline for every entry, so the POSIX loop consumes all 32 roots, including the final `kits/roktpayplus` path. The added regression test verifies both newline termination and preservation of the complete ordered path list.

.github/workflows/staging-step-1.yml
```diff
@@ -567,6 +668,7 @@ jobs:
             TRACK: ${{ inputs.track }}
             NPM_DIST_TAG: ${{ needs.confirm-staging-branch.outputs.npm_dist_tag }}
             NPM_CONFIG_TAG: ${{ needs.confirm-staging-branch.outputs.npm_dist_tag }}
+            NEXT_RELEASE_VERSION_FILE: ${{ runner.temp }}/next-release-version
```
alexs-mparticle: **HIGH — `runner` context is not available in a job-level `env:` block, so this workflow will not parse.**

GitHub's context-availability table permits only `github, needs, strategy, matrix, vars, secrets, inputs` in `jobs.<job_id>.env`. `${{ runner.temp }}` here is rejected at workflow-parse time with `Unrecognized named-value: 'runner'` — the whole workflow fails to queue, for both tracks and for dry runs.

Even in the lenient case where it resolves to empty, the path becomes `/next-release-version`, and `publish-kits.js --preflight-tags` (`scripts/publish-kits.js:602-607`) does an unguarded `fs.writeFileSync` there, which fails with `EACCES` as the `runner` user -> `verifyReleaseCmd` exits 1 -> semantic-release aborts every release.

This same job already uses `$RUNNER_TEMP` correctly inside a `run:` block at line 772 — that env var is the fix, or compute the path in a step.

.github/workflows/staging-step-1.yml
```diff
@@ -730,3 +873,104 @@ jobs:
...
+                  node-version: 24.19.0
+                  registry-url: 'https://registry.npmjs.org'
+
+            - name: Publish kits sequentially and audit all packages
+              # Preserve GitHub's authentic dispatch identity for npm OIDC.
+              # The checked-out release source is bound separately by the
+              # verified tag SHA and exact tarball integrity.
+              run: node scripts/publish-kits.js
+
+            - name: Reverify immutable release tag
```
alexs-mparticle, line 957: **MEDIUM — this `always()` step reads env vars that a failed predecessor never sets.**

The step runs with `if: ${{ always() }}` and `set -eu`, but reads `RELEASE_TAG` / `RELEASE_SHA`, which are only exported to `GITHUB_ENV` by the preceding "Verify immutable release source" step.

If the tag checkout or that verify step fails — the exact case `always()` exists to cover — both variables are unset and this step dies with `RELEASE_TAG: unbound variable`. The genuine failure then gets buried under a spurious shell error, and the tag-movement check this step exists to perform never actually runs.

Suggest guarding on `[ -n "${RELEASE_TAG:-}" ]` and skipping with an explicit message.

scripts/publish-kits.js
```diff
@@ -0,0 +1,643 @@
...
+
+    for (let index = 0; index < leftParts.length; index++) {
+        if (leftParts[index] !== rightParts[index]) {
+            return leftParts[index] > rightParts[index] ? 1 : -1;
+        }
+    }
+    return 0;
+}
+
+function assertDistTagWillNotMoveBackward(
```
alexs-mparticle, line 76: **LOW — the dist-tag rollback guard throws a misleading error, from the wrong layer, if a channel tag ever points at a non-stable version.**

`assertDistTagWillNotMoveBackward` passes the registry's current `dist-tags.<tag>` straight into `compareStableVersions`, which calls `validateVersion` on it (`:52`). If a prerelease is ever published to `next` (e.g. `3.1.0-rc.1`) — or to `latest` on the v2 track — `verifyReleaseCmd` aborts the release with `Expected a stable semantic version, received: 3.1.0-rc.1`, which reads like a bug in the *computed release version* rather than "the current channel tag is a prerelease".

`verifyCurrentReleaseComplete` has the same issue at line 132. Suggest comparing with semver, or failing with a message that names the offending dist-tag.

scripts/preview-kit-release.js
```diff
@@ -0,0 +1,107 @@
...
+const { validateVersion } = require('./prepare-kit-release');
+
+const repositoryRoot = path.resolve(__dirname, '..');
+const npmExecutable = path.join(
+    path.dirname(process.execPath),
+    process.platform === 'win32' ? 'npm.cmd' : 'npm'
+);
+
+function runGit(args) {
+    execFileSync('/usr/bin/git', args, {
```
alexs-mparticle, line 16: **LOW — `runGit` hardcodes `/usr/bin/git`.**

Fine on `ubuntu-latest` and on macOS with Xcode CLT, but `node scripts/preview-kit-release.js 3.1.0` fails with `ENOENT` for anyone whose git is elsewhere (Nix, Homebrew-only on Apple Silicon, Windows) — and it surfaces as a bare `spawnSync /usr/bin/git ENOENT`, not "git not found".

The sibling script resolves `npm` relative to `process.execPath` precisely to avoid PATH assumptions; `git` from PATH (or a `GIT_EXEC` override) would be consistent.

Note the new Jest suite does not catch this: `previewKitRelease('3.0.1', 'latest')` throws on the dist-tag check before `runGit` is reached.

alexs-mparticle (commented):
Code review findings (4). One blocker, one medium, two low.

Verified as sound, no re-review needed: the recovery-mode job-skip chain, `candidate_sha` freezing plus the staging-ancestor/fast-forward assertions, the `24.x` -> `24.19.0` pin that makes pack-integrity equality reproducible, `rm -rf dist` in `release.sh` (BUILDALL regenerates all tracked artifacts), and the new `kits/**/package-lock.json` git asset not sweeping in the 32 kit `node_modules` trees. Also checked against the live registry: all 33 kits validate, all have `latest`, none has `next`, and no kit has a `prepare`/`prepack`/`prepublishOnly` script or `publishConfig.tag` that would break the `npm pack --ignore-scripts` -> `npm publish <tarball>` integrity equality this design depends on.

scripts/publish-kits.js
rmi22186: Resolved in 81df7b79. Stable-version validation now reports the package name, offending dist-tag, and prerelease value. The stable-only channel policy remains intentional, and the regression suite covers the contextual error.

.github/workflows/staging-step-1.yml
rmi22186: Resolved in 81df7b79 (and mirrored to PR #1394 in 50832030). The `always()` step now guards `${RELEASE_TAG:-}` and `${RELEASE_SHA:-}` and exits with an explicit skip message when immutable-source verification did not complete, preserving the original failure.

.github/workflows/staging-step-1.yml
rmi22186: Resolved in 81df7b79 (and mirrored to the paired master workflow in 50832030). `NEXT_RELEASE_VERSION_FILE` is no longer evaluated from job-level `env`; a step now exports the `$RUNNER_TEMP` path through `GITHUB_ENV`, where runner initialization is available.

scripts/preview-kit-release.js
rmi22186: Not changing this in the current PR. The preview targets the supported Ubuntu GitHub runner and macOS development environment, where `/usr/bin/git` is available, and the explicit executable path intentionally avoids the PATH-resolution security warning previously raised by Sonar. We can add a validated override if Nix or Windows preview support becomes a requirement.

alexs-mparticle (approved):

PR #1396, opened 2026-09-03
fix(onetrust): make async test failures observable

rmi22186:
## Summary
- wait for Web SDK v3 identity initialization before asserting OneTrust consent state
- route identity requests through the existing XHR mock and replace the millisecond-sensitive consent assertion
- propagate Karma's exit code so browser test failures fail CI

## Test plan
- [x] Run the OneTrust ChromeHeadless and FirefoxHeadless suite three consecutive times (10/10 passing each run)
- [x] Run the core browser test suite (2,142 passing)
- [x] Run Prettier and `git diff --check`

### DIFF
--- a/kits/onetrust/test/boilerplate/test-karma.js
+++ b/kits/onetrust/test/boilerplate/test-karma.js
@@ -9,4 +9,5 @@ var testCommand = util.format(
 );
 
 console.log('Running command: \n' + testCommand);
-shell.exec(testCommand);
+var result = shell.exec(testCommand);
+process.exitCode = result.code;
--- a/kits/onetrust/test/tests.js
+++ b/kits/onetrust/test/tests.js
@@ -58,70 +58,86 @@ function configureOneTrustForwarderAndInit(
     consentGroups,
     vendorIABConsentGroups,
     vendorGoogleConsentGroups,
+    vendorGeneralConsentGroups,
+    identityCallback
+) {
+    mParticle.config = {
+        requestConfig: false,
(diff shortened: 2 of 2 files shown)


kits/onetrust/test/boilerplate/test-karma.js
crisryantan, line 9: Do we want to change these to ts files now that we are in v3?

rmi22186: @crisryantan - I think a wholesale change would be good in 1 go, but 
this PR is limited in scope to just make failed tests exist properly

kits/onetrust/test/tests.js
```diff
@@ -317,29 +325,35 @@ describe('OneTrust re-initialization tests', function() {
...
+                window.OnetrustActiveGroups = ',1,2,4,userEditable2';
+
+                // When we re-initialize consent, the strictly necessary consent has not been updated
+                // because there is no change in consent
+                configureOneTrustForwarderAndInit(
+                    consentGroups,
+                    null,
+                    null,
+                    null
+                );
```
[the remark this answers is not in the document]
rmi22186: Valid for the original implementation. Commit c7101deb separates asynchronous first-time identity initialization from synchronous kit reinitialization. Awaiting the second identity callback is not safe because reinitialization does not necessarily make another identity request, so that callback may never fire; the second call therefore no longer returns an ignored Promise.

crisryantan (approved):

PR #1405, opened 2026-09-08
fix(rokt): stop emitting ATTRIBUTE_SETTER_CALLED diagnostic logs

mattbodle:
## Background

- `ATTRIBUTE_SETTER_CALLED` was added on 2026-08-31 in #1386 (`65bc3d39`, Alexander Sapountzis) as a reland of closed `mparticle-javascript-integration-rokt` #130 (`302291a`, 2026-08-28).
- It is an INFO diagnostic fired on every `setUserAttribute`, `removeUserAttribute`, `onUserIdentified`, `onLoginComplete`, `onLogoutComplete`, and `onModifyComplete` call, logging attribute *keys* only (never values) so downstream could correlate setter timing with `selectPlacements` from timestamps and page URL.
- In production it is flooding `ecommerce-diagnostics-api` (`reporter: mp-wsdk`, kit 3.1.0) — on the order of tens of millions of INFO logs per week.

## What Has Changed

- Stop emitting `ATTRIBUTE_SETTER_CALLED` from the Rokt kit setter/identity paths.
- Remove `buildSetterDiagnosticLogEntry` and the dedicated setter diagnostic rate-limit transport that existed only to keep this flood from starving other INFO logs.
- Keep `SELECT_PLACEMENTS_DISPATCHED` on its own transport.

## Screenshots/Video

- N/A

## Checklist

- [ ] Self-review completed
- [ ] Tests added or updated
- [ ] Tested locally


kits/rokt/src/diagnosticTiming.ts
```diff
@@ -1,23 +1,12 @@
...
 
 export interface DiagnosticLogEntry {
   message: string;
   code: string;
 }
 
-export function buildSetterDiagnosticLogEntry(source: string, attributeKeys: string[]): DiagnosticLogEntry {
-  return {
-    message: `Rokt Kit: ${source} called [attributeKeys=${attributeKeys.join(',')}]`,
-    code: 'ATTRIBUTE_SETTER_CALLED',
```
alexs-mparticle, line 17: We can probably also remove `SELECT_PLACEMENTS_DISPATCHED` as well.

mattbodle: Done — removed `SELECT_PLACEMENTS_DISPATCHED`, `diagnosticTiming.ts`, and `logPlacementDiagnostic` so the kit no longer emits either timing diagnostic.

alexs-mparticle (approved):

PR #1413, opened 2026-09-10
fix(release): wait for core npm visibility before publishing kits

rmi22186:
## Background

- Staging Step 1 can publish core, then fail kit publish on the first `npm view` miss while the registry is still catching up. That is what happened on [run 34530918523](https://github.com/mParticle/mparticle-web-sdk/actions/runs/34530918523/job/103054484467) for `@mparticle/web-sdk@3.2.0`.
- [#1397](https://github.com/mParticle/mparticle-web-sdk/pull/1397) already waits up to five minutes for the post-publish kit audit. That retry does not cover the pre-kit core visibility gate.

## What Has Changed

- Retry core npm visibility for up to five minutes (60 attempts, 5s apart) before publishing kits.
- Fail immediately on integrity mismatches so a wrong artifact does not sit in the retry loop.
- Keep recovery's `CORE_NOT_PUBLISHED` path, but only after the same wait.
- Document the pre-kit wait in the Step 1 release notes.

## Screenshots/Video

- N/A

## Checklist

- [ ] Self-review completed
- [ ] Tests added or updated
- [ ] Tested locally

## Test plan

- [ ] `npm run test:jest -- --runTestsByPath test/jest/release-scripts.spec.ts`
- [ ] Confirm a later v3 Step 1 kit job logs retry attempts if core is not visible immediately, instead of failing in ~40s

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved package release reliability by waiting for the core package to become visible on npm before publishing related packages.
  * Added retry handling for temporary publication delays and immediate detection of integrity mismatches to prevent unsafe releases.
  * Improved recovery when package metadata or integrity information is temporarily unavailable.

* **Documentation**
  * Updated release workflow documentation to describe the npm visibility check and waiting period before package publication.

PR #1414, opened 2026-09-11
perf(release): publish v3 kits concurrently after core is visible

rmi22186:
## Summary
- Follows [#1413](https://github.com/mParticle/mparticle-web-sdk/pull/1413): after core is visible on npm, kits do not depend on each other, so sequential publish is not required.
- Keeps `waitForPublishedCore` serial, then publishes kits with a bounded worker pool (default concurrency **6**, overridable via `options.concurrency`).
- Attempts every independent kit even when another publish fails, records successful in-flight work for recovery, and reports all failures together after the pool drains. Preflight-all-before-any-publish, provenance, and integrity-mismatch fail-fast are unchanged.
- Buffers each concurrent `npm publish` process output and emits an attributed GitHub log group so package logs do not interleave.
- Updates the Step 1 workflow name and README so they no longer say kits are published sequentially. Rokt-first matrix order is left as-is.

## Test plan
- [x] `npx jest --config jest.config.js --runTestsByPath test/jest/release-scripts.spec.ts --no-coverage` (28 passed)
- [x] `npx prettier --check scripts/publish-kits.js test/jest/release-scripts.spec.ts`
- [x] `npm run lint -- --no-fix`
- [ ] Confirm a future v3 Step 1 run still waits for core, then overlaps kit `npm publish --provenance` up to 6 at a time
- [ ] Confirm multiple kit failures are reported together and recovery skips already-published identical artifacts

## Summary by CodeRabbit

- **New Features**
  - Kit packages are now published concurrently with a bounded worker pool.
  - Publishing tracks completed kits and stops starting new publishes after a failure.
  - Invalid concurrency settings are rejected before publishing begins.

- **Bug Fixes**
  - Recovery guidance now covers identifying and correcting all missing or rejected kits, rather than only the first failure.

- **Documentation**
  - Release workflow documentation and status descriptions now reflect concurrent kit publishing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,7 +155,8 @@ workflows in order:
    registry-preflights all kit artifacts before publishing core. It then checks
    out the immutable release tag, waits up to five minutes for core to become
    visible on npm, and publishes all packages in `kits/publish-matrix.json`
-   sequentially. After all publishes finish, one final audit waits up to five
+   concurrently with a bounded worker pool. After all publishes finish, one
+   final audit waits up to five
    minutes for npm propagation and verifies that
    the core SDK and every kit have the expected version, artifact integrity,
    and npm dist-tag. A rerun skips an existing kit only when its artifact is
@@ -197,8 +198,8 @@ A real v3 Step 1 run can publish core and create the GitHub Release before all
 
 1. Do not start a newer release or run Step 2 or Step 3.
 2. Copy the exact release tag from the failed Step 1 run.
-3. Use the failed job and npm package audit summary to identify the first
-   missing or rejected kit. Correct its npm or trusted-publisher configuration.
+3. Use the failed job and npm package audit summary to identify any
+   missing or rejected kits. Correct their npm or trusted-publisher configuration.
 4. Dispatch **Staging Release - Step 1** from `v3-staging` with `track=v3`,
    `dryRun=false`, and `resumeKitReleaseTag` set to the failed run's exact tag.
 5. Wait for recovery to verify core and all 33 kits at that version on `next`.
--- a/scripts/publish-kits.js
+++ b/scripts/publish-kits.js
@@ -3,7 +3,7 @@
 const fs = require('node:fs');
 const os = require('node:os');
 const path = require('node:path');
-const { execFileSync, spawnSync } = require('node:child_process');
+const { execFileSync, spawn, spawnSync } = require('node:child_process');
 const {
     loadReleaseInventory,
     validateVersion,
@@ -14,6 +14,7 @@ const registry = 'https://registry.npmjs.org';
 const initialV3CoreVersion = '3.0.0';
 const remoteAuditAttempts = 60;
 const remoteAuditDelayMs = 5000;
+const kitPublishConcurrency = 6;
 const npmExecutable = path.join(
     path.dirname(process.execPath),
     process.platform === 'win32' ? 'npm.cmd' : 'npm'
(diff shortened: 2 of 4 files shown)


mattbodle:
The concurrency implementation looks correct to me, and the `firstError` fix in `9e745c2` behaves the way you'd want: because `Promise.all` waits for every in-flight worker, kits that succeed *after* the first failure still land in `results`, so the skip-identical recovery path sees them. Letting in-flight publishes finish rather than killing them is also the right call — a `SIGTERM` mid-`npm publish` could leave a version half-uploaded with no local record.

One thing I'd like to raise before this runs for the first time: **there is no way to exercise the concurrent path short of a real release, and a transient failure is indistinguishable from a real one.**

The `publish-kits` job is gated on `dryRun == 'false'`, and the dry-run path only reaches `--preflight-artifacts`, which packs and `npm view`s without ever calling `publishTarball`. So concurrent publishing gets its first exercise on a live 33-package release.

That is worth weighing against what each concurrent publish actually does. The job runs with `id-token: write` and no `NODE_AUTH_TOKEN`, so this is trusted publishing: every in-flight `npm publish --provenance` independently requests a GitHub Actions OIDC token, exchanges it with npm, and requests a second token for the Sigstore attestation. Six of those in parallel is six times the concurrent load on three services, and as far as I can tell we have no data on how they behave under that pattern.

If any one of them returns a 429, a 5xx, or a socket timeout, fail-fast aborts and you are in the README recovery procedure on a partially published release. `child.on('close')` rejects on any non-zero exit, so there is nothing distinguishing a network blip from a genuine npm or trusted-publisher misconfiguration — and the recovery cost is the same either way.

The test plan on this PR leaves exactly this unchecked:

> - [ ] Confirm a future v3 Step 1 run still waits for core, then overlaps kit `npm publish --provenance` up to 6 at a time

Two suggestions, either of which would cover it:

**1. Retry once on a transient before failing fast.** Capture the child's stderr and re-spawn once for `E429` / `ETIMEDOUT` / `EAI_AGAIN` / 5xx. This pairs well with switching `stdio` from `inherit` to `pipe`, which is worth doing regardless — six concurrent publishes currently interleave their output into one stream with no per-kit attribution, which makes README step 3 ("identify any missing or rejected kits") harder than it was when publishing was sequential. Buffering per child and emitting one `::group::<kit>` block on close would fix both.

**2. Land at a smaller concurrency and raise to 6 after one clean production run.** Most of the win is available well before 6. Modelling 33 kits at 12s per publish and 1s per `npm view`, driven through the real `runWithBoundedConcurrency`:

| concurrency | modelled wall clock | speedup |
| --- | --- | --- |
| 1 (current behaviour) | ~431s | — |
| 2 | ~222s | 1.9x |
| 3 | ~146s | 2.9x |
| 6 (this PR) | ~80s | 5.4x |

Concurrency 3 gets you a bit over half the total win at half the concurrent load on the OIDC and attestation path, and the constant is a one-line change once there is a real run to point at.

Neither is a blocker on the code as written — the logic is sound. It is the unrehearsable first run I would want a margin on.

README.md
```diff
@@ -159,7 +155,8 @@ workflows in order:
    registry-preflights all kit artifacts before publishing core. It then checks
    out the immutable release tag, waits up to five minutes for core to become
    visible on npm, and publishes all packages in `kits/publish-matrix.json`
-   sequentially. After all publishes finish, one final audit waits up to five
+   concurrently with a bounded worker pool. After all publishes finish, one
```
jamesnrokt, line 158: Pre-exisitng but wondering if we should move the entire release process to a dedicated RELEASING.md like we have on [UX Helper](https://github.com/ROKT/rokt-ux-helper-ios/blob/main/RELEASING.md)? Would remove the info that most people landing on the page don't care about ie they're not maintainers

scripts/publish-kits.js
```diff
@@ -467,29 +468,94 @@ function publishTarball(packageInfo, version, distTag) {
...
+            }
+            try {
+                await workerFn(items[index], index);
+            } catch (error) {
+                if (!firstError) {
+                    firstError = error;
+                }
+            }
         }
+    }
```
jamesnrokt, line 538: I believe we should attempt every kit even if one fails, same as we handle CocoaPods.

<details><summary>Why</summary>

`while (!firstError)` (publish-kits.js:525) stops workers from starting any new item once a failure is seen, so kits that haven't started yet never run — confirmed by the "stops starting new kit publishes after a failure" test (8 kits / concurrency 3 / kit-1 fails → only kits 1-3 start). Would need to drop that gate, collect all errors, and decide at the end whether to throw.
</details>

scripts/publish-kits.js
```diff
@@ -467,29 +468,94 @@ function publishTarball(packageInfo, version, distTag) {
...
+
+function resolvePublishConcurrency(concurrency) {
+    const resolved = concurrency == null ? kitPublishConcurrency : concurrency;
+    if (!Number.isInteger(resolved) || resolved < 1) {
+        throw new Error(
+            `Kit publish concurrency must be a positive integer, received ${concurrency}`
+        );
+    }
+    return resolved;
+}
```
jamesnrokt, line 514: nit: no test covers the throw path in resolvePublishConcurrency — could we clamp to a minimum of 1 instead of validating and throwing?

README.md
rmi22186: Agreed this would make the maintainer-only release guidance easier to find, but it is a pre-existing documentation restructure rather than part of concurrent publishing. I’ll keep `RELEASING.md` as a separate follow-up so this PR stays focused.

scripts/publish-kits.js
rmi22186: Agreed and implemented in `723ec4ae`. The bounded workers now attempt every independent kit, retain successful results, and report all publish failures together after the pool drains. Concurrent child output is also buffered into per-package GitHub log groups so each failure remains attributable.

scripts/publish-kits.js
rmi22186: I’m keeping strict validation here: silently clamping `0`, a negative value, or a fraction to serial publishing would hide a release configuration error and unexpectedly slow the run. `resolvePublishConcurrency` executes before preflight or publication, so failing loudly is cheap and safe. Added coverage for `0`, `-1`, and `1.5` in `723ec4ae`.

rmi22186:
Thanks — I adopted the log-attribution part in `723ec4ae`: each publish now pipes and buffers its output, then emits one package-specific GitHub log group instead of interleaving six children.

For failure handling, I followed the independent-package model from the review: all kits are attempted and all failures are reported together after the bounded pool drains. I’m not adding transient retry in this PR because safe retry also needs an idempotent registry check for the case where npm accepted a publish but the client lost the response; stderr classification alone is not sufficient. That can be handled as a focused follow-up. I’m also retaining the bounded concurrency of 6 for this change rather than combining the behavior update with tuning; we can lower or raise it based on the first production timing and service behavior.

PR #1416, opened 2026-09-11
feat(snippet): load Web SDK v3 from the CDN snippets

rmi22186:
## Summary
- Point `snippet.js` and `snippet.rokt.js` at `/js/v3/` with `snippetVersion` 3.0, and rebuild both minified files with uglify.
- Keep the full Rokt preload stubs (`selectPlacements`, `hashAttributes`, `hashSha256`, `setExtensionData`, `use`, `getVersion`, `terminate`, `onShoppableAdsReady`) on the mParticle snippet so a customer who later adopts Rokt can queue those calls before the SDK loads. `Identity.search` was already stubbed and remains in the minified output.
- Add an `onerror` fallback host to the Rokt snippet, since ad blockers commonly target the primary one. The hostname is assembled from fragments so blocklists cannot match it in this file.
- Invoke the Rokt IIFE with the `API_KEY` identifier instead of a placeholder string, so `snippet.rokt.min.js` can be pasted into the docs site as-is. `snippet.js` keeps `'REPLACE WITH API KEY'` because it is a standalone drop-in with no surrounding declaration.

Docs counterpart: ROKT/rokt-docs#1323 consumes `snippet.rokt.min.js` verbatim as the shared loader.

## Test plan
- [x] Confirm `npm run build:snippet` reproduces the committed min files (`uglifyjs -m` output is byte-identical to `snippet.min.js` and `snippet.rokt.min.js`)
- [x] Spot-check `snippet.min.js` loads `jssdkcdns.mparticle.com/js/v3/<key>/mparticle.js`
- [x] Spot-check `snippet.rokt.min.js` loads `/js/v3/<key>/app.js`, and that a blocked primary host falls back to `apps.roktecommerce.com`
- [x] Run snippet tests (`test/snippet/tests-snippet.js`) against both min files: 6 passing, including `Identity.search` and Rokt stubs `getVersion`, `terminate`, and `onShoppableAdsReady`
- [x] Confirm the README minified IIFE matches `snippet.min.js`

## Summary by CodeRabbit

- **New Features**
  - Updated script-tag initialization to load SDK version 3.
  - Added support for `Identity.search` and `Rokt.onShoppableAdsReady`.
  - Added fallback loading when the primary Rokt script domain is unavailable.

- **Changes**
  - Initialization now requires a page-provided `API_KEY`.
  - Updated queued method handling and script-loading paths for version 3.

- **Removed**
  - Removed `eCommerce.logCheckout` from queued API methods.


snippet.rokt.js
```diff
@@ -117,15 +117,41 @@
...
+            '&' +
+            versionQueryArray.join('&');
+        var fallbackAnchor = document.getElementsByTagName('script')[0];
+        fallbackAnchor.parentNode.insertBefore(fallbackScript, fallbackAnchor);
+    };
+
     var firstScript = document.getElementsByTagName('script')[0];
     firstScript.parentNode.insertBefore(script, firstScript);
-})('REPLACE WITH API KEY');
+    // Callers must declare API_KEY in the surrounding page (docs use const API_KEY = "...").
```
jaissica12, line 159: nit: can snippet.rokt.js keep })('REPLACE WITH API KEY') like snippet.js / README? })(API_KEY) throws if API_KEY is not already in scope, so existing pastes and <script src> break. Docs can still wrap with const API_KEY = "..." without changing the snippet.

jaissica12 (changes requested):
LGTM overall, /js/v3/ + snippetVersion 3 on both snippets, README matches snippet.min.js, extra Rokt stubs look right.

One small nit on snippet.rokt.js, can you please fix failing Prettier on the same file (the fallback .join('') wrap)

snippet.rokt.js
rmi22186: For snippet.rokt.js - we should have `API_KEY` remain because it is in-lined code, and not the full script that a developer pastes - and there will be snippet code above it where const API_KEY exists that a developer fills in.  but it's a good note about if there's a better way to do this.  It has to be done this way because we minify snippet.rokt.js to snippet.rokt.min.js, and there's a bunch of configuration above it that does not get minified (ie, all the stuff above line 41 in the Rokt docs as shown below)
image

jaissica12: That sounds right, thanks for clarifying!

jaissica12 (approved):
LGTM 🚀

PR #1421, opened 2026-09-14
fix(rokt): recover a not-ready preselect dispatch across page navigation

crisryantan:
## Summary
- `_preselectState` is in-memory only, so a full navigation before the launcher attaches drops a queued preselect dispatch along with that page's JS context
- Persist a best-effort resolved snapshot to `localStorage` at enqueue time (same namespaced-field + TTL pattern as `activePreselectStorage`), and replay it from whichever page's kit instance becomes ready next
- Independent of pathname on purpose, since surviving a navigation to a different page is the case this recovers

## Test plan
- [x] `npm test` (410 tests passing)
- [x] `npm run build`

## Summary by CodeRabbit

- **New Features**
  - Preselection requests can now persist briefly across navigation and recover automatically when the experience becomes ready.
  - Recovery is limited to matching accounts, identities, and user IDs, helping prevent cross-user or duplicate dispatches.
  - User attributes are captured and reused when available; incomplete requests remain queued until they can be processed.
  - Persisted requests expire automatically and can be cleared for an account.

- **Tests**
  - Expanded coverage for persistence, expiration, recovery, validation, navigation, queue handling, and duplicate prevention.


kits/rokt/src/preselection.ts
```diff
@@ -0,0 +1,228 @@
...
+    return;
+  }
+
+  clearPendingPreselect(host.accountId);
+
+  if (!host.isPreselectionEnabled() || !hasValidIdentity(host.filteredUser)) {
+    return;
+  }
+
+  fireDispatch(host, host.accountId, persisted.pathname, persisted.identifier, persisted.attributes, 'recovered');
```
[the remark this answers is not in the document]
crisryantan: Good catch, fixed in e064c26b. Persisted records now carry the mpid of whoever was signed in when they were written, and `maybeFirePersistedPreselect` discards rather than fires if the current user doesn't match at recovery time.

kits/rokt/src/preselection.ts
```diff
@@ -0,0 +1,228 @@
...
+
+  if (!host.isKitReady()) {
+    enqueuePending(state, { event, pathname });
+    host.logPlacementDiagnostic(buildPreselectDiagnosticLogEntry('queued', 'not_ready'));
+
+    // The kit not being ready is an infra-readiness race, not an attribute problem — a
+    // shopper this far along usually already has everything the config asks for. Snapshot
+    // it now, while the page (and this event) is still alive, so a full navigation away
+    // before the launcher attaches doesn't lose the attempt along with this JS context.
+    if (host.accountId && hasValidIdentity(host.filteredUser)) {
```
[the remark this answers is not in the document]
crisryantan: Deferred for now: isPreselectionEnabled() reads this.launcher.enablePreselection, which is only ever set once the launcher attaches, so there's no enablement signal available at the exact moment we're in the not-ready branch to gate the write on. Recovery already fails safe today (checked, doesn't dispatch, clears the record) if preselection turns out disabled by the time it's read. Tracking a last-observed enablement value across page loads to avoid the write entirely would need new persistent state of its own; happy to take that as a follow-up if the team wants the stronger guarantee.

kits/rokt/src/Rokt-Kit.ts
```diff
@@ -341,7 +353,7 @@ function generateThankYouElementScript(domain: string | undefined) {
 }
 
 function generateBaseUrl(domain: string | undefined) {
-  const resolvedDomain = typeof domain !== 'undefined' ? domain : DEFAULT_ROKT_DOMAIN;
+  const resolvedDomain = domain !== undefined ? domain : DEFAULT_ROKT_DOMAIN;
```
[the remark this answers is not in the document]
crisryantan: This is on Rokt-Kit.ts, which isn't part of this PR's diff (the branch got rebased onto current v3-development after it went stale) — looks like a leftover from before that. Not something this PR touches.

kits/rokt/src/preselection.ts
```diff
@@ -69,6 +99,56 @@ export function dispatchPreselect(host: PreselectHost, options: Record<string, u
...
+  setActivePreselect(activePreselectKey, attributes);
+  host.logPlacementDiagnostic(buildPreselectDiagnosticLogEntry('fired', reason));
+  dispatchPreselect(host, { attributes, preselect: true, identifier, omitUrl: true });
+}
+
+// Recovers a preselect attempt that resolved but couldn't dispatch before the page that
+// queued it went away. Independent of the current pathname on purpose: that's the case
+// being recovered (checkout to its confirmation page), not a reason to discard it. Only
+// fires for the same user who was signed in when it was persisted.
+export function maybeFirePersistedPreselect(host: PreselectHost): void {
```
alexs-mparticle: This feels like we're just adding more functions that do _similar_ preselect logic rather than fixes any inherent bugs within the existing logic.

jamesnrokt (approved):

kits/rokt/src/preselection.ts
crisryantan: Fair, folded it in afc4bbe6: maybeFirePersistedPreselect wasn't called from anywhere except flushPendingPreselectDispatches, so it's now a private step of that flush rather than a second exported "maybe fire" function. The bug itself is real though: the existing queue is a plain in-memory array, gone the moment the page reloads, and that's the majority of lost preselect volume in prod today.

crisryantan: It's not duplicating existing logic though: maybeFirePreselect's decision tree only ever runs against state.pending, an in-memory array that's gone the moment the page reloads, so it structurally can't recover anything across a navigation. maybeFirePersistedPreselect is the piece that makes that recovery possible at all, not a second copy of the same path.

alexs-mparticle (approved):

kits/rokt/src/preselection.ts
```diff
@@ -99,6 +129,56 @@ export function dispatchPreselect(host: PreselectHost, options: Record<string, u
...
+
+  clearPendingPreselect(host.accountId);
+
+  if (!host.isPreselectionEnabled() || !hasValidIdentity(host.filteredUser)) {
+    return;
+  }
+
+  if (getUserId(host.filteredUser) !== persisted.mpid) {
+    return;
+  }
```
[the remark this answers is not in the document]
crisryantan: Good catch, real bug. Fixed in f1db37bf: no longer clears the record until identity is confirmed valid, so if the launcher attaches before mParticle's identity resolution finishes, the record survives for a later flush (onUserIdentified) to pick up instead of being deleted and lost.

kits/rokt/src/preselection.ts
```diff
@@ -122,74 +185,68 @@ export function maybeFirePreselect(
...
+
+  fireDispatch(host, host.accountId, persisted.pathname, persisted.identifier, persisted.attributes, 'recovered');
 }
 
 export function flushPendingPreselectDispatches(
   state: PreselectState,
   host: PreselectHost,
   currentPathname: string = window.location.pathname,
 ): void {
+  recoverPersistedPreselect(host);
```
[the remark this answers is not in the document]
crisryantan: Good catch, real gap, fixed in 64170e09. Rather than clearing the guard entirely, used state.pending itself as the signal: a full navigation is what wipes it, so if it still has an entry for the persisted record's pathname, that's proof this is a same-instance SPA route change, not a cross-page recovery, and maybeFirePersistedPreselect now defers to that entry's own pathname check instead of firing over it.

kits/rokt/src/preselection.ts
```diff
@@ -99,6 +129,71 @@ export function dispatchPreselect(host: PreselectHost, options: Record<string, u
...
+
+  setActivePreselect(activePreselectKey, attributes);
+  host.logPlacementDiagnostic(buildPreselectDiagnosticLogEntry('fired', reason));
+  dispatchPreselect(host, { attributes, preselect: true, identifier, omitUrl: true });
+}
+
+// Recovers a preselect attempt that resolved but couldn't dispatch before the page that
+// queued it went away. Independent of the current pathname on purpose: that's the case
+// being recovered (checkout to its confirmation page), not a reason to discard it. Only
+// fires for the same user who was signed in when it was persisted.
```
rmi22186, line 158: The SPA vs full-navigation split is now handled, but a **full** navigation still fires `persisted.identifier` (`prod.rokt.conf`) with `omitUrl: true` on whatever page becomes ready next — home, event list, a new checkout, another tab.

Checkout → confirmation is one funnel; checkout → abandon is the common one. Two minutes of origin-wide storage is not a redirect buffer. If the product ask is confirmation recovery, gate on a destination pathname / identifier allowlist (or `sessionStorage`, which is tab-scoped and still survives a same-tab full load). As written, surviving a navigation to a *different* page is also surviving a navigation to the *wrong* page.

rmi22186 (commented):
Leftover findings that are not already on this thread (identity-clear, SPA vs full-nav pathname deferral, enablement-at-write, and mpid binding are covered).

kits/rokt/src/preselection.ts
```diff
@@ -113,6 +208,16 @@ export function maybeFirePreselect(
...
     enqueuePending(state, { event, pathname });
     host.logPlacementDiagnostic(buildPreselectDiagnosticLogEntry('queued', 'not_ready'));
+
+    // Not-ready is an infra-readiness race, not an attribute problem, so this usually
+    // resolves. Snapshot it now so a full navigation away doesn't lose it with this page.
+    const mpid = getUserId(host.filteredUser);
+    if (host.accountId && mpid && hasValidIdentity(host.filteredUser)) {
+      const { collected, missingKeys } = collectAttributes(host, event, configEntry);
+      if (missingKeys.length === 0) {
+        setPendingPreselect(host.accountId, pathname, configEntry.targetPageIdentifier, collected, mpid);
```
rmi22186, line 218: `collected` is written into `mp-rokt-kit` here. The only live `PRESELECTION_CONFIG` includes `email`, `firstname`, `lastname`, and `billingzipcode`.

`billingzipcode` and `currency` are on `SELECT_PLACEMENTS_ATTRIBUTE_PERSISTENCE_DENY_LIST`, so they never live on `this.userAttributes` — but `collectAttributes` still takes them from the event / `getAllUserAttributes()` and this snapshot persists them.

Please either strip deny-listed (and identity) keys before `setPendingPreselect`, or document an explicit exception that this partner’s preselect snapshot is allowed to sit in `localStorage`.

kits/rokt/src/preselection.ts
```diff
@@ -99,6 +129,71 @@ export function dispatchPreselect(host: PreselectHost, options: Record<string, u
...
+    return;
+  }
+
+  clearPendingPreselect(host.accountId);
+
+  if (getUserId(host.filteredUser) !== persisted.mpid) {
+    return;
+  }
+
+  fireDispatch(host, host.accountId, persisted.pathname, persisted.identifier, persisted.attributes, 'recovered');
```
rmi22186, line 194: `fireDispatch` keys `activePreselect` on `persisted.pathname` (checkout), even though this recovery is intentionally pathname-independent.

A recovered fire therefore stamps a 60s skip for `/checkout`. Shopper lands on confirmation (recovered), then returns to checkout within a minute for another order: the live preselect no-ops as `active_preselection`.

If the fire is not bound to the current route, the active-preselect identity should not be the *source* checkout path.

kits/rokt/src/pendingPreselectStorage.ts
```diff
@@ -0,0 +1,62 @@
+import { readNamespacedField, writeNamespacedField, removeNamespacedField, LS_NAMESPACE_KEY } from './storage';
+import { isObject, isString } from './utils';
+
+// Covers a checkout-to-confirmation redirect; short enough that a stale, unconsumed entry
+// doesn't get replayed long after the shopper is gone.
+export const PENDING_PRESELECT_TTL_MS = 2 * 60_000;
```
rmi22186, line 6: This uses the same `localStorage` namespace as `activePreselect`. That is origin-wide and shared across tabs: a later attach / `onUserIdentified` in another tab can consume this snapshot and render confirmation placements there.

`sessionStorage` still survives checkout → confirmation in the same tab, including a full page load, without the cross-tab steal. Combined with a 2-minute TTL, abandoned checkout then browsing the site still qualifies for replay.

kits/rokt/src/pendingPreselectStorage.ts
```diff
@@ -0,0 +1,62 @@
...
+  attributes: Record<string, unknown>,
+  mpid: string,
+): void {
+  writeNamespacedField(LS_NAMESPACE_KEY, buildPendingPreselectFieldKey(accountId), {
+    expiresAt: Date.now() + PENDING_PRESELECT_TTL_MS,
+    pathname,
+    identifier,
+    attributes,
+    mpid,
+  });
```
rmi22186, line 57: `writeNamespacedField` returns whether the write succeeded; this drops it. Quota / private mode fails silent, so the in-memory queue is still the only copy and dies on navigation — the failure mode this storage exists to fix.

If persist fails, at least surface it on the existing diagnostic/error path so this does not fail closed with no signal.

kits/rokt/src/pendingPreselectStorage.ts
```diff
@@ -0,0 +1,62 @@
...
+}
+
+export function clearPendingPreselect(accountId: string): void {
+  removeNamespacedField(LS_NAMESPACE_KEY, buildPendingPreselectFieldKey(accountId));
+}
```
rmi22186, line 62: Nothing in the kit calls this on `SESSION_END` (page views / UTM are cleared; this key is not). A new session inside the TTL can still replay the previous checkout snapshot.

If this is meant to be “this checkout attempt,” clear on session end as well as on successful dispatch / definite mpid mismatch.

kits/rokt/src/preselection.ts
crisryantan: Fixed in #1424: collectAttributes can pull these from the event/live store even though this.userAttributes filters them, so the persisted snapshot now runs through removeSelectPlacementsAttributePersistenceDeniedAttributes before it's written.

kits/rokt/src/preselection.ts
crisryantan: Fixed in #1424: a recovered fire now keys the active-preselect dedupe on persisted.identifier instead of persisted.pathname, so it can't stamp a skip window against the next live /checkout fire.

kits/rokt/src/pendingPreselectStorage.ts
crisryantan: Fixed in #1424, switched to sessionStorage. That alone doesn't scope recovery to a specific destination page though, just removes the cross-tab leak, so I've left that half open in the new PR rather than guess at a TTL/allowlist tradeoff.

kits/rokt/src/pendingPreselectStorage.ts
crisryantan: Fixed in #1424: setPendingPreselect now returns whether the write succeeded, and a failed persist logs a PRESELECT_QUEUED diagnostic with reason=persist_failed instead of failing silently.

kits/rokt/src/pendingPreselectStorage.ts
crisryantan: Fixed in #1424: cleared on SESSION_END and onLogoutComplete now too, not just gated by the mpid check at recovery time.

kits/rokt/src/preselection.ts
crisryantan: Agreed, and I didn't fix this one in #1424 on purpose rather than guess. There's no destination pathname in PRESELECTION_CONFIG today, just the source pathname and an opaque targetPageIdentifier, so scoping recovery to 'the confirmation page specifically' needs either a new config field (knowing Cinemark's actual confirmation path) or a call on how much abandon-checkout exposure is acceptable at the current 2min TTL. Not something I want to decide unilaterally in a follow-up PR. Open to either direction if you want to weigh in.

PR #1422, opened 2026-09-14
feat(apv): allowlist common search, product and category query params

calvin-pietersen:
## Summary

Extends the auto page view query-param allowlist (#1366) with the parameter names that carry a shopper's **search term**, a **product or variant identifier**, and a **category** on ecommerce sites. Today only `q` and `search` are on the list, so a search on a site that spells the parameter `keyword`, `searchTerm` or `Ntt` arrives with no term, and product and category pages carry no identifier at all.

```js
// before
data: { hostname: 'shop.example.com', title: 'Results', path: '/browse' }

// after
data: { hostname: 'shop.example.com', title: 'Results', path: '/browse', keyword: 'running shoes' }
```

## What is added

- **Site search** — `query`, `keyword`, `keywords`, `term`, `text`, `searchTerm`, `searchText`, `search_query`, `search_text`, `Ntt`. These are the spellings the major platforms and site-search vendors use (Shopify, Salesforce Commerce Cloud, Magento, Oracle Commerce/Endeca and common custom stacks). Matching is case-insensitive, so one spelling per name suffices; `Search`-style variants fold onto `search`.
- **Product and variant** — `pid`, `productId`, `sku`, `skuId`, `variant`.
- **Category** — `category`.

## What is deliberately not added

- Single-letter names other than `q` (for example WooCommerce's `s`): too generic for a global list.
- `cid`: a customer or client id on many sites, not a category.
- Anything that looks like a free-text field, an email, an order id or a token. The allowlist stays an allowlist.

## Notes

- No behaviour change for params already on the list; `q` still precedes `search`, so the dedup-key ordering the tracker tests rely on is unchanged.
- The "Pagination and search" group is split into "Pagination", "Site search", "Product and variant identifiers" and "Category pages" with a comment on each; no existing entry moved out of order.
- Tests cover each new group, the case folding onto the allowlist spelling, and the deliberate exclusions.

https://claude.ai/code/session_01BhRjsmoxU4MARmWfZzLwzv

## Summary by CodeRabbit

* **Enhancements**
  * Page tracking now recognizes additional search query formats, including common search-term spellings.
  * Product, variant, category, and pagination parameters are now supported in tracked page URLs.
  * Supported parameter matching remains case-insensitive.

* **Bug Fixes**
  * Unsupported query parameters, including unrelated identifiers and ambiguous one-letter terms, continue to be excluded while valid search terms are retained.


src/pageViewTracker.ts
```diff
@@ -81,14 +81,38 @@ export const ALLOWED_QUERY_PARAMS: string[] = [
...
+    'query',
+    'keyword',
+    'keywords',
+    'term',
+    'text',
+    'searchTerm',
+    'searchText',
+    'search_query',
+    'search_text',
+    'Ntt', // Endeca
```
alexs-mparticle, line 104: @calvin-pietersen Is this a specific parameter used by Oracle systems? (assuming Endeca is owned by them) I would provide a link to any third party documentation to help add context.

jamesnrokt: @alexs-mparticle I think we're good as is. Just searching Endeca Ntt query param leads to [this page](https://docs.oracle.com/cd/E29584_01/webhelp/mdex_basicDev/src/rbdv_urlparams_ntt.html) and hardcoding a link to third party docs is likely to lead to a stale link pretty quickly with the standard entropy of the internet

jamesnrokt (approved):
Approved and I've hit run on the workflows so standard checks are running

alexs-mparticle (approved):

PR #1424, opened 2026-09-14
fix(rokt): address late review findings on preselect persistence recovery

crisryantan:
## Summary
Follow-up to #1421 — Addresses 5 of 6 findings:

- `pendingPreselectStorage` now writes to `sessionStorage`, not `localStorage`: tab-scoped, so a different tab can't consume a snapshot meant for this one, while still surviving a same-tab full navigation
- Strip deny-listed attributes (`SELECT_PLACEMENTS_ATTRIBUTE_PERSISTENCE_DENY_LIST`) from the persisted snapshot before writing it — `collectAttributes` can resolve them from the event/live store even though the in-memory bag itself filters them out
- A recovered fire now keys the active-preselect dedupe cache by the target identifier, not the source pathname, so it can't block the next live fire on that same pathname
- `setPendingPreselect` now returns whether the write succeeded; a failed persist logs a diagnostic instead of failing silently
- Clear the pending record on `SESSION_END` and `onLogoutComplete`, rather than relying solely on the mpid check at recovery time

**Not addressed**: a recovered fire still isn't scoped to a specific destination page — surviving a navigation to a different page also means surviving a navigation to the wrong one (e.g. an abandoned checkout, not just checkout → confirmation). Fixing that needs either a destination-pathname allowlist in `PRESELECTION_CONFIG` or a product decision on acceptable exposure — flagging rather than deciding unilaterally.

## Test plan
- [x] `npm test` (423 passing)
- [x] `npm run build`

## Summary by CodeRabbit

* **Bug Fixes**
  * Pending preselection data is now isolated per browser tab and cleared when a session ends or a user logs out.
  * Sensitive deny-listed attributes are removed before pending data is saved.
  * Recovered preselection no longer affects deduplication for active page paths.
  * Persistence failures are handled gracefully and surfaced through diagnostics.


kits/rokt/src/pendingPreselectStorage.ts
```diff
@@ -5,6 +5,11 @@ import { isObject, isString } from './utils';
 // doesn't get replayed long after the shopper is gone.
 export const PENDING_PRESELECT_TTL_MS = 2 * 60_000;
 
+// sessionStorage, not localStorage: tab-scoped, so a different tab can't recover a snapshot
+// meant for this one, but it still survives a same-tab full page navigation (checkout to
+// its confirmation page), which is the only case this needs to survive.
+const PENDING_PRESELECT_STORAGE = (): Storage => window.sessionStorage;
```
alexs-mparticle, line 11: \Using `CONSTANT` notation for a wrapper object is misleading.

kits/rokt/src/pendingPreselectStorage.ts
```diff
@@ -30,33 +35,40 @@ function buildPendingPreselectFieldKey(accountId: string): string {
...
-  writeNamespacedField(LS_NAMESPACE_KEY, buildPendingPreselectFieldKey(accountId), {
-    expiresAt: Date.now() + PENDING_PRESELECT_TTL_MS,
-    pathname,
-    identifier,
-    attributes,
-    mpid,
-  });
+): boolean {
+  return writeNamespacedField(
+    LS_NAMESPACE_KEY,
```
alexs-mparticle, line 59: Nit: This is reusing a constant flagged for local storage as a session storage key. We should either unify these into `STORAGE_NAMESPACE_KEY` or separate them so it's unambiguous.

alexs-mparticle (changes requested):

rmi22186 (approved):
lgtm besides alex's comments

crisryantan:
Closing in favor of #1425 — same content, just moved to a branch name that satisfies the semantic-branch-name check (needs a `fix/` prefix).

kits/rokt/src/pendingPreselectStorage.ts
crisryantan: Fair, fixed in #1425 (ff5c9ddb): renamed to getPendingPreselectStorage.

kits/rokt/src/pendingPreselectStorage.ts
crisryantan: Fixed in #1425 (ff5c9ddb): renamed to STORAGE_NAMESPACE_KEY, shared by both backends now.

crisryantan:
Addressed both of CodeRabbit's findings in #1425: storage-backend resolution now happens inside the try (a getStorage callback, resolved only when the try block runs, not as an eagerly-evaluated default parameter). Left the opener-inherited sessionStorage point as-is for now — blocking recovery whenever window.opener is set would also break a legitimate same-origin-popup confirmation flow if this partner ever uses one, and I have no evidence they do either way, so didn't want to guess.

PR #1425, opened 2026-09-14
fix(rokt): address late review findings on preselect persistence recovery

crisryantan:
## Summary
Replaces #1424, which used a branch name the semantic-branch-name check rejects. Same content plus one more round of fixes from that PR's own review.

Follow-up to #1421 — its review from Rob (rmi22186) landed just after merge. Addresses 5 of his 6 findings:

- `pendingPreselectStorage` now writes to `sessionStorage`, not `localStorage`: tab-scoped, so a different tab can't consume a snapshot meant for this one, while still surviving a same-tab full navigation
- Strip deny-listed attributes (`SELECT_PLACEMENTS_ATTRIBUTE_PERSISTENCE_DENY_LIST`) from the persisted snapshot before writing it — `collectAttributes` can resolve them from the event/live store even though the in-memory bag itself filters them out
- A recovered fire now keys the active-preselect dedupe cache by the target identifier, not the source pathname, so it can't block the next live fire on that same pathname
- `setPendingPreselect` now returns whether the write succeeded; a failed persist logs a diagnostic instead of failing silently
- Clear the pending record on `SESSION_END` and `onLogoutComplete`, rather than relying solely on the mpid check at recovery time

Plus, from review on #1424 itself:
- Storage-backend resolution (`window.localStorage`/`sessionStorage`) now happens inside the existing try/catch instead of as an eagerly-evaluated default parameter, since accessing the property itself can throw under some browser privacy settings
- Renamed `LS_NAMESPACE_KEY` → `STORAGE_NAMESPACE_KEY` (it's shared by both backends now) and the misleadingly-CONSTANT_CASE `PENDING_PRESELECT_STORAGE` → `getPendingPreselectStorage`

**Not addressed**: a recovered fire still isn't scoped to a specific destination page — surviving a navigation to a different page also means surviving a navigation to the wrong one (e.g. an abandoned checkout, not just checkout → confirmation). Fixing that needs either a destination-pathname allowlist in `PRESELECTION_CONFIG` or a product decision on acceptable exposure — flagging rather than deciding unilaterally.

## Test plan
- [x] `npm test` (425 passing)
- [x] `npm run build`

## Summary by CodeRabbit

* **Bug Fixes**
  * Pending preselection data is cleared when a session ends or the user logs out, preventing stale data from being restored after re-login.
  * Persisted preselection now excludes attributes that are not permitted for selected placements.
  * Recovered preselection no longer incorrectly suppresses later live dispatches.

* **Reliability**
  * Improved handling of storage access and write failures without disrupting the experience.
  * Pending preselection is isolated to the current browser tab.


kits/rokt/src/storage.ts
```diff
@@ -14,54 +14,70 @@ export function isLocalStorageAvailable(): boolean {
   }
 }
 
-export function readJSON(key: string): unknown {
+const defaultStorage = (): Storage => window.localStorage;
```
alexs-mparticle, line 17: Not a blocker but since this PR is making storage more reusable, we should add some tests around the interchange between session and local storage. I would make this a follow up PR.

rmi22186 (approved):
LGTM.  thanks for addressing alex's comments from #1424!

alexs-mparticle (approved):

kits/rokt/src/storage.ts
crisryantan: ok will do!

crisryantan: Opened #1426 for this — covers write/read/remove independence between the two backends under the shared STORAGE_NAMESPACE_KEY, plus a failure in one not affecting the other.

PR #1426, opened 2026-09-14
test(rokt): cover localStorage/sessionStorage interchange in storage.ts

crisryantan:
## Summary
Follow-up to #1425 per [Alex's review comment](https://github.com/mParticle/mparticle-web-sdk/pull/1425#discussion_r4006899693): storage.ts's helpers now take a `getStorage` backend argument (localStorage by default, sessionStorage for `pendingPreselectStorage`), sharing the same `STORAGE_NAMESPACE_KEY` string across both. Adds explicit test coverage that the two backends stay independent under that shared key — writes/reads/removes against one don't leak into or get clobbered by the other, and a write failure in one backend doesn't affect the other.

## Test plan
- [x] `npm test` (storage.spec.ts: 33 passing)
- [x] `npm run build`

## Summary by CodeRabbit

- **Tests**
  - Added coverage verifying interchangeable storage backends.
  - Confirmed default local storage behavior and explicit session storage support.
  - Added checks for isolated reads, writes, removals, clearing, namespaced helpers, and storage write failures.


rmi22186 (approved):

kits/rokt/test/src/pendingPreselectStorage.spec.ts
```diff
@@ -70,9 +79,23 @@ describe('pendingPreselectStorage', () => {
...
+    });
+
+    it('returns false and does not throw when the write fails', () => {
+      vi.spyOn(Storage.prototype, 'setItem').mockImplementation(() => {
+        throw new DOMException('quota', 'QuotaExceededError');
+      });
+
+      expect(setPendingPreselect(ACCOUNT_ID, '/checkout', 'target-page', { email: 'a@b.com' }, MPID)).toBe(false);
+
+      vi.restoreAllMocks();
```
alexs-mparticle, line 97: Move `vi.restoreAllMocks()` into the shared `afterEach` instead of calling it inline at the end of the test. If the `expect` on line 95 ever fails, this line never runs and `Storage.prototype.setItem` stays mocked for every test that follows in this file.

kits/rokt/test/src/pendingPreselectStorage.spec.ts
```diff
@@ -14,10 +14,19 @@ const MPID = 'mpid-1';
...
 
   afterEach(() => {
     window.localStorage.clear();
+    window.sessionStorage.clear();
+  });
+
+  it('writes to sessionStorage, not localStorage', () => {
+    setPendingPreselect(ACCOUNT_ID, '/checkout', 'target-page', { email: 'a@b.com' }, MPID);
+
+    expect(readJSON(NAMESPACE_KEY, () => window.sessionStorage)).toHaveProperty(`pendingPreselect:${ACCOUNT_ID}`);
```
alexs-mparticle, line 28: Nit: `NAMESPACE_KEY` here is a hardcoded `'mp-rokt-kit'` literal instead of the `STORAGE_NAMESPACE_KEY` constant exported from `../../src/storage`. If that constant ever changes, this test won't catch the drift.

kits/rokt/test/src/pendingPreselectStorage.spec.ts
```diff
@@ -38,12 +47,12 @@ describe('pendingPreselectStorage', () => {
       setPendingPreselect(ACCOUNT_ID, '/checkout', 'target-page', { email: 'a@b.com' }, MPID);
 
       const fieldKey = `pendingPreselect:${ACCOUNT_ID}`;
-      const blob = readJSON(NAMESPACE_KEY) as Record<string, unknown>;
+      const blob = readJSON(NAMESPACE_KEY, () => window.sessionStorage) as Record<string, unknown>;
```
alexs-mparticle, line 50: Nit: this reaches into the raw namespaced JSON blob to force expiry rather than going through the public API or fake timers. Works, and dodges timer flakiness, but couples the test to storage.ts's internal blob shape. Not blocking, just flagging the white-box approach.

PR #1428, opened 2026-09-15
fix(rokt): Update PENDING_PRESELECT_TTL_MS to 5 minutes

alexs-mparticle:
## Background
The pending preselect snapshot is written on the payment page and recovered on the
confirmation page. At 2 minutes it expires before confirmation loads for any checkout
slower than about the median.

## What Has Changed
- `PENDING_PRESELECT_TTL_MS`: 2 minutes -> 5 minutes. Measured checkout tempo puts p50 at
  1m23s and p90 at 6m50s; 5 minutes is a deliberate stop short of the tail, since an
  unconsumed entry stays replayable for the whole window.
- Added a test pinning the constant.

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Pending selections now remain available for up to five minutes before expiring, improving reliability during longer interactions.

- **Tests**
  - Added coverage verifying the updated five-minute expiration window and exact expiration behavior.


kits/rokt/test/src/pendingPreselectStorage.spec.ts
```diff
@@ -29,6 +29,10 @@ describe('pendingPreselectStorage', () => {
     expect(readJSON(NAMESPACE_KEY, () => window.localStorage)).toBeNull();
   });
 
+  it('pins the pending preselect TTL to 5 minutes', () => {
```
crisryantan: can we test the usage of PENDING_PRESELECT_TTL_MS instead of just confirming its indeed 5minutes?

alexs-mparticle: Good call - swapped it for a behavioural test. It now writes a pending record, advances fake timers to `TTL - 1` and asserts `getPendingPreselect` still returns it, then advances one more ms and asserts it is gone, so it exercises the expiry path in `getPendingPreselect` rather than restating the constant.

Mutation-checked: reverting the constant to `4 * 60_000` fails the new test, and the full kit suite passes at 5 minutes (438 tests).

crisryantan (approved):

PR #1433, opened 2026-09-16
fix(forwarders): pass the flat identities dictionary to kit init

mattbodle:
## Summary

`Forwarders.initForwarders(userIdentities, cb)` expects the flat, name-keyed identities
dictionary — `{ customerid: '123', email: 'a@b.com', google: 'g' }`. Seven call sites reach it,
and they disagree about the shape:

| Caller | Shape passed |
| :-- | :-- |
| `processForwarders` (`forwarders.ts`) | `identifyRequest.userIdentities` — flat |
| `setConsentState` (`identity.ts`) | `.userIdentities` — flat |
| `reinitForwardersOnUserChange` (`identity.ts`) | `.userIdentities` — flat |
| `setUserAttribute` (`identity.ts`) | `getUserIdentities()` — **the `{ userIdentities }` wrapper** |
| `removeUserAttribute` (`identity.ts`) | `getUserIdentities()` — **wrapper** |
| `setUserAttributeList` (`identity.ts`) | `getUserIdentities()` — **wrapper** |
| `removeAllUserAttributes` (`identity.ts`) | `getUserIdentities()` — **wrapper** |

`getUserIdentities()` returns `{ userIdentities: {...} }`, so those four callers pass an object
whose only key is `userIdentities`. `buildFilteredUserIdentities` iterates the keys it is given,
and `IdentityType.getIdentityType('userIdentities')` returns `false`, so the helper produced a
single entry of the form:

```js
{ Type: false, Identity: { customerid: '123', email: 'a@b.com', google: 'g' } }
```

`Type: false` never matches an entry in a numeric `userIdentityFilters` list (`inArray` is
`indexOf`-based), so the kit's `init` received one malformed entry carrying every identity of the
current user, including identity types configured to be withheld from that kit. This is
observable whenever a kit is initialised after SDK start — for example a kit gated by a
`filteringUserAttributeValue` conditional-forwarding rule, which becomes eligible on the first
matching `setUserAttribute` call.

The repository already asserts the intended contract in
`should filter user identities from forwarder on init and bring customerid as first ID`: `init`
receives only allowed identities as `{ Type: <number>, Identity: <string> }`.

## Changes

- **`src/identity.ts`** — unwrap the dictionary at the four call sites, matching the three that
  already did.
- **`src/helpers.ts`** — `buildFilteredUserIdentities` now skips identity names that do not map to
  an identity type instead of emitting a `Type: false` entry. This matches what `appendUserInfo`
  in `src/user-utils.ts` already does when it builds `event.UserIdentities`, so the two paths now
  agree. The check is an explicit `=== false`: `IdentityType.Other` is `0`, so a falsy check would
  drop a valid identity type.
- **`src/mp-instance.ts`** — `_Forwarders` was typed `any`, so no call site was checked. Only
  `initForwarders` is typed now; every other member stays `any` as before. Passing the wrapper is
  now a compile error (`TS2559: Type 'IdentityApiData' has no properties in common with type
  'UserIdentities'`), which is what stops a future call site repeating this.
- **`src/apiClient.ts`** — `IAPIClient.prepareForwardingStats` declared `event: IUploadObject`
  while its own implementation takes `SDKEvent`. Typing `initForwarders` requires the declaration
  and the implementation to agree, so the declaration now says `SDKEvent`. Type-only change.

## Tests

- `test/src/tests-forwarders.ts` — two new cases. A kit with
  `userIdentityFilters: [IdentityType.Google]` and a `filteringUserAttributeValue` rule is
  initialised late, once by a `setUserAttribute` call (`includeOnMatch: true`) and once by a
  `removeUserAttribute` call (`includeOnMatch: false`). Each asserts `initCalled` is `false`
  before the write and `true` after it, so the assertions cannot pass by `init` never running;
  then that the two allowed identities are present and well formed, and that the withheld value
  appears nowhere in what the kit received.
- `test/src/tests-helpers.js` — two new cases on `filterUserIdentities`: an object of unrecognised
  keys (including the `userIdentities` wrapper) yields an empty result, and `other` /
  `email_sha256`, whose identity type is `0`, still survive.

All three behavioural cases were confirmed to fail before the change and pass after it. The fourth
is a regression guard for the `0` type and passes either way.

Local runs: karma `TOTAL: 1075 SUCCESS`, jest 739 passed in 30 suites, `eslint` clean, and
`tsc --noEmit` reports no errors in `src/` or `test/` (the 8 pre-existing errors are all in
`node_modules` type definitions and are unchanged).

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected user identity filtering so unsupported identities are excluded while valid identity types remain available.
  * Ensured forwarders receive the correct user identity data when setting, removing, or clearing user attributes.
  * Forwarder initialization now respects configured user-attribute rules and passes only eligible identities.
  * Improved forwarding-stat event handling for more consistent event data processing.

* **Tests**
  * Expanded coverage for identity filtering, supported identity types, and forwarder initialization scenarios.


mattbodle:
This reference is pre-existing rather than newly added: `v3-development` carries `_Forwarders: any; // https://go.mparticle.com/work/SQDSDKS-5767` on this same declaration, and the change expands that declaration while keeping its existing tracker link. It also matches the convention already used across this codebase — `src/` contains 132 `go.mparticle.com/work/SQDSDKS-*` references in 29 files on `v3-development`, including four in `apiClient.ts`.

rmi22186 (approved):

PR #1435, opened 2026-09-16
fix(persistence): validate stored session and per-MPID record shapes

mattbodle:
## Summary

`encodePersistence` writes `gs.csm` as an array of MPIDs and every top level MPID record as an
object. `decodePersistence` accepts whatever the stored JSON happens to contain. The two sides of
that boundary disagree, so a stored value of another shape decodes without error — which means
`initializeStorage`'s existing clear-and-continue path for unreadable storage never sees it — and is
only dereferenced later, part way through `Identity.parseIdentityResponse`.

Two consumers assume the written shape without checking it:

- **`gs.csm`.** `hydrateStoreFromPersistence` copies it into `_Store.currentSessionMPIDs`, and
  `Store.addMpidToSessionHistory` then calls `indexOf` and `push` on it. A stored object throws at
  `indexOf`; a stored string throws at `push`. `reduceAndEncodePersistence` reads the same field
  directly, independently of hydration.
- **A top level MPID record that is not an object.** `findMpidForRequestedIdentity` reads `.mpid`
  off every top level key, so a `null` record throws `TypeError`. Every other consumer of those
  records guards on truthiness, which is why loading and every other code path succeed.

`parseIdentityResponse` applies a successful response as one ordered sequence of Store, persistence
and kit updates inside a single `try`, and its `catch` logs and invokes the developer callback
without undoing what it had already applied. So either throw leaves the response half applied:
`_Store.mpid` has switched to the new MPID, while `swapCurrentUser`, `setUserIdentities`,
`findPrevCookiesBasedOnUI`, `storeDataInMemory`, kit re-initialisation, `setUserIdentity` /
`onLoginComplete` / `onUserIdentified`, the `user_identity_change` event and `processQueuedEvents`
are all skipped — and the callback still reports the server's `200`. For the `csm` case the value is
also re-encoded and written back on every `update()`, because `isNonEmptyArrayOrObject` only drops a
falsy, primitive or empty value, so a non-empty object survives across page loads.

Observed in a browser with a stored `gs.csm` of `{"a":1}`, a session id and no `les`, and a login
answered `200`:

```
_Store.currentSessionMPIDs        {"a": 1}
kit.setUserIdentity called        false
kit.onLoginComplete called        false
kit.onUserIdentified called       false
login callback httpCode           200
login callback user identities    {}          <- the identities just supplied
_Store.mpid                       "loginMPID" <- already switched
_Store.isInitialized              true
stored cu                         "previousMPID" <- never swapped
user_identity_change event sent   false
logged                            Error parsing JSON response from Identity server:
                                  TypeError: _this.currentSessionMPIDs.indexOf is not a function
```

## Change

Validate both shapes in `decodePersistence`, which is the single place every consumer of stored
persistence already routes through (`getCookie` and `getLocalStorage` both call it, and
`getPersistence`, `storeDataInMemory`, `reduceAndEncodePersistence` and `savePersistence` are only
ever reached with its output). That means the value is neither hydrated into the Store nor written
back out on the next `update()`, and no guard is needed at each dereference:

- `decodeGlobalSettings` drops `csm` when the decoded value is not an array. Hydration then falls
  back to `_Store.currentSessionMPIDs`, which is `[]` — the value the Store is constructed with and
  the value `SessionEnd` resets it to — and the next identity call appends the current MPID the
  usual way. `reduceAndEncodePersistence` already treats an absent `csm` as `[]`.
- `decodeMpidRecords` drops a top level record that is not an object, so it is not copied into
  `_Store.nonCurrentUserMPIDs` and not written back out by `setLocalStorage` or `setCookie`.
- `findMpidForRequestedIdentity` also checks the record is an object before reading `.mpid`, so the
  assumption is checked where it is used and not only where the data is decoded.

Behaviour for well formed persistence is unchanged: `decodeGlobalSettings` and `decodeMpidRecords`
take the same path as before for an array `csm` and for object records.

## Tests

Added to `test/src/tests-persistence.ts`, all driving a real `login` answered `200` over a planted
persistence record:

- a stored `csm` that is an object (throws at `indexOf`), a string (throws at `push`) and a number;
- a `null` top level record;
- each asserting the response applies in full — kit `setUserIdentity` / `onLoginComplete` /
  `onUserIdentified`, the identities present on the user handed to the developer callback, `cu` and
  the new MPID's identities persisted, and the `user_identity_change` event sent;
- each also asserting on the raw stored value rather than the decoded view, since decoding is where
  the value is now dropped: the malformed value is absent from storage afterwards, and — as the
  positive half of that check — the correct value is present. For the object `csm` and the `null`
  record, whose values the existing encoding would otherwise write back, absence is the load-bearing
  assertion. For the string and number `csm`, which `isNonEmptyArrayOrObject` already drops on the
  first `update()` regardless, the load-bearing assertion is the positive one: `csm` is stored, as
  an array of MPIDs. Without the change `push`/`indexOf` throws, `currentSessionMPIDs` keeps the
  primitive, and encoding drops it, so nothing is stored under `csm` at all;
- a positive control in every test that a well formed sibling MPID record still round trips, plus a
  dedicated test that a well formed `csm` array and well formed records are handled exactly as
  before;
- a fixture check in each test that the planted record was actually adopted, so none of the above
  can pass because the SDK never read the record;
- a test recording the current behaviour when applying an identity response fails part way through,
  so that gap is visible in the suite (see below).

Each new test was run against the unmodified `src/` first and observed to fail
(`kit setUserIdentity: expected false to equal true`), then to pass with the change.

## Not changed here

`parseIdentityResponse` applying a response non-atomically, and reporting success afterwards, is a
broader issue than the two stored shapes that trigger it. `_Store.isInitialized` is read at eight
places across `mp-instance.ts` and `identityApiClient.ts` and exposed publicly as
`mParticle.isInitialized()`; it gates `mParticle.ready()`, the pre-init ready queue and
`RoktManager.onIdentityComplete`. Not setting it would leave those queues undrained, and rolling
`_Store.mpid` back would also have to undo `setFirstSeenTime`, `setLastSeenTime`, the identity cache
write and the persistence `update()` already performed. That is a separate change; this one removes
the stored-shape triggers and leaves a test documenting the current behaviour.

---

## Unit tests

`test/jest/persistence.spec.ts` gains seven unit tests alongside the karma integration tests, covering the three shape checks directly: a non-array `csm` (object, string and number), a non-object top level record including `null`, and `findMpidForRequestedIdentity` skipping a non-object value. Each pairs the malformed input with a well formed value that must still be decoded or selected, so none of them passes if the surrounding function stops doing its work.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved resilience when stored persistence data is malformed or incomplete.
  - Invalid records are safely ignored or removed during loading, allowing identity sign-ins to complete successfully.
  - Valid sibling records and global settings are preserved while corrupted entries are discarded.
  - Invalid stored CSM values are removed instead of being loaded.
  - Added safeguards for persistence entries missing required MPID information.
  - Improved handling of partial identity-response failures during sign-in.


rmi22186 (approved):
Tests are repetitive, so adding a commit, but otherwise LGTM!

PR #1439, opened 2026-09-17
feat(rokt): let a preselect dispatch fire without its optional attributes

crisryantan:
## Summary

Every key in `attributeKeys` is mandatory today, so a single unresolved key requeues the dispatch instead of firing it. The keys most often unresolved at checkout are not the ones that break a cache match on arrival, so the strictness costs dispatches without buying accuracy.

This adds `optionalAttributeKeys`, a subset `collectAttributes` does not count against completeness, and marks `firstname` and `lastname` optional.

## What is deliberately unchanged

The cache match set. `buildCacheMatchKeys` still returns `attributeKeys`, so an unresolved optional key is still sent as a cache-match key and the controller records it as unset rather than ignoring it. A confirmation page that does supply the key therefore still discards the cached entry, instead of serving offers that were selected without it.

## Why matched-but-optional is the useful shape

A key can be important to offer selection and still be unreliable at checkout, and today those two properties are in direct conflict: requiring the key blocks most dispatches, and dropping it lets a stale value go unnoticed. This flag is what resolves that, and it turns out to be the common case rather than an edge one.

Looking at candidate partners for the next rollout, several of the attributes that carry the most weight in offer selection are either not resolved at checkout yet or are post-purchase by nature, so under today's all-or-nothing rule they cannot be matched on at all. With this flag they can be matched whenever they are present and forgiven when they are not, which is strictly better than both current options.

`firstname` and `lastname` are the two marked optional here because the cache already forgives a match value the arrival stopped reporting, and these are the two the arrival most often omits. The other configured keys are reported on the confirmation page far more consistently, so making them optional would mostly manufacture mismatches.

## Not a step toward dropping them

An earlier version of this description floated removing `firstname` and `lastname` from `attributeKeys` as a follow-up. That was the wrong direction and I have withdrawn it. Both keys are used in offer selection, so taking them out of the match set would let a cached response be reused after they had changed, which is the failure this cache exists to prevent. Keeping them matched while not requiring them to fire is the whole point.

## Trade-off

For a page that today does not dispatch at all, we now dispatch, and the entry is usable whenever the confirmation page also has no value for those keys. The cost is a speculative call that goes unused when it does. Happy to drop this if you would rather keep the stricter gate.

## Testing

Five new specs: the optional key is omitted from the dispatch when unresolved, sent when it resolves, a required key still blocks, the optional list is matched case-insensitively, and an unresolved optional key stays in `cacheMatchKeys` while being absent from the attributes. Full kit suite passes.


kits/rokt/src/preselection.ts
```diff
@@ -108,14 +108,17 @@ function collectAttributes(
...
   const collected: Record<string, unknown> = {};
   const missingKeys: string[] = [];
   for (const key of configEntry.attributeKeys) {
     const eventValue = host.getEventAttributeValue(event, key);
     const value = !isEmpty(eventValue) ? eventValue : (host.userAttributes[key] ?? livePersistedAttributes[key]);
     if (isEmpty(value)) {
-      missingKeys.push(key);
+      if (!optionalKeys.has(key.toLowerCase())) {
+        missingKeys.push(key);
+      }
```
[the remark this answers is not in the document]
crisryantan: The unset state is recorded, just in the controller rather than the kit. cacheMatchKeys still carries firstname, so resolveMatchedAttributes hits its `!(key in matchableAttributes)` branch and stores UNSET_MATCH_ATTRIBUTE for it, which is the representation the cache compares against.

Putting `''` or null in the attributes bag instead lands on that same sentinel via isEmptyMatchValue, but it also puts an empty firstname on the /offers request, since attribute values go over the wire verbatim. That changes what we send for no change in cache behaviour, and a second parallel bag in the kit to avoid it is complexity for the same outcome.

Fair that nothing pinned the invariant though, so a9084f37 adds a test through the real kit: attributes omits the key, cacheMatchKeys keeps it.

kits/rokt/src/preselection.ts
```diff
@@ -108,14 +108,17 @@ function collectAttributes(
   configEntry: PreselectionConfigEntry,
 ): { collected: Record<string, unknown>; missingKeys: string[] } {
   const livePersistedAttributes = host.filteredUser?.getAllUserAttributes?.() || {};
+  const optionalKeys = new Set((configEntry.optionalAttributeKeys ?? []).map((key) => key.toLowerCase()));
```
alexs-mparticle: nit:
```suggestion
  const optionalPreselectionKeys = new Set((configEntry.optionalAttributeKeys ?? []).map((key) => key.toLowerCase()));
```

alexs-mparticle (changes requested):

kits/rokt/src/preselection.ts
crisryantan: Taken, renamed to optionalPreselectionKeys.

alexs-mparticle (approved):

PR #1445, opened 2026-09-18
fix(rokt): validate recovered preselection attributes

mattbodle:
## Background

Persisted Rokt preselection attempts must still satisfy the current configuration before recovery dispatches them. Persistence intentionally removes some cart attributes, so an otherwise complete attempt can become incomplete after navigation.

## What Has Changed

- Validate the saved source path, target identifier, persistence-safe attributes, and required values before dispatching a recovered preselection.
- Clear and skip incomplete or stale attempts while recording the missing persisted attribute.
- Cover incomplete recovery and changed configuration.

## Related PR

- The focused `cartItems` configuration change is isolated in #1448.

## Verification

- Rokt kit tests: 457 passed.
- Rokt kit build completed.
- Changed files have no non-formatting ESLint errors. The existing kit-wide formatter divergence remains and is unchanged by this PR.

## Checklist

- [ ] Self-review completed
- [ ] Tests added or updated
- [ ] Tested locally


mattbodle:
Re: @coderabbitai's [review](https://github.com/mParticle/mparticle-web-sdk/pull/1445#pullrequestreview-5253435118) and [checks](https://github.com/mParticle/mparticle-web-sdk/pull/1445#issuecomment-5737333687):

- **Attribute casing:** this is an existing collection behavior, not a compatibility regression introduced here. At base `c48c541d`, `collectAttributes` already reads `host.userAttributes[key]` / `livePersistedAttributes[key]` exactly and writes snapshots using `collected[key]` with the configured spelling. Executing the actual base/head functions with synthetic exact-case, mixed-case required and mixed-case optional inputs produces identical results. Required mixed-case inputs are already rejected before snapshot persistence; optional inputs remain optional. Case-insensitive source lookup would be a separate behavior change that also needs consistent event/user precedence and dispatch/matching tests. This PR deliberately preserves that behavior while preventing incomplete recovered dispatches.
- **Docstring warning:** the extracted internal helper and regression test names describe the behavior directly. No additional explanatory comment is needed for that extraction; avoiding restatement keeps this bounded fix readable.
- **ESLint tool warning:** the bot attempted to parse the kit's JavaScript flat-config module as YAML. No lint configuration or dependencies changed here. Root lint passes; local comparison finds only the existing kit formatter conflict and no non-formatting errors in the changed files. The PR describes that limitation rather than claiming a clean native kit lint run.

The current head is `2c495cc3`. Core and kit tests, Node 18–25 Jest, integration tests, coverage checks, BrowserStack (including beta), and Android bridge tests have passed. The older cancelled BrowserStack push run is superseded by the successful PR run. The local Jest teardown failure remains disclosed separately from the successful CI runs.

mattbodle:
@coderabbitai review

crisryantan:
Can we just make a PR to simply remove cartItems for dsw for now? I think the cacheMatchKeys is kind of conflicting with the optionalAttributrKeys

mattbodle:
Addressed in a4c170e5: removed `cartItems` directly from the DSW preselection configuration and reverted the `cacheMatchKeys` override and its supporting test machinery. The focused config test and all 458 Rokt kit tests pass.

mattbodle:
@coderabbitai review

kits/rokt/src/preselection.ts
```diff
@@ -101,30 +101,35 @@ function getUserId(filteredUser: IMParticleUser | null | undefined): string | nu
...
   const livePersistedAttributes = host.filteredUser?.getAllUserAttributes?.() || {};
-  const optionalPreselectionKeys = new Set((configEntry.optionalAttributeKeys ?? []).map((key) => key.toLowerCase()));
 
   const collected: Record<string, unknown> = {};
-  const missingKeys: string[] = [];
   for (const key of configEntry.attributeKeys) {
     const eventValue = host.getEventAttributeValue(event, key);
     const value = !isEmpty(eventValue) ? eventValue : (host.userAttributes[key] ?? livePersistedAttributes[key]);
     if (isEmpty(value)) {
-      if (!optionalPreselectionKeys.has(key.toLowerCase())) {
```
crisryantan, line 119: Why are we removing this again

mattbodle: The validation is still applied here; it moved into `getMissingRequiredAttributeKeys` so the same required/optional-key logic can also validate a recovered persisted snapshot, where there is no event or host collection step to reuse. `collectAttributes` now builds `collected` first and returns `missingKeys` from that shared helper.

crisryantan:
Had a look at this against prod because we're chasing DSW's hit rate. The recovery-validation half looks right to me. On the `cartItems` removal I think the evidence says it won't move the number, and I'd rather flag that now than have it land and look like it didn't work.

### cartItems is never the only mismatching key

Pulled the `mismatchedKeys` payload from `PRESELECT_CACHE_MISS_KEY_MISMATCH`. DSW's sample is thin because client logging is 1% sampled, 3 mismatches in 12h:

```
attribute.totalprice, attribute.cartItems                    x2
attribute.firstname, attribute.lastname, attribute.cartItems x1
```

`cartItems` is in all three, so the hunch is right as far as it goes. But it is never alone, so removing it from the key leaves every one of those still mismatching on something else.

Ulta has the same attribute configured and far more volume, 112 sampled mismatches. Dropping the `identifier`-only rows (those are the checkout page's own call consulting the cache, never hit candidates) leaves 19 real ones:

| mismatched keys | n |
|---|---|
| `amount, cartItems` | 10 |
| `amount` | 6 |
| `amount, cartItems, paymenttype, ccbin` | 2 |
| `amount, cartItems, paymenttype` | 1 |

`amount` is in 19 of 19. `cartItems` appears alone in zero. So on both partners the price attribute is the binding constraint, which makes sense given the order total finalises between the payment page and the confirmation page as tax, shipping and promos land.

For scale, DSW is currently 27 hits against 298 key mismatches, so about 8%, versus Ulta at roughly 19%.

### The price attribute looks safe to drop

Checked `lake_campaignstore.audience` for live published audiences, 83,171 of them:

| targeted attribute | live audiences |
|---|---|
| `rokt.totalprice` | **0** |
| `rokt.cartItems` | **0** |
| `rokt.amount` | **7** (2 live campaigns) |
| `rokt.paymenttype` | 593 |
| `rokt.ccbin` | 367 |
| `rokt.loyaltytier` | 18 |

Worth noting the naive check is misleading here: 78,325 audiences contain `"amount"`, but that is `pricing.amount`, the bid, not cart value. The targeting form is `{"attributeName": "rokt.amount", "conditionType": 15, ...}`.

Of the 7, none have served Ulta in the last 7 days. One campaign has served DSW, 17 events in 7 days.

### Suggestion

Drop the price attribute from the cache key alongside `cartItems`, so `totalprice` for DSW and `amount` for Ulta. Nothing targets either of `rokt.totalprice` or `rokt.cartItems`, and `rokt.amount` is 7 audiences fleet-wide with essentially no exposure on these two partners.

Two caveats worth deciding on rather than assuming:

- If we stop declaring it we should stop sending it too, otherwise the cache is looser than the wire, which AGENTS.md calls out as the one direction we don't want.
- DSW's config uses `totalprice` while the audience targeting uses `rokt.amount`. Whether those are the same thing after attribute mapping is worth confirming against the registry before anyone leans on the 17-event number.

Happy to be wrong on this if the recovery path is the dominant cause, and DSW's `miss_no_entry` is 346 against 298 key mismatches so that half of the PR may well be the bigger win.

crisryantan:
Follow-up on whether the checkout placement needs `cartItems` even if the confirmation one doesn't. Checked the audiences that actually served each page type on both partners over 7 days, joined to their published targeting:

| partner | page type | audiences served | events | using `cartItems` | using price |
|---|---|---|---|---|---|
| Ulta | Payment | 20 | 191,733 | **0** | 0 |
| Ulta | Confirmation | 588 | 57,776 | **0** | 0 |
| DSW | Payment | 85 | 14,923 | **0** | 1 |
| DSW | Confirmation | 552 | 10,749 | **0** | 3 |

Nothing serving either page type on either partner targets `cartItems`. The price attribute shows up on 4 of DSW's 637 serving audiences and none of Ulta's 608.

Worth separating the two things anyway, because they are independent. `PRESELECTION_CONFIG.attributeKeys` governs the speculative dispatch and the cache key for the confirmation page. The payment page makes its own ordinary `selectPlacements` call with the full attribute set and is not touched by this config at all, so removing a key here cannot degrade the checkout placement even in principle.

So the `cartItems` removal is safe on both counts. My earlier point stands that it alone will not recover hits, since the price attribute is in every sampled candidate miss, but there is no reason to hold it back.

mattbodle:
The focused `cartItems` configuration change has been split into #1448. This PR now contains only the recovered-preselection validation work.

crisryantan (approved):

PR #1447, opened 2026-09-19
fix(rokt): remove cart items from preselection

mattbodle:
## Background

Confirmation-page preselection should not include `cartItems` yet. Removing it from the configured attribute set also keeps it out of cache matching without adding a second cache-key configuration path.

## What Has Changed

- Remove `cartItems` from the confirmation-page preselection configuration.
- Add a focused regression test for the confirmation-page configuration.

## Related PR

- Recovered preselection validation remains independently scoped in #1445.

## Verification

- Focused confirmation-page configuration test: 1 passed.
- Rokt kit tests: 447 passed.
- Rokt kit build completed; existing repository TypeScript diagnostics remain unchanged.
- Changed files have no non-formatting ESLint errors.

## Checklist

- [ ] Self-review completed
- [ ] Tests added or updated
- [ ] Tested locally


kits/rokt/test/src/preselectionConfig.spec.ts
```diff
@@ -0,0 +1,15 @@
+import { describe, expect, it } from 'vitest';
+import { PRESELECTION_CONFIG } from '../../src/preselectionConfig';
+
+describe('preselection config', () => {
```
alexs-mparticle, line 4: I don't think this test is necessary since we're just updating a config

PR #1448, opened 2026-09-19
fix(rokt): remove cart items from preselection

mattbodle:
## Background

Confirmation-page preselection should not include `cartItems` yet. Removing it from the configured attribute set also keeps it out of cache matching without adding a second cache-key configuration path.

## What Has Changed

- Remove `cartItems` from the confirmation-page preselection configuration.

## Verification

- Rokt kit tests: 446 passed.
- Rokt kit build completed; existing repository TypeScript diagnostics remain unchanged.
- Changed files have no non-formatting ESLint errors.

## Checklist

- [ ] Self-review completed
- [ ] Tests added or updated
- [ ] Tested locally


kits/rokt/test/src/preselectionConfig.spec.ts
```diff
@@ -0,0 +1,15 @@
+import { describe, expect, it } from 'vitest';
+import { PRESELECTION_CONFIG } from '../../src/preselectionConfig';
+
+describe('preselection config', () => {
```
crisryantan: We don't need this test

crisryantan (approved):
Id cleanup the test but happy to approve once that's done

kits/rokt/test/src/preselectionConfig.spec.ts
mattbodle: Removed in a63113e1. The Rokt kit suite passes with 446 tests.

alexs-mparticle (approved):

alexs-mparticle (approved):

